PR #6, opened 2021-10-18
Remove hard-coded secret key in api.auth

JenySadadia:
1. Remove hard-coded secret key in api.auth and load it from the settings instead
2. Created secret key using openssl command and stored it in newly created .env file
3. Added an environment .env file to docker-compose.yaml
4. Updated README.md with instructions to create the key and store it in the .env file

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,15 @@ use it with other tools: http://localhost:8001/openapi.json
 
 ## Authentication
 
+### Generate SECRET KEY and add it in environment file
+Generate a new key for Authentication using following command:
+```
+$ openssl rand -hex 32
+```
+Store the generated key to .env file and give its path to docker-compose.yaml file.
+We have loaded secret key using FastAPI settings feature in api.auth
+Please find the env.sample file in the base directory to store secret key and copy the file to your .env file.
+
 Some parts of the API require the user to be authenticated, for example to
 submit data to store in the database or access restricted data.  The first
 thing to do is create a user account.  For example, with a user called `bob`
--- a/api/auth.py
+++ b/api/auth.py
@@ -6,7 +6,7 @@
 from datetime import datetime, timedelta
 from jose import JWTError, jwt
 from passlib.context import CryptContext
-from pydantic import BaseModel
+from pydantic import BaseModel, BaseSettings
 from typing import Optional
 from .models import User
 
@@ -20,12 +20,11 @@ class TokenData(BaseModel):
     username: Optional[str] = None
 
 
-class Authentication:
-    # ToDo: use settings https://fastapi.tiangolo.com/advanced/settings/
+class Settings(BaseSettings):
+    secret_key: str
 
-    # To generate a secret key: openssl rand -hex 32
-    SECRET_KEY = \
-        "fad13f9cf6e5ff51daac4e6784529daa9d1dc8a80703730221f1090066d929c4"
+
+class Authentication:
 
     ALGORITHM = "HS256"
 
@@ -35,6 +34,7 @@ class Authentication:
     def __init__(self, db):
         self._pwd_context = CryptContext(schemes=["bcrypt"], deprecated="auto")
         self._db = db
+        self._settings = Settings()
 
     def get_password_hash(self, password):
         return self._pwd_context.hash(password)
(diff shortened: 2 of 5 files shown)


JenySadadia:
@gctucker Modified commit to pass all the checks.

README.md
```diff
@@ -152,3 +152,11 @@ example, another Thing was created in the meantime):
...
 {"things":[{"_id":"615f3313b231dd5cccf9c996","name":"answer","value":42},{"_id":"615f33abb231dd5cccf9c997","name":"spanner","value":18}]}
 ```
+
+### Generate SECRET KEY and add it in environment file
+Generate a new key for Authentication using following command:
+```
+$ openssl rand -hex 32
+```
+Store the generated key to .env file and give its path to docker-compose.yaml file.
+We have loaded secret key using FastAPI settings feature in api.auth
```
gctucker, line 162: Please move this to the `## Authentication` section as the secret key is required before being able to create user tokens.

api/auth.py
```diff
@@ -20,12 +20,22 @@ class TokenData(BaseModel):
     username: Optional[str] = None
 
 
+class Settings(BaseSettings):
+    secret_key: str
+
+    class Config:
+        env_file = "../.env"
```
gctucker, line 27: This can be dropped, the environment file will be loaded by docker-compose and its contents will be exported as environment variables.  So when the container is started, an environment variable `SECRET_KEY` should already be there.  Then the FastAPI settings will read that and populate the `Settings.secret_key` value.

See the documentation here: https://fastapi.tiangolo.com/advanced/settings/

api/auth.py
```diff
@@ -20,12 +20,22 @@ class TokenData(BaseModel):
...
+
+
 class Authentication:
     # ToDo: use settings https://fastapi.tiangolo.com/advanced/settings/
```
gctucker, line 31: This comment can be dropped as settings are now being used.

api/auth.py
```diff
@@ -20,12 +20,22 @@ class TokenData(BaseModel):
...
 
+    settings = Settings()
```
gctucker, line 33: Please move this to the `Authentication()` constructor:
```python
    def __init__(self, db):
        self._pwd_context = CryptContext(schemes=["bcrypt"], deprecated="auto")
        self._db = db
        self._settings = Settings()
```
Then `SECRET_KEY` can be replaced with `self._settings.secret_key` in the class methods.

As the settings are loaded dynamically, they should be a member of each `Authentication` instance which is created at runtime rather than a static member of the class which is created only once when the Python module is loaded.

.env
```diff
@@ -0,0 +1 @@
+SECRET_KEY=40341b8c51e75263efc9ca40f05bb6132255d3399c7804c46dcf6892e364081d
```
gctucker, line 1: Each instance should have its own unique key, so the `.env` file should not be kept under source control.  We could however provide a sample file or provide an example directly in the documentation (README), such as:

```
# Generate with openssl rand -hex 32
SECRET_KEY=
```

api/auth.py
```diff
@@ -20,12 +20,22 @@ class TokenData(BaseModel):
...
+    print("secret_key :", settings.secret_key)
     # To generate a secret key: openssl rand -hex 32
-    SECRET_KEY = \
-        "fad13f9cf6e5ff51daac4e6784529daa9d1dc8a80703730221f1090066d929c4"
+    # SECRET_KEY = \
+    # "fad13f9cf6e5ff51daac4e6784529daa9d1dc8a80703730221f1090066d929c4"
```
gctucker, line 37: This can now be removed.

api/auth.py
```diff
@@ -20,12 +20,22 @@ class TokenData(BaseModel):
...
+
+    class Config:
+        env_file = "../.env"
+
+
 class Authentication:
     # ToDo: use settings https://fastapi.tiangolo.com/advanced/settings/
 
+    settings = Settings()
+    print("secret_key :", settings.secret_key)
```
gctucker, line 34: This may be useful to you to check the changes are working, but please remove it before the PR can be merged.

.env
gctucker: The new commit has a `.env` file with `SECRET_KEY=`, which is fine except that the file is still under source control.  So when a user edits it, it will be seen as a change by Git.  This can be a problem as users are likely to commit their secret key or they will always have a change never committed in this file.

I guess a simple way to avoid this issue is to rename this `.env` file as `env.sample` and mention in the README that it needs to be copied to `.env` to be loaded by docker-compose.  It's probably also worth adding `.env` to the `.gitignore` file to be sure it's never added in git.

gctucker:
> Okay. @gctucker Thanks for all your support :)

You're welcome.  Thanks for contributing :)

PR #19, opened 2021-10-29
docker/ssh : Add SSH container

JenySadadia:
Add Dockerfile in docker/ssh with openssh server installed.
Add kernelci user in ssh/Dockerfile.

### DIFF
--- a/docker/ssh/Dockerfile
+++ b/docker/ssh/Dockerfile
@@ -1,0 +1,24 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2021 Jeny Sadadia
+# Author: Jeny Sadadia <jeny.sadadia@gmail.com>
+
+FROM debian:bullseye-slim
+MAINTAINER "KernelCI TSC" <kernelci-tsc@groups.io>
+
+# Install openssh-server
+RUN apt update && apt install  openssh-server -y
+
+# Create privilege separation directory
+RUN mkdir -p /run/sshd
+
+# Create kernelci user
+RUN useradd kernelci -u 1000 -d /home/kernelci -s /bin/bash
+RUN mkdir -p /home/kernelci
+RUN chown kernelci: /home/kernelci
+
+# SSH server will run on PORT 22
+EXPOSE 22
+
+# Run SSH server in detached mode
+CMD ["/usr/sbin/sshd","-D"]


JenySadadia:
Hi @gctucker,
I inspected the issue "Missing privilege separation directory: /run/sshd" on docker run command.

We used sshd binary to run ssh server on our previous command.
```
 $ sudo /usr/sbin/sshd -D
```
The standard way to start openssh server is to use following command :
```
$ sudo /etc/init.d/ssh start -D
```

The above script is creating privilege directory /run/sshd if it is missing from the location.
Below is the code snippet from /ect/init.d/ssh file :
```
check_privsep_dir() {
    # Create the PrivSep empty dir if necessary
    if [ ! -d /run/sshd ]; then
        mkdir /run/sshd
        chmod 0755 /run/sshd
    fi
}
```
When we use above command to run the server, we don't need to create /run/sshd manually.

I modified CMD in Dockerfile and it resolved our issue. I am able to SSH on docker container after the latest change in the Dockerfile.
I have committed the changes for PR.

Kindly review it. Thanks.

gctucker:
Thanks for investigating.  One thing though, is that Docker containers typically don't run services in the same way that init processes normally do.  Instead of running a System V script from `/etc/init.d/` or starting a systemd service, the Dockerfile is the one in charge of starting the service.  While running the init.d script appears to work in this case with `sshd -D`, it's still not ideal as the script is calling `start-stop-daemon` which is intended to run services in the background.

Here's an article explaining why Docker images don't normally have systemd support, which is the same reason why System V (/etc/init.d/) scripts aren't normally run in Docker:
https://medium.com/swlh/docker-and-systemd-381dfd7e4628

So I think it's best practice to not call the `/etc/init.d/sshd` script but include any required steps in the Dockerfile, which in this case means creating the `/run/sshd` directory in the Docker image.  Having found out about the `/etc/init.d/ssh` script is really helpful in any case, as it shows there really is a requirement for the `/run/sshd` directory to be created before starting `sshd`.  I hope this makes sense :)

JenySadadia:
Hi @gctucker,

Thanks for the explanation and sharing article link.
Now I got why /etc/init.d/sshd didn't work when I tried it without using detach mode like normally we do on linux machine.
I think I need to rollback the commit to prior version.
I will commit the changes soon.
Thanks for your guidance throughout the whole contribution period :)

docker/ssh/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# Create kernelci user
+RUN useradd kernelci -u 1000 -d /home/kernelci -s /bin/bash
+RUN mkdir -p /home/kernelci
+RUN chown kernelci: /home/kernelci
+
+# SSH server will run on PORT 22
+EXPOSE 22
+
+# Run SSH server in detached mode
+CMD ["/usr/sbin/sshd","-D"]
```
gctucker, line 24: Note: The red sign at the end of the line is because the file is missing a final new line.  It's a UNIX thing and Git will also complain about it.  Not a big issue, if you don't mind pushing to the branch once again then it would be a bit cleaner to resolve it.

Here's a related discussion about this topic:
https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

JenySadadia: Committed the changes. @gctucker

gctucker (approved):
Great, tested again locally with the latest changes.

PR #34, opened 2021-11-15
api.db: fix obj.id check

gctucker:
The objects passed to Database.create() always have a .id attribute as
it's defined in the Pydantic models.  However, it should be None in
the case of a new object since it hasn't been saved in the database
yet.  As such, fix the check to only create objects when the .id is
None.

Fixes: 54f51945e591 ("api: add api.db.Database() abstraction class")


mgalka (approved):
looks good

Issue #37, opened 2021-11-19
Align api.models.Revision with meta-data produced by kernelci-core


gctucker:
The `api.models.Revision` model serves the same purpose as the `revision` entry in the build meta-data produced by kernelci-core.  As things currently stand, this is the model from #35:

```python
class Revision(BaseModel):
    tree: str
    branch: str
    commit: str
```
and this is an example of what kernelci-core stores in bmeta.json:
```json
    "revision": {
        "branch": "master",
        "commit": "fa55b7dcdc43c1aa1ba12bca9d2dd4318c2a0dbf",
        "describe": "v5.16-rc1",
        "describe_verbose": "v5.16-rc1",
        "tree": "mainline",
        "url": "https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git"
    }
```

Aligning both things means we could then directly submit a Revision object from the build meta-data, and conversely the meta-data could be recreated when getting a Node.Revision object back from the API.  As outlined in the past, `describe_verbose` seems really superfluous.  It is very useful to have a documented and reproducible way to generate this string, but there's no value of having two versions of `git describe`.  So the code currently creating `describe_verbose` could be kept to populate the `describe` attribute, and the `describe_verbose` attribute could be dropped.

Essentially, this is the current Git command from `kernelci.build.get_describe_verbose()`:
```
git describe --always --match=v[1-9]\*
```
which relies on all the version tags from mainline or stable to have been fetched first.  This provides a unified version of describing the commit in a more human-readable way than the plain SHA1.

Then the `tree` attribute has shown to be very useful, as Git URLs sometimes change or alternative ones may be used to access the same repository.  In summary, this is what the proposed models and meta-data should look like:

```python
class Revision(BaseModel):
    tree: str
    url: str
    branch: str
    commit: str
    describe: str
```

```json
    "revision": {
        "tree": "mainline",
        "url": "https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git",
        "branch": "master",
        "commit": "fa55b7dcdc43c1aa1ba12bca9d2dd4318c2a0dbf",
        "describe": "v5.16-rc1"
    }
```

The `Revision` model could of course have some extra validation to check the fields follow the expected format (valid URL etc.).

- [x] #42 
- [ ] ~~kernelci-core bmeta.json update to drop describe_verbose~~  This is not necessary, there's no requirement for full compatibility between the legacy system and the new API.

gctucker:
[...]

PR #42, opened 2021-11-22
api.models: add Revision.url and .describe

gctucker:
Add .url and .describe fields to the Revision model, to contain the
Git URL and Git describe strings.  This should help align with the
current bmeta.json schema used by kernelci-core.

Part of #37


gctucker:
Tested OK manually with the PoC.

Issue #43, opened 2021-12-06
Mongodb docker container issue on WSL


JenySadadia:
**Issue :** 

Received below error when tried to start mongo db container on WSL.
```
Starting kernelci-db ... done
Attaching to kernelci-db
kernelci-db | {"t":{"$date":"2021-12-06T10:54:24.100+00:00"},"s":"I",  "c":"NETWORK",  "id":4915701, "ctx":"-","msg":"Initialized wire specification","attr":{"spec":{"incomingExternalClient":{"minWireVersion":0,"maxWireVersion":13},"incomingInternalClient":{"minWireVersion":0,"maxWireVersion":13},"outgoing":{"minWireVersion":0,"maxWireVersion":13},"isInternalClient":true}}}
kernelci-db | {"t":{"$date":"2021-12-06T10:54:24.100+00:00"},"s":"I",  "c":"CONTROL",  "id":23285,   "ctx":"main","msg":"Automatically disabling TLS 1.0, to force-enable TLS 1.0 specify --sslDisabledProtocols 'none'"}
kernelci-db | {"t":{"$date":"2021-12-06T10:54:24.100+00:00"},"s":"W",  "c":"ASIO",     "id":22601,   "ctx":"main","msg":"No TransportLayer configured during NetworkInterface startup"}
kernelci-db | {"t":{"$date":"2021-12-06T10:54:24.100+00:00"},"s":"I",  "c":"NETWORK",  "id":4648601, "ctx":"main","msg":"Implicit TCP FastOpen unavailable. If TCP FastOpen is required, set tcpFastOpenServer, tcpFastOpenClient, and tcpFastOpenQueueSize."}
kernelci-db | {"t":{"$date":"2021-12-06T10:54:24.101+00:00"},"s":"W",  "c":"ASIO",     "id":22601,   "ctx":"main","msg":"No TransportLayer configured during NetworkInterface startup"}
kernelci-db | {"t":{"$date":"2021-12-06T10:54:24.101+00:00"},"s":"I",  "c":"REPL",     "id":5123008, "ctx":"main","msg":"Successfully registered PrimaryOnlyService","attr":{"service":"TenantMigrationDonorService","ns":"config.tenantMigrationDonors"}}
[...]

PR #45, opened 2021-12-07
docker/docker-compose.yaml : Add fix for mongodb container on WSL

JenySadadia:
Added docker volume for mongodb.

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -21,7 +21,7 @@ services:
     container_name: 'kernelci-db'
     image: 'mongo:5.0'
     volumes:
-      - './db/data:/data/db'
+      - mongodata:/data/db
 
   redis:
     container_name: 'kernelci-redis'
@@ -36,3 +36,6 @@ services:
       - './storage/data:/usr/share/nginx/html'
     ports:
       - 8002:80
+
+volumes:
+  mongodata:


gctucker:
Tip: If you add something like "Fixes #43" in the PR description it should automatically link it with the issue.

JenySadadia:
> Tip: If you add something like "Fixes #43" in the PR description it should automatically link it with the issue.

Thanks for the tip. I will use that from now onwards :)

gctucker:
I'm hitting this error:
```
Successfully built dd4ee458c5d7
Successfully tagged docker_api:latest
Creating kernelci-db      ... error
Creating kernelci-redis ... 
Creating kernelci-api   ... 
Creating kernelci-storage ... 
Creating kernelci-redis   ... done
Creating kernelci-api     ... done
Creating kernelci-storage ... done

ERROR: for db1  Cannot create container for service db1: Conflict. The container name "/kernelci-db" is already in use by container "4d3cd3ea6adb23a9efb5be0cc7d189857936edfc8e5b879608fc0c8b5d79e409". You have to remove (or rename) that container to be able to reuse that name.
ERROR: Encountered errors while bringing up the project.
```
Maybe it's just a problem with my local setup, I'll try again from a clean setup but still it would be interesting to understand why this is happening.

gctucker:
OK I just had to remove an old stale container.  For some reason it wasn't a problem with the volume using a bind-mounted directory.

docker/docker-compose.yaml
```diff
@@ -17,11 +17,11 @@ services:
     env_file:
       - ../.env
 
-  db:
+  db1:
```
gctucker, line 20: Why was this renamed to `db1`?  I think it should be OK to leave it as `db`.

gctucker:
I renamed the service back to `db` and got this to work fine locally.  Of course this created a new empty Mongo database but that's normal since it's a different Docker volume (it's possible to transfer the data if needed, for example using mongodump / mongorestore).

docker/docker-compose.yaml
JenySadadia: I hit the same issue of ``` ERROR: for db  Cannot create container for service db: Conflict. The container name "/kernelci-db" is already in use ```. So I tried to rename the service to db1 and check.
It was for testing only.
Yes, we don't need to rename the service name.

JenySadadia:
> I renamed the service back to `db` and got this to work fine locally. Of course this created a new empty Mongo database but that's normal since it's a different Docker volume (it's possible to transfer the data if needed, for example using mongodump / mongorestore).

Okay. I pushed the changes with 'db' service name.

gctucker (approved):
Tested OK locally using kernelci-pipeline.

PR #48, opened 2021-12-09
Fix pub/sub events : publish

JenySadadia:
api/client.py: Fix Unprocessable Entity Exception for listen
api/pubsub.py: Fix cloudevents MissingRequiredFields exception for publish

### DIFF
--- a/api/pubsub.py
+++ b/api/pubsub.py
@@ -70,6 +70,11 @@ class PubSub:
         await self._redis.publish(channel, message)
 
     async def publish_cloudevent(self, channel, data, attributes=None):
+        """
+        CloudEvent contructor needs to have below params:
+            attributes : dict with type and source keys
+            data : dict with message key
+        """
         if attributes is None:
             attributes = {
                 "type": "api.kernelci.org",
--- a/api/main.py
+++ b/api/main.py
@@ -152,4 +152,4 @@ async def listen(sub_id: int, user: User = Depends(get_user)):
 async def publish(raw: dict, channel: str, user: User = Depends(get_user)):
     attributes = dict(raw)
     data = attributes.pop('data')
-    await pubsub.publish_cloudevent(channel, attributes, data)
+    await pubsub.publish_cloudevent(channel, data, attributes)


gctucker (commented):
It seems it would be better to have 2 separate PRs as this is fixing 2 separate issues, so we could merge them independently.  It's good they're in 2 separate patches.

There's just a small thing about `client.py`, otherwise it looks good.

I don't understand the issue with the CloudEvent data though.

gctucker:
> It's good they're in 2 separate patches.

Actually this PR has only one patch, so that's a problem too as there are really 2 unrelated changes involved here.

Maybe you could drop the `client.py` changes from this PR and create another one to address it separately?  The title of the PR is about the pub/sub issue.

JenySadadia:
> 

Yes @gctucker , I will close this one and open 2 separate PRs.

api/client.py
```diff
@@ -51,10 +54,13 @@ def listen(args):
...
-        url = urllib.parse.urljoin(args.url, path)
-        res = requests.post(url, headers=headers)
-        res.raise_for_status()
+        try:
+            path = '/'.join(['unsubscribe', args.channel])
+            url = urllib.parse.urljoin(args.url, path)
+            res = requests.post(url, headers=headers)
+            res.raise_for_status()
+        except Exception as e:
+            print(f"Error: {e}")
```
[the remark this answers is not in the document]
JenySadadia: > I don't think we need a try-except block here as nothing is being done differently when catching the exception. It's important to catch the KeyboardInterrupt in the main loop to stop and unsubscribe, but then if unsubscribing fails there's nothing we can do about it other than print the exception - which is what happens if you don't catch it at all.

In case of unsubscribing, If we don't catch the exception, It ends up throwing it:
```
Traceback (most recent call last):
  File "/home/kernelci/api/client.py", line 94, in <module>
    args.func(args)
  File "/home/kernelci/api/client.py", line 71, in listen
    res.raise_for_status()
  File "/home/kernelci/.local/lib/python3.10/site-packages/requests/models.py", line 953, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 422 Client Error: Unprocessable Entity for url: http://localhost:8000/unsubscribe/node
```
That's why I added try-except block in finally.

api/pubsub.py
```diff
@@ -70,10 +70,19 @@ async def publish(self, channel, message):
...
+        """
+        event_data = attributes
+        attributes = None
         if attributes is None:
             attributes = {
                 "type": "api.kernelci.org",
                 "source": self._settings.cloud_events_source,
             }
-        event = CloudEvent(attributes=attributes, data=data)
+        event = CloudEvent(attributes=attributes, data=event_data)
```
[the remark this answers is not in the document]
JenySadadia: Here is what we are receiving in publish_cloudevent function arguments:
```
attributes {'message': 'hii'}
data {'specversion': '1.0', 'id': '8cf20531-f441-4ce3-b08d-02229f2b9b6f', 'source': 'https://api.kernelci.org/', 'type': 'api.kernelci.org', 'time': '2021-12-09T09:32:33.784288+00:00'}
```
Now, CloudEvent constructor will have two parameters : arguments & data.
In the arguments : A dictionary with type and source keys should be passed
In the data : A dictionary with message should be passed.

So I created event_data variable from the `attributes {'message': 'hii'}` functional argument. That leaves data argument unchanged in case we need to use it in the future.

I think we can drop the `if attributes is None:` condition and create another attributes dict with type and source from what we have received in `data {'specversion': '1.0', 'id': '8cf20531-f441-4ce3-b08d-02229f2b9b6f', 'source': 'https://api.kernelci.org/', 'type': 'api.kernelci.org', 'time': '2021-12-09T09:32:33.784288+00:00'}`

JenySadadia:
I have dropped client.py changes from this PR.
Added a few code changes in pubsub.py as well.

api/pubsub.py
```diff
@@ -70,10 +70,26 @@ async def publish(self, channel, message):
         await self._redis.publish(channel, message)
 
     async def publish_cloudevent(self, channel, data, attributes=None):
-        if attributes is None:
```
gctucker, line 73: Maybe the actual fix is to swap the `data` and `attributes` arguments here:
https://github.com/kernelci/kernelci-api/blob/main/api/main.py#L155

gctucker:
```python
await pubsub.publish_cloudevent(channel, attributes, data)
```

gctucker: This is not used by the proof-of-concept at the moment as all events are generated by the API when Node objects get created or updated.  So that would explain why it's currently broken, if only `client.py` or other ways to use the API make use of this.

JenySadadia: > Maybe the actual fix is to swap the `data` and `attributes` arguments here: https://github.com/kernelci/kernelci-api/blob/main/api/main.py#L155

Okay. Should I change the main.py and push the changes?

gctucker: Yes.  Having another patch to add a docstring in `publish_cloudevent()` is also useful.

gctucker:
Tested OK locally in conjunction with #49 using `client.py`.

The commit subject doesn't quite show what the commit is doing, something like `api.main: fix publish endpoint` would seem clearer.

JenySadadia:
Modified commit subject accordingly.

gctucker (approved):

PR #49, opened 2021-12-09
pub/sub events issue : listen

JenySadadia:
api/client.py : Fix Unprocessable Entity Exception for listen
Provided subscription ID instead of channel name

### DIFF
--- a/api/client.py
+++ b/api/client.py
@@ -38,9 +38,13 @@ def listen(args):
     print("Press Ctrl-C to stop.")
 
     try:
+        # Fetch subscription ID from /subscribe endpoint response
+        subscription_id = str(res.json()['id'])
+        # Pass subscription ID to /listen endpoint URL
+        listen_path = '/'.join(['listen', subscription_id])
+        url = urllib.parse.urljoin(args.url, listen_path)
+
         while True:
-            path = '/'.join(['listen', args.channel])
-            url = urllib.parse.urljoin(args.url, path)
             res = requests.get(url, headers=headers)
             res.raise_for_status()
             json_data = res.json().get('data')
@@ -51,8 +55,9 @@ def listen(args):
     except KeyboardInterrupt as e:
         print(f"Stopping.")
     finally:
-        path = '/'.join(['unsubscribe', args.channel])
-        url = urllib.parse.urljoin(args.url, path)
+        # Pass subscription ID to /unsubscribe endpoint URL
+        unsubscribe_path = '/'.join(['unsubscribe', subscription_id])
+        url = urllib.parse.urljoin(args.url, unsubscribe_path)
         res = requests.post(url, headers=headers)
         res.raise_for_status()
 


gctucker (commented):
The unsubscribe call should also use `subscription_id`, in the `finally` block.

JenySadadia:
> The unsubscribe call should also use `subscription_id`, in the `finally` block.

Pushed code changes for the unsubscribe endpoint.

api/client.py
```diff
@@ -51,7 +55,8 @@ def listen(args):
     except KeyboardInterrupt as e:
         print(f"Stopping.")
     finally:
-        path = '/'.join(['unsubscribe', args.channel])
+        # Pass subscription ID to /unsubscribe endpoint URL
+        path = '/'.join(['unsubscribe', str(res.json()['id'])])
```
gctucker, line 59: The `res` variable is also used in the `while True` loop, so we can't use it in `finally` as it will probably be redefined when listening for events.  Using subscription_id in `finally` would be clearer.

api/client.py
```diff
@@ -38,8 +38,12 @@ def listen(args):
     print("Press Ctrl-C to stop.")
 
     try:
+        # Fetch subscription ID from /subscribe endpoint response
+        subscription_id = res.json()['id']
```
gctucker, line 42:
```suggestion
        subscription_id = str(res.json()['id'])
        listen_path = /'.join(['listen', subscription_id])
```

api/client.py
```diff
@@ -38,8 +38,12 @@ def listen(args):
...
+
         while True:
-            path = '/'.join(['listen', args.channel])
+            # Pass subscription ID to /listen endpoint URL
+            path = '/'.join(['listen', str(subscription_id)])
```
gctucker, line 46:
```suggestion
```
We can move this outside of the `while` loop now.

api/client.py
```diff
@@ -38,8 +38,12 @@ def listen(args):
...
             url = urllib.parse.urljoin(args.url, path)
```
gctucker, line 47:
```suggestion
            url = urllib.parse.urljoin(args.url, listen_path)
```

api/client.py
gctucker, line 59:
```suggestion
        path = '/'.join(['unsubscribe', subscription_id])
```

gctucker (commented):
A couple of things to clean up.

JenySadadia:
Pushed code according to the suggested changes. @gctucker

api/client.py
```diff
@@ -38,9 +38,13 @@ def listen(args):
...
     try:
+        # Fetch subscription ID from /subscribe endpoint response
+        subscription_id = str(res.json()['id'])
+
         while True:
-            path = '/'.join(['listen', args.channel])
-            url = urllib.parse.urljoin(args.url, path)
+            # Pass subscription ID to /listen endpoint URL
+            listen_path = '/'.join(['listen', subscription_id])
+            url = urllib.parse.urljoin(args.url, listen_path)
```
gctucker, line 47: That should work, but what I was suggesting is that the `url` could be defined outside of the `while` loop since it never changes.  It's a very minor thing though.

JenySadadia: Yes, that makes sense.
I will move listen_path and ```url``` outside the ```while``` loop.

gctucker:
Tested OK locally in conjunction with #48 using `client.py`.

Like for #48, the commit subject should probably be something more like `api.client: fix call to the /listen endpoint`

JenySadadia:
Modified the commit subject accordingly.

gctucker (approved):

Issue #36, opened 2021-11-17
Initial API unit tests


gctucker:
This task is about setting up an environment and mechanism for automatically running unit tests for the KernelCI API.  That is to say, the code inside the `api` directory.

There are several ways to do this, so the first step is to determine how to bring up a test environment with Mongo DB and Redis (either stubs / mock interfaces or temporary data brought up with docker-compose).  We also need to determine which test framework to use, such as [pytest](https://docs.pytest.org/en/6.2.x/).  Then we should have at least one unit test to prove it works.

mgalka:
It's been decided that for kernelci-api testing we'll use [pytest](https://docs.pytest.org/en/6.2.x/). 
The result of this issue should be a simple unit test for the "[/](https://github.com/kernelci/kernelci-api/blob/15380acb93c292d7fe049cfc56cc2f7843b18956/api/main.py#L49)" endpoint of the API. This is a very simplistic test case which can be treated as a "hello world". Apart from the test case itself the `requirements-dev.txt` should be created with all packages required for testing added. There should be also instructions and if needed a script to run unit tests for `kernelci-api` project.

gctucker:
Moving this to the next Milestone as we've done the initial investigation about which framework to use but adding the actual unit tests seems outside the scope of the initial proof-of-concept.

PR #50, opened 2021-12-10
Initial API unit test case

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/36

test/test_api.py : Add API test for root endpoint
test/requirements-dev.txt : Add pytest


gctucker:
Note: It's true this wasn't a unit test so it's OK to close it now.  Having functional test using a full instance will be useful later, and we may implement them in a similar way as this PR.  But we can revisit that later.

PR #53, opened 2021-12-17
README.md: Renamed shell script used for testing env setup

JenySadadia:

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,9 +230,9 @@ has all the Python dependencies installed (essentially, `cloudevents`).
 
 Please follow below instructions to test API endpoints.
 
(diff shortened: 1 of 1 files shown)


gctucker:
Note: The convention for commit messages is to use the present tense, so it should be "rename" rather than "renamed".  But that's not a big issue, something to keep in mind for future commits.

JenySadadia:
> Note: The convention for commit messages is to use the present tense, so it should be "rename" rather than "renamed". But that's not a big issue, something to keep in mind for future commits.

Yes, will keep it in mind.
Thanks.

PR #55, opened 2021-12-18
Add endpoints to get node by attributes

JenySadadia:
README.md: Add section 'Node' to create and get node objects
api/db.py: Add find_by_attribute function
api/main.py: Add endpoints to get all added nodes,
node by attribute, node by sub-attribute of revision attribute.


api/main.py
```diff
@@ -101,6 +101,33 @@ async def get_node(node_id: str):
...
+
+
+@app.get('/node/revision/{attr_name}/{attr_value}', response_model=Node)
+async def get_node(attr_name: str, attr_value: str):
+    return await db.find_by_attribute(Node, "revision", attr_name, attr_value)
+
+
+@app.get('/node', response_model=Node)
+async def get_node(request: Request):
+    params = str(request.query_params)
```
gctucker, line 121: The `Request.query_params` attribute is a dictionary with values already parsed.  So there's no need to convert it into a string and then split it up again.  It's summarised here: https://stackoverflow.com/a/67412008

So I think the URL handler could just be:
```python
return await db.find_by_attributes(Node, dict(request.query_params))
```

api/main.py
```diff
@@ -101,6 +101,33 @@ async def get_node(node_id: str):
...
+async def get_node(attr_name: str, attr_value: str):
+    return await db.find_by_attribute(Node, attr_name, attr_value)
+
+
+@app.get('/node/revision/{attr_name}/{attr_value}', response_model=Node)
+async def get_node(attr_name: str, attr_value: str):
+    return await db.find_by_attribute(Node, "revision", attr_name, attr_value)
+
+
+@app.get('/node', response_model=Node)
```
gctucker, line 119: The response should be a list of `Node` objects, so I believe that should be reflected into the `response_model` value.  See the docs: https://fastapi.tiangolo.com/tutorial/response-model/

> It receives the same type you would declare for a Pydantic model attribute, so, it can be a Pydantic model, but it can also be, e.g. a list of Pydantic models, like List[Item].

api/main.py
```diff
@@ -101,6 +101,33 @@ async def get_node(node_id: str):
     return await db.find_by_id(Node, node_id)
 
 
+@app.get('/nodes')
+async def nodes():
+    return {'nodes': await db.find_all(Node)}
```
gctucker, line 106: I think we may add something like that later, with pagination and range parameters to only get a subset of all the objects in the returned value.  It's probably better to leave it out for now.

api/main.py
```diff
@@ -101,6 +101,33 @@ async def get_node(node_id: str):
...
+
+
+@app.get('/node/{attr_name}/{attr_value}', response_model=Node)
+async def get_node(attr_name: str, attr_value: str):
+    return await db.find_by_attribute(Node, attr_name, attr_value)
+
+
+@app.get('/node/revision/{attr_name}/{attr_value}', response_model=Node)
+async def get_node(attr_name: str, attr_value: str):
+    return await db.find_by_attribute(Node, "revision", attr_name, attr_value)
```
gctucker, line 116: This is a bit too specific to `Node.revision`, we can use the dotted syntax instead e.g. `revision.attr=value`.  In fact we may be able to pass it as-is when doing a Mongo DB query as it's supported as well.  For example:
```
> db.node.findOne({'revision.describe': 'v5.16-rc5-224-g9eaa88c7036e'})
{
	"_id" : ObjectId("61bdb69baddf9068ffcb7b37"),
	"kind" : "node",
	"name" : "checkout",
	"revision" : {
		"tree" : "mainline",
		"url" : "https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git",
		"branch" : "master",
		"commit" : "9eaa88c7036eda3f6c215f87ca693594cf90559b",
		"describe" : "v5.16-rc5-224-g9eaa88c7036e"
	},
	"parent" : null,
	"status" : true
}
```

gctucker:
From the commit subject:

> README.md: Add section 'Node' to create and get node objects
> api/db.py: Add find_by_attribute function
> api/main.py: Add endpoints to get all added nodes,

I think this shows there could actually be 3 commits, one for each logical change.  For example, first you makde the change in `api.db` to implement `find_by_attributes()`, then a second commit is to implement the URL handler in `api.main`, and a third commit is to update the README.md.

JenySadadia:
Done code changes as per feedback. 
Separated into 3 commits.

api/db.py
```diff
@@ -42,6 +42,14 @@ async def find_by_id(self, model, obj_id):
         obj = await col.find_one({'_id': ObjectId(obj_id)})
         return model(**obj) if obj else None
 
+    async def find_by_attribute(self, model, query_list):
```
gctucker, line 45:
```suggestion
    async def find_by_attributes(self, model, query_list):
```

api/main.py
```diff
@@ -101,6 +102,16 @@ async def get_node(node_id: str):
     return await db.find_by_id(Node, node_id)
 
 
+@app.get('/nodes')
+async def nodes():
+    return {'nodes': await db.find_all(Node)}
+
+
+@app.get('/node', response_model=List[Node])
```
gctucker, line 110: In fact I guess this should be `/nodes` since it's returning a number of Node objects rather than just one.

api/main.py
```diff
@@ -101,6 +102,16 @@ async def get_node(node_id: str):
...
+async def get_node(request: Request):
+    return await db.find_by_attribute(Node, dict(request.query_params))
```
gctucker, line 112:
```suggestion
    return await db.find_by_attributes(Node, dict(request.query_params))
```

api/db.py
```diff
@@ -42,6 +42,14 @@ async def find_by_id(self, model, obj_id):
...
+        col = self._get_collection(model)
+        objs = []
+        data = col.find(query_list)
+        async for obj in data:
+            objs.append(model(**obj))
```
gctucker, line 50: This should be using `await` rather than `async`, as `async` is to declare functions that will be running asynchronous commands whereas `await` is when calling code that runs asynchronous commands.

Maybe the syntax could be simplified using list comprehension (I haven't tried to run this):
```python
data = col.find(attributes)
return list(await model(**obj) for obj in data)
```

api/db.py
gctucker, line 45:
```suggestion
    async def find_by_attribute(self, model, attributes):
```

gctucker: The argument passed from the query is a dictionary with attribute key/value pairs, not a list.  So I think `attributes` is a better name, also we should be adding docstrings to clarify how this works.

gctucker:
> Done code changes as per feedback. Separated into 3 commits.

There's a problem between commits 1 and 2, commit one includes some changes in `api/main.py` which are then undone by commit 2.

api/main.py
gctucker: In fact this probably does the same as the handler above with no query parameters.  I think we should drop this, what do you think?

JenySadadia: Yes, I tested it. We can drop this.

api/db.py
JenySadadia: I used ```return await data.to_list(None)``` for it.

JenySadadia:
We need to have a separate endpoint to get node by id. We can not use /nodes with attributes endpoint for it. 
Because it needs to use ObjectId(obj_id) rather than just attribute value. 

```obj = await col.find_one({'_id': ObjectId(obj_id)}) ```

What do you suggest on this? @gctucker

**Comment by gctucker** [comment]
> We need to have a separate endpoint to get node by id. We can not use /nodes with attributes endpoint for it. Because it needs to use ObjectId(obj_id) rather than just attribute value.
> 
> `obj = await col.find_one({'_id': ObjectId(obj_id)}) `
> 
> What do you suggest on this? @gctucker

Well then we keep `async def get_node(node_id: str):` for retrieving a Node object using its id, and we can mention in the documentation for the `get_nodes_by_attributes()` endpoint that it's not meant to be used with object ids.  If it becomes an issue, we could replace any `id` value in the query dictionary with `ObjectId` in kernelci.db, but I don't think we need to do that now and it's probably better to keep the code as simple as possible initially.

File: api/db.py (hunk 8 above)
Comment (gctucker) [reply api/db.py]: OK, but that's not the same as this:
```python
```
The returned value from `data.to_list(None)` is a list of dictionaries, whereas the code above returns a list of `Node` objects.  The `Node` objects then get converted to JSON by `api.main` in the same way as the dictionaries so the data returned by the API looks the same, but it's not quite the same from an implementation point of view.  The `find_by_id()` method returns a `Node` object using `return model(**obj)` so I think it would be better to stay consistent with the `find_by_attributes()` method.  Other parts of the code may be using `api.db` and expect actual `Node` objects rather than dictionaries.

JenySadadia:
> > We need to have a separate endpoint to get node by id. We can not use /nodes with attributes endpoint for it. Because it needs to use ObjectId(obj_id) rather than just attribute value.
> > `obj = await col.find_one({'_id': ObjectId(obj_id)}) `
> > What do you suggest on this? @gctucker
> 
> Well then we keep `async def get_node(node_id: str):` for retrieving a Node object using its id, and we can mention in the documentation for the `get_nodes_by_attributes()` endpoint that it's not meant to be used with object ids. If it becomes an issue, we could replace any `id` value in the query dictionary with `ObjectId` in kernelci.db, but I don't think we need to do that now and it's probably better to keep the code as simple as possible initially.

Done modifying the README.md file.

api/db.py
JenySadadia: Yes, makes sense.

JenySadadia:
Pushed after rebasing on top of the main.

gctucker (commented):
Tested OK locally:
```
$ curl "http://localhost:8001/nodes?name=checkout&revision.describe=v5.16-rc6-46-g3f066e882bf1"
[{"_id":"61c37cc25b81dd46486b4d48","kind":"node","name":"checkout","revision":{"tree":"mainline","url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git","branch":"master","commit":"3f066e882bf1373db023ada606e0dca1af375d9b","describe":"v5.16-rc6-46-g3f066e882bf1"},"parent":null,"status":true},{"_id":"61c37d125b81dd46486b4d4a","kind":"node","name":"checkout","revision":{"tree":"mainline","url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git","branch":"master","commit":"3f066e882bf1373db023ada606e0dca1af375d9b","describe":"v5.16-rc6-46-g3f066e882bf1"},"parent":null,"status":true},{"_id":"61c37dee5b81dd46486b4d4c","kind":"node","name":"checkout","revision":{"tree":"mainline","url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git","branch":"master","commit":"3f066e882bf1373db023ada606e0dca1af375d9b","describe":"v5.16-rc6-46-g3f066e882bf1"},"parent":null,"status":true}]
```

gctucker:
Please also add a description of what each change does with commit messages.  This is to make it possible to follow the changes just by looking a `git log` and without looking at the diff.  Things that seem obvious now might not be in a few months' time, or people not familiar with the code won't find the history useful with empty commit messages.

gctucker (approved):

gctucker:
LGTM

A couple of notes, about things to do next:

We will need to add unit tests for this endpoint, like all endpoints.  While trying things out by hand, I noticed that `status` being a boolean didn't work as it was not converted from a string.  Also the `parent_id` didn't work for similar reasons, it would need to be converted to an `ObjectId` for Mongo DB.  I'm not sure yet how to solve the issue of type conversion, I guess this could be done in the object models.  

Also we can probably drop the `Thing` model and associated endpoints now that we have real `Node` objects.

PR #59, opened 2021-12-20
Run Pylint on api.models using GitHub actions

gctucker:
Add a check in the GitHub actions to run Pylint on `api.models` and fix all the errors reported by it so the score is 10.0.  Ignore warnings about the number of public methods on classes as model classes don't have any methods by design.

Other modules in the `api` package should be added as a follow-up in the same way.


gctucker:
Tested OK using `kernelci-pipeline`.

Issue #16, opened 2021-10-26
Add SSH container to upload files to storage


gctucker:
Once #15 is complete and there is a container to serve files from storage using HTTP, another container can be added with an SSH server to upload files remotely.  This may not be very useful for a local setup where files can just be copied to the `storage/data` directory, but it will be required when deploying the system on a remote server.

Steps:
* #19 
  * Add a `docker/ssh/Dockerfile` based on `debian:bullseye-slim` with `openssh` installed
  * Create a `kernelci` user as in `docker/api/Dockerfile`
* Add a `docker/ssh/authorized_keys` file and add a command to copy it to `/home/kernelci/.ssh` in the `Dockerfile`
* Copy the `/etc/ssh/sshd_config` file from the Docker image into `docker/ssh/sshd_config` and add it as well in the `Dockerfile` so we can keep config changes under source control
* Add an `ssh` section to [`docker-compose`](https://github.com/kernelci/kernelci-api/blob/main/docker/docker-compose.yaml) to create a container using the files available in `docker/ssh`
* Add a volume entry to mount the `storage/data` directory in the `ssh` container in `/home/kernelci/data`

Note: The `authorized_keys` file should not be kept under source control as it will contain the list of authorized SSH keys for each deployment of the container.  Instead, the documentation (or README) should have instructions on how to create an SSH key with `ssh-keygen` and add the public key to the file.

gctucker:
The first PR was merged and GitHub closed this issue automatically, but more things need to be done so re-opening it now.

PR #60, opened 2021-12-21
Add SSH container

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/16

docker/ssh/Dockerfile: Add command to copy sshd_config file
docker/docker-compose.yaml: Add ssh section
docker/ssh: Add authorized_keys.sample file
README.md: Add 'Generate SSH key' section

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,3 +245,22 @@ Run below command from kernelci-api/api directory:
 pytest
 ```
 This will start running test cases from kernelci-api/api/test_api.py and display results.
+
+## Generate SSH key and copy to SSH directory
+
+Use below command to generate SSH key. It will store private key to `/home/username/.ssh/id_rsa_kernelci-api` file and public key to `/home/username/.ssh/id_rsa_kernelci-api.pub` file.
+
+```
+$ ssh-keygen -f ~/.ssh/id_rsa_kernelci-api
+```
+
+Use below command to copy the public key to ssh/user-data directory.
+```
+$ cat ~/.ssh/id_rsa_kernelci-api.pub >> docker/ssh/user-data/authorized_keys
+```
+
+SSH docker container will have /home/kernelci/.ssh/authorized_keys file.
+Now, the user will able to SSH to container using private key.
+```
+$ ssh -i ~/.ssh/id_rsa_kernelci-api -p 8022 kernelci@localhost
+```
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -37,5 +37,15 @@ services:
     ports:
       - 8002:80
 
+  ssh:
+    container_name: 'kernelci-ssh'
+    build:
+      context: ssh
+    volumes:
+      - './storage/data:/home/kernelci/data'
+      - './ssh/user-data:/home/kernelci/.ssh'
+    ports:
+      - '8022:22'
+
 volumes:
   mongodata:
--- a/docker/ssh/Dockerfile
+++ b/docker/ssh/Dockerfile
@@ -20,5 +20,8 @@ RUN chown kernelci: /home/kernelci
 # SSH server will run on PORT 22
 EXPOSE 22
 
+# Copy sshd_config file
+COPY sshd_config /etc/ssh/
+
 # Run SSH server in detached mode
 CMD ["/usr/sbin/sshd","-D"]
--- a/docker/ssh/sshd_config
+++ b/docker/ssh/sshd_config
@@ -1,0 +1,123 @@
+#	$OpenBSD: sshd_config,v 1.103 2018/04/09 20:41:22 tj Exp $
+
+# This is the sshd server system-wide configuration file.  See
+# sshd_config(5) for more information.
+
+# This sshd was compiled with PATH=/usr/bin:/bin:/usr/sbin:/sbin
+
+# The strategy used for options in the default sshd_config shipped with
+# OpenSSH is to specify options with their default value where
+# possible, but leave them commented.  Uncommented options override the
+# default value.
+
+Include /etc/ssh/sshd_config.d/*.conf
+
+#Port 22
+#AddressFamily any
+#ListenAddress 0.0.0.0
+#ListenAddress ::
+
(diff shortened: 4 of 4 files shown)


docker/ssh/sshd_config
```diff
@@ -0,0 +1,123 @@
...
+# sshd_config(5) for more information.
+
+# This sshd was compiled with PATH=/usr/bin:/bin:/usr/sbin:/sbin
+
+# The strategy used for options in the default sshd_config shipped with
+# OpenSSH is to specify options with their default value where
+# possible, but leave them commented.  Uncommented options override the
+# default value.
+
+Include /etc/ssh/sshd_config.d/*.conf
```
gctucker, line 13: Actually, I guess another option would be to have another file copied to this directory as it should override things in the main `sshd_config`.  I think we can keep this approach of editing the file for now, it will depend on the kind of changes we want to make to the configuration.

docker/ssh/Dockerfile
```diff
@@ -20,5 +20,11 @@ RUN chown kernelci: /home/kernelci
 # SSH server will run on PORT 22
 EXPOSE 22
 
+# Copy authorized_keys file
+COPY authorized_keys /home/kernelci/.ssh
```
gctucker, line 24: We might need to run `mkdir -p /home/kernelci/.ssh` first as I don't think the directory will be created automatically by `useradd`.  If we don't create it, then the `authorized_keys` file will be copied as a `.ssh` file.

Another approach is to add a volume for the `.ssh` directory with the `authorized_keys` file provided by the host, maybe that would be more flexible.  It means the container could be created even if there isn't an `authorized_keys` file as it can be added later while the container is running.

mgalka: I like the latter approach more as it makes the authorized_keys file persistent.

docker/docker-compose.yaml
```diff
@@ -37,5 +37,13 @@ services:
     ports:
       - 8002:80
 
+  ssh:
+    container_name: 'kernelci-ssh'
+    build:
+      context: ssh
+    volumes:
+      - './storage/data:/home/kernelci/data'
+      - './ssh:/home/kernelci/.ssh'
```
gctucker, line 46: This works, but we have files there which aren't really needed:
```
$ docker-compose run -u kernelci ssh /bin/bash
kernelci@222ff8a81d96:~$ ls -l /home/kernelci/.ssh
total 8
-rw-r--r-- 1 kernelci kernelci  652 Dec 23 09:53 Dockerfile
-rw-r--r-- 1 kernelci kernelci    0 Dec 23 09:53 authorized_keys.sample
-rw-r--r-- 1 kernelci kernelci 3289 Dec 23 09:53 sshd_config
```
So maybe we should have a sub-directory, for example `docker/ssh/user-data` and then mount that in `.ssh`:
```yaml
  ssh:
    container_name: 'kernelci-ssh'
    build:
      context: ssh
    volumes:
      - './storage/data:/home/kernelci/data'
      - './ssh/user-data:/home/kernelci/.ssh'
```
The `authorized_keys.sample` file would then be in that sub-directory.

gctucker, line 46:
```suggestion
      - './ssh/user-data:/home/kernelci/.ssh'
```

JenySadadia: Oh! I forgot to commit that change.
Just did.

gctucker (approved):
Tested OK manually:
```
$ docker-compose exec -u kernelci ssh /bin/bash
kernelci@b98946ed0e9b:/$ cd
kernelci@b98946ed0e9b:~$ ls -la .ssh
total 8
drwxr-xr-x 2 kernelci kernelci 4096 Dec 24 10:49 .
drwxr-xr-x 4 kernelci kernelci 4096 Dec 24 10:54 ..
-rw-r--r-- 1 kernelci kernelci    0 Dec 24 10:49 authorized_keys.sample
```

README.md
```diff
@@ -245,3 +245,11 @@ Run below command from kernelci-api/api directory:
...
 ```
 This will start running test cases from kernelci-api/api/test_api.py and display results.
+
+## Generate SSH key
+
+Use below command to generate SSH key. It will store private key to `authorized_keys` file and public key to `authorized_keys.pub` file.
+
+```
+$ ssh-keygen -f authorized_keys
+```
```
gctucker, line 255: Actually that's not really how `authorized_keys` is meant to be used: the private key should stay on the host outside of Docker, and the public key should be listed in `authorized_keys`.  For example:
```
$ ssh-keygen -f ~/.ssh/id_rsa_kernelci-api
$ cat ~/.ssh/id_rsa_kernelci-api.pub >> docker/ssh/user-data/authorized_keys
```
Then the user should be able to ssh to the running Docker container using:
```
$ ssh -i ~/.ssh/id_rsa_kernelci-api -p 8022 kernelci@localhost
Linux b57d306e7629 4.19.0-10-amd64 #1 SMP Debian 4.19.132-1 (2020-07-24) x86_64

The programs included with the Debian GNU/Linux system are free software;
the exact distribution terms for each program are described in the
individual files in /usr/share/doc/*/copyright.

Debian GNU/Linux comes with ABSOLUTELY NO WARRANTY, to the extent
permitted by applicable law.
kernelci@b57d306e7629:~$ 
```
with this added change to export the container's port 22 to the host as 8022 (arbitrary number, related to adding parameters for port numbers):
```diff
diff --git a/docker/docker-compose.yaml b/docker/docker-compose.yaml
index 65d7b729fcb7..15b1f5b60b89 100644
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -44,6 +44,8 @@ services:
     volumes:
       - './storage/data:/home/kernelci/data'
       - './ssh/user-data:/home/kernelci/.ssh'
+    ports:
+      - '8022:22'
 
 volumes:
   mongodata:
```

gctucker (commented):
As per my last comment, I think we should a `ports` attribute to make the SSH server accessible from the host.  Also the README.md should be tweaked a bit.

gctucker (approved):

PR #61, opened 2021-12-21
Fix GitHub actions running checks

gctucker:
Use `actions/setup-python` to install the actual Python version required, and reduce to only testing with Python 3.10 as it's the only version that actually works with all the current requirements.  Also consolidate how requirements-dev.txt is used to simplify things.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,8 +36,7 @@ jobs:
         working-directory: docker/api
         run: |
           python -m pip install --upgrade pip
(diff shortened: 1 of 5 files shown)


mgalka (approved):

JenySadadia:
Need to modify ```API Testing``` section in README.md as shell scripts have been removed in this PR.

gctucker:
> Need to modify `API Testing` section in README.md as shell scripts have been removed in this PR.

Thanks, I've just added a commit for it.

PR #62, opened 2021-12-22
api/main.py: Fix undefined 'channel' variable in listen endpoint

JenySadadia:

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -143,7 +143,7 @@ async def listen(sub_id: int, user: User = Depends(get_user)):
     if msg is None:
         raise HTTPException(
             status_code=status.HTTP_404_NOT_FOUND,
-            detail=f"Not subscribed to channel: {channel}"
+            detail=f"Not subscribed to channel with id: {sub_id}"
         )
     return msg
 


gctucker (approved):
Thanks for the fix, LGTM.

Tip: You can add a `Fixes:` trailer in the commit message when fixing an issue introduced by an earlier commit.  For example, in this case:
```
Fixes: ea56015f09b9 ("api: add pubsub and handlers in main")
```
The commit could be found using `git blame`, for example:
```
$ git blame main api/main.py | grep "Not subscribed"
ea56015f09b98 (Guillaume Tucker 2021-10-18 22:44:51 +0100 146)             detail=f"Not subscribed to channel: {channel}"
```
Then to have a one-line description:
```
$ git show --oneline ea56015f09b98 | head -1
ea56015f09b9 api: add pubsub and handlers in main
```
with the convention to add '("   ")` around the commit subject as some standard tools rely on that to parse the Git log.

Well this sounds like something else to add to a "Git How-To" documentation page :)

JenySadadia:
Thanks for the information @gctucker 
I will use this convention for future commits.

PR #63, opened 2021-12-22
docker: move docker-compose.yaml file to root directory

gctucker:
Move the docker-compose.yaml file to the root directory of the project
so it can be used more easily.  This makes it more consistent with
kernelci-pipeline too.  Update the README.md accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ The FastAPI server and Mongo DB can all be started at the same time using
 the `db` hostname.
 
 ```
-$ docker-compose -f docker/docker-compose.yaml up --build
+$ docker-compose up --build
 [...]
 kernelci-api | INFO:     Application startup complete.
 ```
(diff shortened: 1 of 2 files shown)


gctucker:
Note: If there are errors with conflicting container names, run `docker system prune` and then `docker-compose up --build` should work.

README.md
```diff
@@ -53,13 +53,16 @@ use it with other tools: http://localhost:8001/openapi.json
...
 ```
 $ openssl rand -hex 32
 ```
-Store the generated key to .env file and give its path to docker-compose.yaml file.
-We have loaded secret key using FastAPI settings feature in api.auth
-Please find the env.sample file in the base directory to store secret key and copy the file to your .env file.
+
+Store the generated key to `.env` file and give its path to
+`docker-compose.yaml` file.  We have loaded secret key using FastAPI settings
+feature in api.auth Please find the env.sample file in the base directory to
```
JenySadadia, line 64: Suggested change:

```feature in api.auth. Please find the env.sample file in the base directory to```

JenySadadia:
@gctucker I would suggest a minor change in README, otherwise tested OK locally.

PR #65, opened 2021-12-22
api.auth: fix all Pylint errors

gctucker:
Fix all Pylint errors in api.auth and enable it in GitHub checks.


gctucker (commented):
Tested OK via GitHub checks and locally with kernelci-pipeline.

Issue #67, opened 2021-12-28
Add /token endpoint test for negative path


mgalka:
Current test for the /token endpoint covers only the positive path. It doesn't test the case when user doesn't get authenticated.

We need to add separate test case to cover this part of the `/token` endpoint handler.
It should check if a proper exception is raised.

PR #72, opened 2021-12-28
api/test_api.py: Add /token endpoint test for negative path

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/67

 Negative testing of /token endpoint by providing incorrect
password.

### DIFF
--- a/api/test_api.py
+++ b/api/test_api.py
@@ -96,3 +96,33 @@ def test_me_endpoint(mock_get_current_user):
     assert response.status_code == 200
     assert ('_id' and 'username' and 'hashed_password' and
             'active') in response.json()
+
+
+def test_token_endpoint_incorrect_password(mock_db_find_one):
+    """
+    Test Case : Test KernelCI API token endpoint for negative path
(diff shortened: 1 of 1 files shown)


api/test_api.py
```diff
@@ -96,3 +96,33 @@ def test_me_endpoint(mock_get_current_user):
     assert response.status_code == 200
     assert ('_id' and 'username' and 'hashed_password' and
             'active') in response.json()
+
+
+def test_token_endpoint_negative_path(mock_db_find_one):
```
mgalka, line 101: I'd rather choose more informative name. like `test_token_endpoint_wrong_password`, there can be more than one "negative" path to test in future. It also simplifies reading test logs when the test case names are more descriptive/

mgalka (changes requested):

api/test_api.py
JenySadadia: Renamed the function.

mgalka (approved):

Issue #68, opened 2021-12-28
Add /subscribe endpoint test


mgalka:
Add unit test test for `/subscribe` endpoint.

The test should mock the pubsub.subscribe method

PR #73, opened 2021-12-29
api/test_api.py: Add /subscribe endpoint test

JenySadadia:
Fixes: https://github.com/kernelci/kernelci-api/issues/68

api/test_api.py: Add fixture to mock PubSub._init_sub_id
api/test_api.py: Add fixture to mock PubSub.subscribe
api/test_api.py:  Add test_subscribe_endpoint

### DIFF
--- a/api/test_api.py
+++ b/api/test_api.py
@@ -134,3 +134,11 @@ def mock_init_sub_id(mocker):
     mocker.patch('api.pubsub.PubSub._init_sub_id',
                  side_effect=async_mock)
     return async_mock
+
+
+@pytest.fixture()
+def mock_subscribe(mocker):
+    async_mock = AsyncMock()
+    mocker.patch('api.pubsub.PubSub.subscribe',
+                 side_effect=async_mock)
+    return async_mock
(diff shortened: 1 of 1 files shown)


api/test_api.py
```diff
@@ -96,3 +97,52 @@ def test_me_endpoint(mock_get_current_user):
...
+                hashed_password='$2b$12$CpJZx5ooxM11bCFXT76/z.o6HWs2sPJy4iP8.'
+                                'xCZGmM8jWXUXJZ4K',
+                active=True)
+    mock_get_current_user.return_value = user
+
+    subscribe = Subscription(id=1, channel='abc')
+    mock_subscribe.return_value = subscribe
+
+    with TestClient(app) as client:
+        # Used 'TestClient' with a 'with' to trigger startup event on app
```
mgalka, line 136: It may be more accurate to use an actual name of the language feature i.e. `context manager` instead of `'with'`
e.g.
> Use context manager to trigger a startup event on the app object

mgalka (changes requested):

api/test_api.py
JenySadadia: Modified the comment.

mgalka (approved):

PR #76, opened 2021-12-31
api/test_api.py:add test cases for listen endpoint

### LINKED ISSUE #70: Add /listen endpoint test
Add unittest for /listen endpoint.

The test should mock the pubsub.listen method.
It should test both, the positive and the negative path.

alexandrasp:
Fixes #70 

Include listen endpoint test cases for positive and negative cases, respectively with answers 200 and 404.

Signed-off-by: Alexandra Pereira <alexandra.pereira@collabora.com>

### DIFF
--- a/api/test_api.py
+++ b/api/test_api.py
@@ -606,3 +606,49 @@ def test_get_all_nodes_empty_response(mock_get_current_user,
         print("response.json()", response.json())
         assert response.status_code == 200
         assert len(response.json()) == 0
+
+
+@pytest.fixture()
+def mock_init_sub_id(mocker):
+    async_mock = AsyncMock()
+    mocker.patch('api.pubsub.PubSub._init_sub_id',
+                 side_effect=async_mock)
+    return async_mock
+
+
+@pytest.fixture()
+def mock_listen(mocker):
+    async_mock = AsyncMock()
+    mocker.patch('api.pubsub.PubSub.listen',
+                 side_effect=async_mock)
+    return async_mock
+
+
+def test_listen_endpoint(mock_get_current_user,
+                         mock_init_sub_id,
+                         mock_listen):
+    """
+    Test Case : Test KernelCI API GET /listen endpoint for the
+    positive path
+    Expected Result :
+        HTTP Response Code 200 OK
+        Listen for events on a channel.
+    """
+    user = User(username='bob',
+                hashed_password='$2b$12$CpJZx5ooxM11bCFXT76/z.o6HWs2sPJy4iP8.'
+                                'xCZGmM8jWXUXJZ4K',
+                active=True)
+    mock_get_current_user.return_value = user
+    mock_listen.return_value = 'Listening for events on channel 1'
+    with TestClient(app) as client:
+        response = client.get(
+            "/listen/1",
+            headers={
+                "Accept": "application/json",
+                "Authorization": "Bearer "
+                "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9."
+                "eyJzdWIiOiJib2IifQ.ci1smeJeuX779PptTkuaG1S"
+                "Edkp5M1S1AgYvX8VdB20"
+            },
+        )
+        assert response.status_code == 200
--- a/api/test_api.py
+++ b/api/test_api.py
@@ -652,3 +652,29 @@ def test_listen_endpoint(mock_get_current_user,
             },
         )
         assert response.status_code == 200
+
+
+def test_listen_endpoint_not_found(mock_get_current_user,
+                                   mock_init_sub_id,
+                                   mock_listen):
+    """
+    Test Case : Test KernelCI API GET /listen endpoint for the
+    negative path
+    Expected Result :
+        HTTP Response Code 404 Not Found
+        Channel not available or not subscribed to channel with an id
+    """
+    mock_listen.return_value = None
+
+    with TestClient(app) as client:
+        response = client.get(
+            "/listen/1",
+            headers={
+                "Accept": "application/json",
+                "Authorization": "Bearer "
+                "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9."
+                "eyJzdWIiOiJib2IifQ.ci1smeJeuX779PptTkuaG1S"
+                "Edkp5M1S1AgYvX8VdB20"
+            },
+        )
+        assert response.status_code == 404
--- a/api/test_api.py
+++ b/api/test_api.py
@@ -678,3 +678,29 @@ def test_listen_endpoint_not_found(mock_get_current_user,
             },
         )
         assert response.status_code == 404
+
+
+def test_listen_endpoint_without_token(mock_get_current_user,
+                                       mock_init_sub_id,
+                                       mock_listen):
+    """
+    Test Case : Test KernelCI API GET /listen endpoint for the
+    negative path
+    Expected Result :
+        HTTP Response Code 401 Unauthorized
+        The request requires user authentication by token in header
+    """
+    user = User(username='bob',
+                hashed_password='$2b$12$CpJZx5ooxM11bCFXT76/z.o6HWs2sPJy4iP8.'
+                                'xCZGmM8jWXUXJZ4K',
+                active=True)
+    mock_get_current_user.return_value = user
+    mock_listen.return_value = 'Listening for events on channel 1'
+    with TestClient(app) as client:
+        response = client.get(
+            "/listen/1",
+            headers={
+                "Accept": "application/json"
+            },
+        )
+        assert response.status_code == 401


api/test_api.py
```diff
@@ -96,3 +96,60 @@ def test_me_endpoint(mock_get_current_user):
...
+
+
+def test_listen_positive(mock_get_current_user, mock_init_sub_id, mock_listen):
+
+    user = User(username='bob',
+                hashed_password='$2b$12$CpJZx5ooxM11bCFXT76/z.o6HWs2sPJy4iP8.'
+                                'xCZGmM8jWXUXJZ4K',
+                active=True)
+    mock_get_current_user.return_value = user
+    mock_listen.return_value = 'String'
```
JenySadadia, line 124: I think we should return the more relevant string message for `mock_listen.return_value` instead of `String`. 
Something like what the actual /listen endpoint would return  `Listening for events on channel 1`

JenySadadia:
It is a good practice to split a PR into various commits if PR has multiple functions implemented.
Here, you can add separate commits for `test_listen_positive` and `test_listen_negative` instead of just a single commit.

api/test_api.py
mgalka: I agree, I'd rather have something more informative there

mgalka (changes requested):
@alexandrasp , Could you rebase your PR with the current kernelci-api/main?

api/test_api.py
```diff
@@ -606,3 +606,59 @@ def test_get_all_nodes_empty_response(mock_get_current_user,
...
+
+@pytest.fixture()
+def mock_listen(mocker):
+    async_mock = AsyncMock()
+    mocker.patch('api.pubsub.PubSub.listen',
+                 side_effect=async_mock)
+    return async_mock
+
+
+def test_listen_positive(mock_get_current_user, mock_init_sub_id, mock_listen):
```
JenySadadia, line 627: Rename the function from `test_listen_positive` to `test_listen_endpoint`.

api/test_api.py
```diff
@@ -606,3 +606,59 @@ def test_get_all_nodes_empty_response(mock_get_current_user,
...
+                "Authorization": "Bearer "
+                "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9."
+                "eyJzdWIiOiJib2IifQ.ci1smeJeuX779PptTkuaG1S"
+                "Edkp5M1S1AgYvX8VdB20"
+            },
+        )
+        assert response.status_code == 200
+
+
+def test_listen_negative(mock_get_current_user, mock_init_sub_id, mock_listen):
```
JenySadadia, line 649: Rename the function from `test_listen_negative` to `test_listen_endpoint_not_found`.

JenySadadia:
I think we should also cover a test case with a negative path when the user has passed an incorrect bearer token to the /listen endpoint.

api/test_api.py
```diff
@@ -606,3 +606,63 @@ def test_get_all_nodes_empty_response(mock_get_current_user,
...
+    async_mock = AsyncMock()
+    mocker.patch('api.pubsub.PubSub.listen',
+                 side_effect=async_mock)
+    return async_mock
+
+
+def test_listen_endpoint(mock_get_current_user,
+                         mock_init_sub_id,
+                         mock_listen):
+
```
mgalka, line 630: Please add docstrings

api/test_api.py
```diff
@@ -606,3 +606,63 @@ def test_get_all_nodes_empty_response(mock_get_current_user,
...
+                "Edkp5M1S1AgYvX8VdB20"
+            },
+        )
+        assert response.status_code == 200
+
+
+def test_listen_endpoint_not_found(mock_get_current_user,
+                                   mock_init_sub_id,
+                                   mock_listen):
+
```
mgalka, line 654: Please, add docstring

mgalka (changes requested):

gctucker:
I guess we'll also need to have this PR merged before we start moving tests to the `tests` directory.

api/test_api.py
alexandrasp: done

api/test_api.py
alexandrasp: done

api/test_api.py
```diff
@@ -606,3 +606,100 @@ def test_get_all_nodes_empty_response(mock_get_current_user,
...
+    mock_listen.return_value = None
+
+
+    with TestClient(app) as client:
+        response = client.get(
+            "/listen/1",
+            headers={
+                "Accept": "application/json",
+                "Content-Type": "application/x-www-form-urlencoded"
+            },
```
alexandrasp, line 702: @JenySadadia did you mean a test like this?

JenySadadia: Actually, you need to pass the Bearer token in the header. The token will not match with the user returned by `mock_get_current_user`. Hence it will respond with `401 Unauthorized` error.
You also need to rename this test function. It is a duplicate name, matches the previous function's name.

JenySadadia (changes requested):
Actually, you need to pass the Bearer token in the header. The token will not match with the user returned by mock_get_current_user. Hence it will respond with 401 Unauthorized error.
You also need to rename this test function. It is a duplicate name, matches the previous function's name.

alexandrasp:
> Actually, you need to pass the Bearer token in the header. The token will not match with the user returned by `mock_get_current_user`. Hence it will respond with `401 Unauthorized error`. You also need to rename this test function. It is a duplicate name, matches the previous function's name.

@JenySadadia I've tried this that you suggested, but it is not working as expected returning `401`,  it is returning `200`. I also checked the docs available and it doesn't help. Do you have any idea about what I could try? Maybe @gctucker can give  some input as well.

JenySadadia:
> > Actually, you need to pass the Bearer token in the header. The token will not match with the user returned by `mock_get_current_user`. Hence it will respond with `401 Unauthorized error`. You also need to rename this test function. It is a duplicate name, matches the previous function's name.
> 
> @JenySadadia I've tried this that you suggested, but it is not working as expected returning `401`, it is returning `200`. I also checked the docs available and it doesn't help. Do you have any idea about what I could try? Maybe @gctucker can give some input as well.
> 
> Additionally, I tried to check how this mock handle authentication and related things without success yet.

Hey, when you pass incorrect TOKEN in the request, you will receive 401 error.
```
docker-compose exec api /bin/sh -c '\
TOKEN=anc \
/usr/bin/env python3 /home/kernelci/api/client.py \
listen abc'
Traceback (most recent call last):
  File "/home/kernelci/api/client.py", line 83, in <module>
    args.func(args)
  File "/home/kernelci/api/client.py", line 35, in listen
    res.raise_for_status()
  File "/home/kernelci/.local/lib/python3.10/site-packages/requests/models.py", line 960, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: http://localhost:8000/subscribe/abc
```
You need to cover this scenario in your test case. 

For the correct TOKEN, we will get logs like below:
```
docker-compose exec api /bin/sh -c '\
TOKEN=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJib2IifQ.tYsu5nu2njDzrkrf_sVK1BRzCL6pUuop-YQuFnZ2xLo \
/usr/bin/env python3 /home/kernelci/api/client.py \
listen abc'
Listening for events on channel abc.
Press Ctrl-C to stop.
```
Please feel free to ping me if you still have any doubts about this.

api/test_api.py
```diff
@@ -606,3 +606,103 @@ def test_get_all_nodes_empty_response(mock_get_current_user,
...
+    mock_get_current_user.return_value = user
+    mock_listen.return_value = 'Listening for events on channel 1'
+    with TestClient(app) as client:
+        response = client.get(
+            "/listen/1",
+            headers={
+                "Accept": "application/json",
+                "Content-Type": "application/x-www-form-urlencoded"
+            },
+            data={'username': 'bob', 'password': 'abc'}
```
JenySadadia, line 706: Remove `data` dictionary from here. As we don't need to pass username and password. Only Token would work for this endpoint

api/test_api.py
```diff
@@ -606,3 +606,103 @@ def test_get_all_nodes_empty_response(mock_get_current_user,
...
+                                'xCZGmM8jWXUXJZ4K',
+                active=True)
+    mock_get_current_user.return_value = user
+    mock_listen.return_value = 'Listening for events on channel 1'
+    with TestClient(app) as client:
+        response = client.get(
+            "/listen/1",
+            headers={
+                "Accept": "application/json",
+                "Content-Type": "application/x-www-form-urlencoded"
```
JenySadadia, line 704: Remove `Content-Type` from headers dict.

api/test_api.py
```diff
@@ -606,3 +606,103 @@ def test_get_all_nodes_empty_response(mock_get_current_user,
...
+                "Authorization": "Bearer "
+                "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9."
+                "eyJzdWIiOiJib2IifQ.ci1smeJeuX779PptTkuaG1S"
+                "Edkp5M1S1AgYvX8VdB20"
+            },
+        )
+        assert response.status_code == 404
+
+
+def test_listen_endpoint_incorrect_password(mock_get_current_user,
```
JenySadadia, line 683: We should also rename the function with some relevant name e.g. `test_listen_endpoint_without_token`.

JenySadadia (changes requested):
Few things to tweak otherwise LGTM :)

mgalka (approved):

api/test_api.py
alexandrasp: Ok, I did it, but if I remove data and put a invalid token, the api will return 200.

JenySadadia: We don't need token for this test case either.
LGTM :)

JenySadadia (approved):

Issue #66, opened 2021-12-24
Implement HTTP PUT for updating Node objects


gctucker:
The current implementation relies only on HTTP POST for both creating and updating existing Node objects:

```python
@app.post('/node', response_model=Node)
async def post_node(node: Node, token: str = Depends(get_user)):
```

However, this doesn't follow [RFC 2616](https://www.ietf.org/rfc/rfc2616.txt) which is also explained by the comparison between `POST` and `PUT` methods on [restfulapi.net](https://restfulapi.net/rest-put-vs-post/).

Instead, `POST` should only be about creating a new entry in the database and `PUT` should be used to replace an existing one.  Then `PATCH` may be used to update only some of the fields, but that's something we can keep for later.

This task is about removing the logic in the `post_node()` handler which determines whether to add or update a Node document as instead this should be explicitly done by choosing the appropriate method by the client:

* Add `put_node()` handler to replace an existing Node object, with the `id` in the URL like for getting a single node (e.g. `PUT http://host:8001/node/<id>` with the data as JSON).  If there is no object with the provided `id` then return an error.  It should not matter if there is an `id` in the JSON data since it's already provided in the URI, which is the one the implementation should use to update the document in the database.  This should generate an `updated` event for the pub/sub as in the current `post_node()` handler.
* Remove the logic in `post_node()` that would update an existing Node.  Instead, return an HTTP 403 error if there was an `id` in the JSON data being posted as the convention is for the server to create the `id` rather than the client.  Then check if there is any node matching all the fields in the provided data, and return an HTTP 403 error if there is one.  This is to avoid creating duplicates.  Otherwise, add the Node to the database and keep the logic to return HTTP 400 if a `ValueError` is caught.  If it worked, generate a `created` event for pub/sub and return the object to the client as in the current implementation.
* Update the implementation in [`kernelci.data.kernelci_api.submit()`](https://github.com/kernelci/kernelci-core/blob/main/kernelci/data/kernelci_api.py#L72) to use `PUT` if there is an `id` in each item or `POST` if there isn't one.  This effectively moves the logic that was in the `post_node()` handler on the client side.

In addition to making the API more correct according to the RFC 2616 standard, it also means we'll be able to use a single `POST` for atomically adding documents that don't already exist.  This will be useful for new kernel revisions in particular.

PR #86, opened 2022-01-03
Implement HTTP PUT to update Node object

JenySadadia:
Fixes: https://github.com/kernelci/kernelci-api/issues/66

api/main.py: Add put_node handler
api/main.py: Remove the logic to update Node from post_handler()

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -10,6 +10,7 @@ from .db import Database
 from .models import Node, User
 from .pubsub import PubSub, Subscription
 from typing import List
+from bson import ObjectId
 
 app = FastAPI()
 db = Database()
@@ -98,6 +99,21 @@ async def post_node(node: Node, token: str = Depends(get_user)):
         else:
             obj = await db.update(node)
             op = 'updated'
+    except ValueError as e:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=str(e)
+        )
+    await pubsub.publish_cloudevent('node', {'op': op, 'id': str(obj.id)})
+    return obj
+
+
+@app.put('/node/{node_id}', response_model=Node)
+async def put_node(node_id: str, node: Node, token: str = Depends(get_user)):
+    try:
+        node.id = ObjectId(node_id)
+        obj = await db.update(node)
+        op = 'updated'
     except ValueError as e:
         raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST,
--- a/api/main.py
+++ b/api/main.py
@@ -93,12 +93,8 @@ async def get_nodes(request: Request):
 @app.post('/node', response_model=Node)
 async def post_node(node: Node, token: str = Depends(get_user)):
     try:
-        if node.id is None:
-            obj = await db.create(node)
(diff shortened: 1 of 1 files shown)


JenySadadia:
Pushed code for the first task for the issue: https://github.com/kernelci/kernelci-api/issues/66

api/main.py
```diff
@@ -97,8 +97,10 @@ async def post_node(node: Node, token: str = Depends(get_user)):
             obj = await db.create(node)
             op = 'created'
         else:
-            obj = await db.update(node)
-            op = 'updated'
+            # If node id is passed in request JSON
```
gctucker, line 116: The `Database.create()` method already raises an error if the id is not None:
```python
        if obj.id is not None:
            raise ValueError(f"Object cannot be created with id: {obj.id}")
```
So I think we can just remove the `if`/`else` logic from here, and let the `Database` class raise a `ValueError` when there is a problem.  That will then result in `HTTP 403` when the exception is caught below.

api/main.py
```diff
@@ -94,8 +94,24 @@ async def get_nodes(request: Request):
 async def post_node(node: Node, token: str = Depends(get_user)):
     try:
         if node.id is None:
-            obj = await db.create(node)
-            op = 'created'
+            # Check if node exists or not
+            find_node = await db.find_one(
```
gctucker, line 98: The issue here is that there is still a potential race condition between doing the `db.find_one()` call and the `db.create()` one.  So maybe we need to rely on some Mongo DB feature to specify that documents need to be unique with some fields, which would be described in the model.  I see it's possible to do that using an index:
https://docs.mongodb.com/manual/tutorial/unique-constraints-on-arbitrary-fields/

I haven't looked at how to do that with Pydantic models but hopefully there's a simple solution for it.  Maybe this can be removed from this PR and added as a follow-up one: first we'll have a solution to differentiate POST and PUT requests and assume there aren't any race condition issues.  Then we can look into having a way to guarantee some Mongo DB documents don't have duplicates.

JenySadadia: I think we should create a separate issue for this.

gctucker: Yes, so for this PR we can just drop the code that looks for an existing node with the same attributes.

JenySadadia: Then I need to drop the two commits from the PR https://github.com/kernelci/kernelci-api/pull/86/commits/55f71a99da736fe3eb322be54e18c326200699d5 and https://github.com/kernelci/kernelci-api/pull/86/commits/c14e6db6cb49207bb87183ad6cb93d9c25e336fa.

api/main.py
```diff
@@ -96,8 +97,25 @@ async def post_node(node: Node, token: str = Depends(get_user)):
             obj = await db.create(node)
```
gctucker, line 97: I still think we can just remove the whole `if` / `else` logic here and just call `db.create()`.  If there is a `node.id`, an exception will be raised by `db.create()` and then the exception handler below will return HTTP 400 which seems like what would be expected with invalid data.

JenySadadia: Done the changes.

gctucker (approved):

PR #87, opened 2022-01-05
Fix get node by attributes endpoint for parent attribute

JenySadadia:
api/main.py: Add status and parent parameters to get_nodes
api/models.py: Add Node.modify_query_params

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -107,3 +107,9 @@ class Node(DatabaseModel):
 
     def update(self):
         self.updated = datetime.utcnow()
+
+    @classmethod
+    def modify_parent(cls, query_param_dict):
+        """Modify query parameter dictionary for parent"""
+        query_param_dict['parent'] = ObjectId(query_param_dict['parent'])
+        return query_param_dict
--- a/api/main.py
+++ b/api/main.py
@@ -87,7 +87,10 @@ async def get_node(node_id: str):
 
 @app.get('/nodes', response_model=List[Node])
 async def get_nodes(request: Request):
-    return await db.find_by_attributes(Node, dict(request.query_params))
+    query_param_dict = dict(request.query_params)
+    if 'parent' in query_param_dict:
+        query_param_dict = Node.modify_parent(query_param_dict)
+    return await db.find_by_attributes(Node, query_param_dict)
 
 
 @app.get('/get_root_node/{node_id}', response_model=Node)


gctucker (commented):
I don't think we should be implementing this in `api.main`, instead we could have something in `api.models` to convert the strings into types that can be understood in the queries sent to Mongo DB.

api/models.py
```diff
@@ -74,4 +74,4 @@ class Node(ModelId):
     name: str
     revision: Revision
     parent: Optional[PyObjectId]
-    status: Optional[bool] = None
+    status: Optional[bool] = False
```
gctucker, line 77: We need to keep the default value as `None` to indicate there's no result yet.  So I think we'll need to fix the issue differently, maybe using an enum with string values.

gctucker (commented):
There are still a few issues with this PR, I'll come back to it later in more detail.

JenySadadia:
Hi @gctucker ,
We can have 2 approaches to resolve the issue related to `parent`.

Either we can have a method (I have defined `modify_parent`) to convert parent from `string` to `ObjectId` as `request.query_params` is converting everything to string.

Or else we can specify all the query parameters explicitly in `get_nodes` endpoint just like I have defined `parent`. That way we will not be needing `request.query_params` anymore and type conversion will not be needed. The downside of this approach is we need to define each and every member of Node here.

api/models.py
JenySadadia: Resolved issue with `status`.

gctucker:
OK, exposing all the node attributes individually in the endpoint would be a bad design I think as the filters should accept any combination of attributes.  I'll take a look at how this can be done in the model instead.

JenySadadia:
Rebased the PR.
Removed `parent` from request params for `get_nodes` endpoint.
@gctucker As query params will be in `str` type only, I couldn't find any way to internally type cast the request query params but to use `modify_parent`.

api/models.py
```diff
@@ -30,6 +31,14 @@ def validate(cls, value):
         return ObjectId(value)
 
 
+class StatusValues(enum.Enum):
+    """Enumeration to declare values to be used for Node.status"""
+
+    PENDING = "pending"
+    TRUE = "true"
+    FALSE = "false"
```
gctucker, line 39: We should have `PASS` and `FAIL` rather than `TRUE` and `FALSE` which were due to the boolean type. They really mean whether the test passed or failed.

gctucker:
Adding `staging-skip` label as this is changing the schema with the `status` field but the pipeline clients aren't being updated yet to reflect that change.

gctucker:
```
today at 11:05:03 AMpydantic.error_wrappers.ValidationError: 1 validation error for Node
today at 11:05:03 AMstatus
today at 11:05:03 AM  value is not a valid enumeration member; permitted: 'pending', 'true', 'false' (type=type_error.enum; enum_values=[<StatusValues.PENDING: 'pending'>, <StatusValues.TRUE: 'true'>, <StatusValues.FALSE: 'false'>])
```

gctucker:
Thanks, the code and git commits look good.  There's an error in the logs though:
http://staging.kernelci.org:9088/container/774494d23047
```
2022-02-24T09:26:13.827701928Z Listening for events... 
2022-02-24T09:26:13.827772527Z Press Ctrl-C to stop.
2022-02-24T09:27:22.536443683Z Printing node for {'_id': '62174f35d2575d454931a581', 'kind': 'node', 'name': 'checkout', 'revision': {'tree': 'kernelci', 'url': 'https://github.com/kernelci/linux.git', 'branch': 'staging-mainline', 'commit': 'fc5b1486b7664213f96da97ad8515b36d9875bb9', 'describe': 'staging-mainline-20220223.0'}, 'parent': None, 'status': True, 'artifacts': {'tarball': 'http://staging.kernelci.org:9080/linux-kernelci-staging-mainline-staging-mainline-20220223.0.tar.gz'}, 'created': '2022-02-24T09:26:13.483000', 'updated': '2022-02-24T09:27:22.450000'}
2022-02-24T09:27:22.536479083Z Submitting node to KCIDB
2022-02-24T09:27:22.583493557Z Traceback (most recent call last):
2022-02-24T09:27:22.583520957Z   File "/home/kernelci/pipeline/send_kcidb.py", line 85, in <module>
2022-02-24T09:27:22.584004455Z     status = opts.command(configs, opts)
2022-02-24T09:27:22.584023255Z   File "/home/kernelci/pipeline/send_kcidb.py", line 46, in __call__
2022-02-24T09:27:22.584294954Z     self.send_revision(node)
2022-02-24T09:27:22.584335554Z   File "/home/kernelci/pipeline/send_kcidb.py", line 79, in send_revision
2022-02-24T09:27:22.584409554Z     self.client.submit(self.get_revision(node))
2022-02-24T09:27:22.584424554Z   File "/home/kernelci/.local/lib/python3.10/site-packages/kcidb/__init__.py", line 85, in submit
2022-02-24T09:27:22.594922226Z     raise NotImplementedError
2022-02-24T09:27:22.594943526Z NotImplementedError
```
I haven't set up the credentials on staging yet so maybe it's related, but it looks like there's also a problem in the code. Have you seen this issue with your local setup?

gctucker:
Sorry the review above was meant for the KCIDB changes in https://github.com/kernelci/kernelci-pipeline/pull/22

JenySadadia:
> Sorry the review above was meant for the KCIDB changes in [kernelci/kernelci-pipeline#22](https://github.com/kernelci/kernelci-pipeline/pull/22)

Okay.

api/models.py
JenySadadia: Updated the PR for this. Need to update README.md after this gets approved.

gctucker:
> Adding `staging-skip` label as this is changing the schema with the `status` field but the pipeline clients aren't being updated yet to reflect that change.

@JenySadadia I don't think this has been addressed yet, please confirm. Once that's done then we can remove the skip labels and have it tested on staging.

JenySadadia:
> > Adding `staging-skip` label as this is changing the schema with the `status` field but the pipeline clients aren't being updated yet to reflect that change.
> 
> @JenySadadia I don't think this has been addressed yet, please confirm. Once that's done then we can remove the skip labels and have it tested on staging.

Created PR https://github.com/kernelci/kernelci-pipeline/pull/27 to reflect changes to client scripts.

gctucker:
Looks like there's still a piece missing:
```
today at 8:05:43 AMListening for events... 
today at 8:05:43 AMPress Ctrl-C to stop.
today at 8:05:44 AMTime                        Commit        Status    Name
today at 8:05:44 AMTraceback (most recent call last):
today at 8:05:44 AM  File "/home/kernelci/./pipeline/notifier.py", line 67, in <module>
today at 8:05:44 AM    status = opts.command(configs, opts)
today at 8:05:44 AM  File "/home/kernelci/./pipeline/notifier.py", line 52, in __call__
today at 8:05:44 AM    status=status_map[obj['status']],
today at 8:05:44 AMKeyError: None
today at 8:53:54 AMContainer stopped
```
https://staging.kernelci.org:9088/container/c990c46c02c1

Probably `trigger.py` is not setting the status, but then it should have been set to "pending" by default in the API. Something to investigate.

JenySadadia:
@gctucker Added code changes specific to Node.status into different PR: https://github.com/kernelci/kernelci-api/pull/123

JenySadadia:
Could you please review this again @gctucker ?

gctucker (approved):
Tested OK on staging. I think this can be reworked a bit to be more generic, but I'll make a follow-up PR and you can review it so we can resolve this one now.

PR #89, opened 2022-01-10
Initial PubSub tests

mgalka:
test/test_pubsub: add initial PubSub.subscribe tests
    
Add tests subscribing to single and multiple channels using api.pubsub.PubSub.subscribe method.


test/test_pubsub.py
```diff
@@ -0,0 +1,58 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2021 Michal Galka
```
gctucker, line 3:
```suggestion
# Copyright (C) 2021 Collabora Limited
```

gctucker (commented):
LGTM

Note: As part of the tests to be added, it would be good to check whether the pub/sub id keeps increasing after doing a subscribe/unsubscribe.  With a real Redis, the id number keeps incrementing until the persistent data is reset to ensure older ids can't collide with new subscriptions.

gctucker:
One thing though, the tests aren't run by GitHub actions because it's only looking for tests within the `api` directory.

test/test_pubsub.py
mgalka: Nice catch. I also corrected the year as we're having 2022.

mgalka:
> One thing though, the tests aren't run by GitHub actions because it's only looking for tests within the `api` directory.

I modified GitHub actions config to scan the test directory, too.

gctucker (approved):
LGTM

PR #95, opened 2022-01-12
Fix pylint errors in api.db and run it in GitHub actions

gctucker:
Fix all pylint errors in `api.db` module and run it in GitHub actions.

### DIFF
--- a/api/db.py
+++ b/api/db.py
@@ -3,13 +3,21 @@
 # Copyright (C) 2021 Collabora Limited
 # Author: Guillaume Tucker <guillaume.tucker@collabora.com>
 
+"""Database abstraction"""
+
 from bson import ObjectId
-from pymongo import MongoClient
-from .models import Node, User
 from motor import motor_asyncio
+from .models import Node, User
 
 
-class Database(object):
+class Database:
+    """Database abstraction class
+
+    This class provides an abstraction layer to access the Mongo DB database
+    asynchronously using the models defined in `.models`.  *host* is the
+    hostname where the database is, and *db_name* is the name of the database.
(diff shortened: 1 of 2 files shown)


api/db.py
```diff
@@ -32,21 +41,37 @@ async def find_all(self, model):
...
+        """
         col = self._get_collection(model)
         data = await col.find(attributes).to_list(None)
         return list(model(**obj) for obj in data)
 
     async def create(self, obj):
+        """Create a database document from a model object
+
+        For a given object instance from .models, create a document in the
+        database.  Add the new id to the object and return it.
```
JenySadadia, line 73: I would suggest,
```
"""Create a database document from a model object
        For a given object instance from .models, create a document in the
        database.  Add the new id to the object and return it. 
        Raise 'ValueError' if the object is provided along with the id.
"""
```

api/db.py
```diff
@@ -56,6 +81,11 @@ async def create(self, obj):
         return obj
 
     async def update(self, obj):
+        """Update an existing document from a model object
+
+        For a given object instance from .models, update the document in the
+        database with a matching id.  The object is then returned.
```
JenySadadia, line 87:
```
"""Update an existing document from a model object
        For a given object instance from .models, update the document in the
        database with a matching id.  The object is then returned.
        Raise 'ValueError' if the object is provided without the id or
        provided object id does not match with any of the objects from the database.
"""
```

JenySadadia (commented):
I think docstring should include possible exceptions a function might throw.

JenySadadia:
LGTM :)

PR #96, opened 2022-01-13
test: split test cases defined in api/test_api.py

JenySadadia:
Move related test cases from api/test_api.py to
various files under test directory.


JenySadadia:
We can remove api/test_api.py file once we have Alex's PR merged.

mgalka (approved):

PR #97, opened 2022-01-13
api/pubsub.py: use 'await' with aioredis unsubscribe() call

JenySadadia:
Fix warning:
api/pubsub.py:52: RuntimeWarning: coroutine 'PubSub.execute_command'
was never awaited :sub.unsubscribe()


gctucker (commented):
Small things that could be improved: The commit subject is a bit too long and there could be a `Fixes:` trailer with the commit that introduced the problem.

PR #98, opened 2022-01-19
api/models.py: modify Revision.describe to be Optional

JenySadadia:
trigger script from the kernelci-pipeline will post
a new node with an empty Revision.describe property.
When a new revision is available, the client will
update the revision information with the git describe.

Fixes: 15380ac ("api.models: add Revision.url and .describe")

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -65,7 +65,7 @@ class Revision(BaseModel):
     url: str
     branch: str
     commit: str
-    describe: str
+    describe: Optional[str] = None
 
 
 class Node(ModelId):


gctucker:
Tiny detail, but it's causing the trailer to not be parsed automatically:
```
Fixes:15380ac ("api.models: add Revision.url and .describe")
```
is missing a space after the `:`, it should be:
```
Fixes: 15380ac ("api.models: add Revision.url and .describe")
```

gctucker (approved):

PR #99, opened 2022-01-19
Pylint api.pubsub

gctucker:
Fix Pylint warnings in api.pubsub and run it in GitHub Actions

### DIFF
--- a/api/pubsub.py
+++ b/api/pubsub.py
@@ -3,46 +3,66 @@
 # Copyright (C) 2021 Collabora Limited
 # Author: Guillaume Tucker <guillaume.tucker@collabora.com>
 
-import aioredis
+"""Pub/Sub implementation"""
+
 import asyncio
+
+import aioredis
 from cloudevents.http import CloudEvent, to_json
 from pydantic import BaseModel, BaseSettings
 
 
 class Settings(BaseSettings):
+    """Pub/Sub settings loaded from the environment"""
     cloud_events_source: str = "https://api.kernelci.org/"
     redis_host: str = "redis"
     redis_db_number: int = 1
 
 
 class Subscription(BaseModel):
+    """Pub/Sub subscription object model"""
     id: int
     channel: str
 
 
 class PubSub:
+    """Pub/Sub implementation class
+
+    This class provides a Pub/Sub implementation based on Redis.  The `host`
+    and `db_number` parameters can be specified to connect to a particular
+    Redis database.  By default, the `redis` host will be used as it is
+    available from the `docker-compose` setup.
+    """
+
     ID_KEY = 'kernelci-api-pubsub-id'
 
     @classmethod
-    async def create(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


api/pubsub.py
```diff
@@ -3,40 +3,54 @@
...
 
     async def _init_sub_id(self):
         await self._redis.setnx(self.ID_KEY, 0)
 
     async def subscribe(self, channel):
+        """Subscribe to a Pub/Sub channel
+
+        Subscribe to a given channel and return a Subscription object
+        containing the subscription id which can then be used again in other
+        methods.
```
JenySadadia, line 52: I would suggest,
```     
        Subscribe to a given channel asynchronously and return a Subscription object
        containing the subscription id and channel which can then be used again in other
        methods.
```

api/pubsub.py
```diff
@@ -45,6 +59,11 @@ async def subscribe(self, channel):
             return Subscription(id=sub_id, channel=channel)
 
     async def unsubscribe(self, sub_id):
+        """Unsubscribe from a Pub/Sub channel
+
+        Unsubscribe from a channel using the provided subscription id as found
+        in a Subscription object.
```
JenySadadia, line 65: I would suggest,
```
        Unsubscribe from a channel asynchronously using the provided subscription id as found
        in a Subscription object and return True. Return False if subscription id can not be found 
        in `_subscriptions` dictionary. 
```

JenySadadia (commented):
I have submitted a few suggestions for the docstring of subscribe and unsubscribe methods.
LGTM otherwise.

api/pubsub.py
gctucker: Only the id is needed for other methods in this class, which is why I didn't mention the channel explicitly.  It's part of the Subscription model, which is mentioned as the type of the returned object.

api/pubsub.py
gctucker: The `_subscriptions` is a private attribute of the class, so it's an implementation detail and not something that should be shared with users.  But yes, returning True / False should be documented (or maybe the method should raise an exception).  So I'll fix that.

JenySadadia: @gctucker I think this is not resolved yet.

JenySadadia (commented):
@gctucker I think the docstring for `unsubscribe` is not resolved yet.

gctucker:
It should be resolved in #105.  It became a larger change with exceptions so I moved it to a separate PR as this wasn't related to fixing Pylint errors any more.

gctucker:
@JenySadadia Could you please approve this PR unless you have any other comments?  That way we could have it merged and rebase #105 as at the moment it's based on top of this one.

JenySadadia:
> @JenySadadia Could you please approve this PR unless you have any other comments? That way we could have it merged and rebase #105 as at the moment it's based on top of this one.

Sure.

JenySadadia (approved):

Issue #40, opened 2021-11-19
Add Redis options to the Settings


gctucker:
Some parameters such as the Redis host and the database number are currently passed to the PubSub constructor in `api.pubsub`.  There isn't a way to change these values from the `api.main` where the object is created, which can cause some limitations.  To be able to use another Redis instance, attributes can be added to the Settings class with default values.  Then the constructor can take `None` by default and rely on the settings if nothing is provided by the caller.

PR #100, opened 2022-01-21
api/pubsub.py: add redis option to Settings class

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/40

Define 'redis_host' and 'redis_database_number' in Settings class.
Use redis options from Settings class in the PubSub constructor.

### DIFF
--- a/api/pubsub.py
+++ b/api/pubsub.py
@@ -11,6 +11,8 @@ from pydantic import BaseModel, BaseSettings
 
 class Settings(BaseSettings):
     cloud_events_source: str = "https://api.kernelci.org/"
+    redis_host: str = "redis"
+    redis_db_number: int = 1
 
 
 class Subscription(BaseModel):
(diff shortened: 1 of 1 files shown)


api/pubsub.py
```diff
@@ -27,7 +29,8 @@ async def create(self, *args, **kwargs):
         await pubsub._init_sub_id()
         return pubsub
 
-    def __init__(self, host='redis', db=1):
+    def __init__(self, host=Settings().redis_host,
+                 db=Settings().redis_db_number):
```
gctucker, line 33: There is already a `Settings` object created in `self._settings`. So I think the default values in `__init__` should be `None` and then values from `self._settings` should be used.  For example:
```python
if host is None:
    host = self._settings.host
```

gctucker (approved):

Issue #38, opened 2021-11-19
Add created and updated times to the Node model


gctucker:
A key piece of information is the date and time when data is being stored in the database.  As such, the Node model should have a `created` field with the date / time when the Node was first created in the database and an `updated` field whenever its data is being changed.  These should not be sent by the client but automatically added by the backend, like for the `_id` field.

Note: The `updated` value should be the same as `created` initially, until it actually gets updated later on.

- [ ] `created` and `updated` date / time optional fields added to the Node model
- [ ] `api.db` logic added to automatically add `created` for new Node objects and `updated` when they get changed
- [ ] verified with PoC pipeline that the date / time can be retrieved and are correct

PR #101, opened 2022-01-24
Add created and updated times to the Node model

JenySadadia:

api/models.py: add 'created' and 'updated' fields to Node
test/test_node_handler.py: add timestamp fields in response json
Store created and updated timestamp for Node

### DIFF
--- a/api/db.py
+++ b/api/db.py
@@ -93,6 +93,9 @@ class Database:
         if obj.id is None:
             raise ValueError("Cannot update object with no id")
         col = self._get_collection(obj.__class__)
+        db_obj = await col.find_one({'_id': ObjectId(obj.id)})
+        if db_obj:
+            Node.set_created(obj, db_obj.get('created'))
         res = await col.replace_one(
             {'_id': ObjectId(obj.id)}, obj.dict(by_alias=True)
         )
--- a/api/db.py
+++ b/api/db.py
@@ -82,7 +82,7 @@ class Database:
         obj.id = res.inserted_id
         return obj
 
-    async def update(self, obj):
+    async def update(self, model, obj):
         """Update an existing document from a model object
 
         For a given object instance from .models, update the document in the
@@ -94,7 +94,7 @@ class Database:
             raise ValueError("Cannot update object with no id")
         col = self._get_collection(obj.__class__)
         db_obj = await col.find_one({'_id': ObjectId(obj.id)})
-        if db_obj:
+        if db_obj and model == Node:
             Node.set_created(obj, db_obj.get('created'))
         res = await col.replace_one(
             {'_id': ObjectId(obj.id)}, obj.dict(by_alias=True)
--- a/api/models.py
+++ b/api/models.py
@@ -6,7 +6,7 @@
 """KernelCI API model definitions"""
 
 from typing import Optional
-
+from datetime import datetime
 from bson import ObjectId
 from pydantic import BaseModel, Field
 
@@ -75,3 +75,5 @@ class Node(ModelId):
     revision: Revision
     parent: Optional[PyObjectId]
     status: Optional[bool] = None
+    created: Optional[datetime] = Field(default_factory=datetime.utcnow)
+    updated: Optional[datetime] = Field(default_factory=datetime.utcnow)
--- a/api/models.py
+++ b/api/models.py
@@ -77,3 +77,8 @@ class Node(ModelId):
     status: Optional[bool] = None
     created: Optional[datetime] = Field(default_factory=datetime.utcnow)
     updated: Optional[datetime] = Field(default_factory=datetime.utcnow)
+
+    @classmethod
+    def set_created(cls, obj, value):
+        """Set the value of 'created' Node attribute"""
+        setattr(obj, 'created', value)
--- a/api/main.py
+++ b/api/main.py
@@ -108,7 +108,7 @@ async def post_node(node: Node, token: str = Depends(get_user)):
 async def put_node(node_id: str, node: Node, token: str = Depends(get_user)):
     try:
         node.id = ObjectId(node_id)
-        obj = await db.update(node)
+        obj = await db.update(Node, node)
         op = 'updated'
     except ValueError as e:
         raise HTTPException(
--- a/test/test_node_handler.py
+++ b/test/test_node_handler.py
@@ -116,8 +116,8 @@ def test_create_node_endpoint(mock_get_current_user, mock_init_sub_id,
             )
         print("response.json()", response.json())
         assert response.status_code == 200
-        assert ('_id', 'kind', 'name', 'revision', 'parent',
-                'status') == tuple(response.json().keys())
+        assert ('_id', 'kind', 'name', 'revision', 'parent', 'status',
+                'created', 'updated') == tuple(response.json().keys())
 
 
 def test_get_node_by_attributes_endpoint(mock_get_current_user,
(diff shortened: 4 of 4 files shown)


JenySadadia:
Fixed node test cases in test/test_node_handler.py
I have not committed changes to api/test_api.py as the file is going to be removed when https://github.com/kernelci/kernelci-api/pull/102 gets merged.

gctucker:
Please fix the Pylint issue:
```
api/db.py:11:0: C0411: standard import "from datetime import datetime" should be placed before "from bson import ObjectId" (wrong-import-order)
```

JenySadadia:
> Please fix the Pylint issue:
> 
> ```
> api/db.py:11:0: C0411: standard import "from datetime import datetime" should be placed before "from bson import ObjectId" (wrong-import-order)
> ```

Done.

gctucker:
This is the part of the Pydantic documentation I was looking for, it shows how to specify how to construct a default value dynamically with `default_factory`:
https://pydantic-docs.helpmanual.io/usage/models/#field-with-dynamic-default-value

api/models.py
```diff
@@ -75,3 +75,5 @@ class Node(ModelId):
     revision: Revision
     parent: Optional[PyObjectId]
     status: Optional[bool] = None
+    created: Optional[datetime] = None
```
gctucker, line 78: To set the created time automatically when the object is created:
```suggestion
    created: Optional[datetime] = Field(default_factory=datetime.utcnow)
```

api/db.py
```diff
@@ -93,6 +98,9 @@ async def update(self, obj):
         if obj.id is None:
             raise ValueError("Cannot update object with no id")
         col = self._get_collection(obj.__class__)
+        if obj.__class__ == Node:
+            obj.created = obj.id.generation_time
```
gctucker, line 102: The `created` time should be set when the object is created using the default value constructor.  So I don't think we need to update it here, only the `updated` time needs to. It would also be better to have an `update()` method in the base class of the models so we don't need to check the type of the object in `Database`.

api/db.py
```diff
@@ -93,6 +93,9 @@ async def update(self, obj):
         if obj.id is None:
             raise ValueError("Cannot update object with no id")
         col = self._get_collection(obj.__class__)
+        node = await col.find_one({'_id': ObjectId(obj.id)})
```
gctucker, line 96: The variable should not be `node` as we don't know what type of object is being updated here. We could find out using `obj.__class__` but it's better to not rely on that in the `Database` class.

api/db.py
```diff
@@ -93,6 +93,9 @@ async def update(self, obj):
...
+        if node:
+            Node.set_created(obj, node.get('created'))
```
gctucker, line 98: OK I see, the `created` field is set in `api.main` when making an object if the client didn't send it. Likewise, the `updated` one is also automatically set as a side-effect. But this should be managed by the model rather than `api.main` or `api.db`. So I think we have 2 different issues for `created` and `updated`, we probably should be addressing them separately. Maybe this PR could focus on the `created` field to get it fixed, and then a follow-up would address the `updated` one.

We may need to have some properties set for each field to describe how they should be handled. For example, `created` should not be saved again in the database. So either it has to match exactly what is already there, or an incoming value has to be discarded. I'm not sure which is the best way to do this, maybe the latter with immutable values enforced in Mongo DB would be ideal as the document saved would then have to be what the client sent with the real `created` date.

With the current pipeline, when a node gets updated, normally the data is coming from an existing node with all the fields including `created`. Then the `status` and maybe other fields get updated, but the `created` time should remain the same.  Is that not the case?

api/models.py
```diff
@@ -75,3 +75,10 @@ class Node(ModelId):
     revision: Revision
     parent: Optional[PyObjectId]
     status: Optional[bool] = None
+    created: Optional[datetime] = Field(default_factory=datetime.utcnow)
+    updated: Optional[datetime] = Field(default_factory=datetime.utcnow)
+
+    @classmethod
+    def set_created(cls, obj, value):
```
gctucker, line 82: This should be the same as doing `obj.created = sometime`, right?

JenySadadia: Yes, it is the same. But we can not write Node specific logic to Database class. So, I created Node.set_created to do the same.

api/db.py
JenySadadia: Yes, I need to rename the `node` obj. Another way to check the object class is `type` operator.

api/db.py
JenySadadia: Yes, it is the case.
As you have suggested using immutable type, I have already tried it with `created` field. I used the below statement to make it immutable.
```
created: Optional[datetime] = Field(default_factory=datetime.utcnow, allow_mutation=False)
```
But it tuned out that It doesn't affect when used with `default_factory`, as `default_factory` is updating the value whenever the object is being saved to the DB.

If I don't put the logic to set `created` field from DB object, `default_factory` will update it automatically. Hence we needed the logic to retrieve the field from the DB and set the field to that value.

api/db.py
gctucker: We can rely on `obj.__class__` to find the matching database collection in the dictionary using `_get_collection()`, but then we shouldn't be making the logic of the code depend on the class of the model.  Instead, all the models should have the same abstract methods to handle the data.

api/db.py
gctucker: I believe `allow_mutation` just means that the attribute in the object can't be changed. That's not the same as making it immutable in Mongo DB.

I think we should require the client to either not send `created` or send the real `created` time which should be the same as what is in the database. Then if we make the field in the database immutable, it will cause an error if the value from the client is different. If we can't make the field immutable in Mongo DB (although I believe we can with indexes), then we could add a method to the model to check whether an incoming object contains valid data and compare the immutable fields there instead. If they're not the same then it would raise an exception.

gctucker:
I've tried experimenting by hand with this:
```
$ docker-compose run api /bin/bash
kernelci@feacc3d3e72b:~$ python3
>>> import api.models
>>> node = api.models.Node(name='hack', revision=api.models.Revision(tree='oak', url='https://kernelci.org', branch='main', commit='1234abcd'))
>>> node.created
datetime.datetime(2022, 2, 2, 10, 34, 27, 995285)
>>> node = api.models.Node(name='hack', revision=api.models.Revision(tree='oak', url='https://kernelci.org', branch='main', commit='1234abcd'))
>>> node.created
datetime.datetime(2022, 2, 2, 10, 36, 18, 365837)
```
which is working as expected.  The tricky part is about handling this in the API and the database.

gctucker:
As long as there is no code changing the value of `created` then it should always keep the value assigned to it when the object was created.

Here's an other experiment:
```
>>> from datetime import datetime
>>> now = datetime.utcnow()
>>> now
datetime.datetime(2022, 2, 2, 10, 44, 22, 832346)
>>> node = api.models.Node(name='hack', revision=api.models.Revision(tree='oak', url='https://kernelci.org', branch='main', commit='1234abcd'))
>>> node.created
datetime.datetime(2022, 2, 2, 10, 45, 52, 179756)
>>> node = api.models.Node(name='hack', revision=api.models.Revision(tree='oak', url='https://kernelci.org', branch='main', commit='1234abcd'), created=now)
>>> node.created
datetime.datetime(2022, 2, 2, 10, 44, 22, 832346)
```

JenySadadia:
I have created two separate PRs to add 'created' and 'updated' fields.
https://github.com/kernelci/kernelci-api/pull/109
https://github.com/kernelci/kernelci-api/pull/111

Hence, closing this one.

PR #103, opened 2022-01-25
test/test_node_handler.py: add test_create_node_with_id

JenySadadia:
Add test case when node attributes along with node id
is passed to POST node request.


JenySadadia:
As per our discussion, we need database mocking here.
Should I close this PR to put the issue on hold? @mgalka

mgalka:
There seem to be no good mocking library for asynchronous usage of MongoDB. It seems that it's a better idea to make this test case a part of end-to-end tests.

Issue #28, opened 2021-11-11
Add Pub/Sub event filters


gctucker:
In addition to being able to choose which channel to subscribe, we need to add arbitrary filters to only receive some events.  When subscribing to the `node` channel, the client could provide a dictionary with keys and values that have to match the object which caused the event.  For example:

To only get notifications for a particular type of nodes:
```json
{"name": "git-checkout"}
```

To only get notifications for a particular kernel tree:
```json
{"revision":
  {"tree": "mainline"}
}
```

- [ ] `api.pubsub` implementation updated to accommodate for filters associated with each subscription
- [ ] `api.main` updated accordingly to handle the requests
- [ ] Documentation updated accordingly

PR #104, opened 2022-01-26
Pubsub event filter

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/28

api.pubsub: add class members for pubsub event filters
api.pubsub: filter pubsub events in listen method
api/main.py: add filter in subscribe
api/main.py: add data fields to 'create' and 'update' events
test/test_pubsub: add _filters to unsubscribe test cases 
test/test_pubsub: add _filters to subscribe test cases
test/test_subscribe_handler: add 'filter' to subscribe test case
api/client: add filter dict to subscribe post data

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -100,7 +100,11 @@ async def post_node(node: Node, token: str = Depends(get_user)):
             status_code=status.HTTP_400_BAD_REQUEST,
             detail=str(e)
         )
-    await pubsub.publish_cloudevent('node', {'op': op, 'id': str(obj.id)})
+    await pubsub.publish_cloudevent('node', {
+                                    'op': op, 'id': str(obj.id),
+                                    'name': str(obj.name),
+                                    'revision': str(obj.revision)}
+                                    )
     return obj
 
 
@@ -115,7 +119,11 @@ async def put_node(node_id: str, node: Node, token: str = Depends(get_user)):
             status_code=status.HTTP_400_BAD_REQUEST,
             detail=str(e)
         )
-    await pubsub.publish_cloudevent('node', {'op': op, 'id': str(obj.id)})
+    await pubsub.publish_cloudevent('node', {
+                                    'op': op, 'id': str(obj.id),
+                                    'name': str(obj.name),
+                                    'revision': str(obj.revision)}
+                                    )
     return obj
 
 
--- a/api/main.py
+++ b/api/main.py
@@ -9,7 +9,7 @@ from .auth import Authentication, Token
 from .db import Database
 from .models import Node, User
 from .pubsub import PubSub, Subscription
-from typing import List
+from typing import List, Optional
 from bson import ObjectId
 
 app = FastAPI()
@@ -123,8 +123,9 @@ async def put_node(node_id: str, node: Node, token: str = Depends(get_user)):
 # Pub/Sub
 
 @app.post('/subscribe/{channel}', response_model=Subscription)
-async def subscribe(channel: str, user: User = Depends(get_user)):
-    return await pubsub.subscribe(channel)
+async def subscribe(channel: str, filters: Optional[dict] = None,
+                    user: User = Depends(get_user)):
+    return await pubsub.subscribe(channel, filters)
 
 
 @app.post('/unsubscribe/{sub_id}')
--- a/api/pubsub.py
+++ b/api/pubsub.py
@@ -10,6 +10,7 @@ import asyncio
 import aioredis
 from cloudevents.http import CloudEvent, to_json
 from pydantic import BaseModel, BaseSettings
+from typing import Optional
 
 
 class Settings(BaseSettings):
@@ -23,6 +24,7 @@ class Subscription(BaseModel):
     """Pub/Sub subscription object model"""
     id: int
     channel: str
+    filter: Optional[dict] = None
 
 
 class PubSub:
@@ -52,11 +54,12 @@ class PubSub:
         self._redis = aioredis.from_url(f'redis://{host}/{db_number}')
         self._subscriptions = {}
         self._lock = asyncio.Lock()
+        self._filters = {}
 
     async def _init_sub_id(self):
         await self._redis.setnx(self.ID_KEY, 0)
 
-    async def subscribe(self, channel):
+    async def subscribe(self, channel, filter=None):
         """Subscribe to a Pub/Sub channel
 
         Subscribe to a given channel and return a Subscription object
@@ -67,8 +70,9 @@ class PubSub:
         async with self._lock:
             sub = self._redis.pubsub()
             self._subscriptions[sub_id] = sub
+            self._filters[sub_id] = filter
             await sub.subscribe(channel)
-            return Subscription(id=sub_id, channel=channel)
+            return Subscription(id=sub_id, channel=channel, filter=filter)
 
     async def unsubscribe(self, sub_id):
         """Unsubscribe from a Pub/Sub channel
@@ -82,6 +86,7 @@ class PubSub:
             if sub is None:
                 raise ValueError(f"Invalid subscription id: {sub_id}")
             self._subscriptions.pop(sub_id)
+            self._filters.pop(sub_id)
             await sub.unsubscribe()
 
     async def listen(self, sub_id):
--- a/api/pubsub.py
+++ b/api/pubsub.py
@@ -6,11 +6,11 @@
 """Pub/Sub implementation"""
 
 import asyncio
+from typing import Optional
 
 import aioredis
-from cloudevents.http import CloudEvent, to_json
+from cloudevents.http import CloudEvent, to_json, from_json
 from pydantic import BaseModel, BaseSettings
-from typing import Optional
 
 
 class Settings(BaseSettings):
@@ -89,6 +89,34 @@ class PubSub:
             self._filters.pop(sub_id)
             await sub.unsubscribe()
 
+    def pubsub_event_filter(self, sub_id, msg):
+        """Filter Pub/Sub events
+
+        Filter received Pub/Sub event using provided dictionary.
+        Return True if the event matches with the filter, otherwise False.
+        """
+        event_filter_status = True
+        filters = self._filters.get(sub_id)
+        if not filters:
+            return event_filter_status
+        msg_data = from_json(msg['data']).data
+        for key, value in filters.items():
+            if key not in msg_data.keys():
+                continue
+
+            if key == 'revision':
+                msg_revision_dict = get_revision_dict(msg_data)
+                for revision_key, revision_value in value.items():
+                    if revision_key not in msg_revision_dict:
+                        continue
+                    if revision_value != msg_revision_dict[revision_key]:
+                        event_filter_status = False
+                        break
+            elif value != msg_data[key]:
+                event_filter_status = False
+                break
+        return event_filter_status
+
     async def listen(self, sub_id):
         """Listen for Pub/Sub messages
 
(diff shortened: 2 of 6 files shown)


api/main.py
```diff
@@ -100,7 +100,11 @@ async def post_node(node: Node, token: str = Depends(get_user)):
             status_code=status.HTTP_400_BAD_REQUEST,
             detail=str(e)
         )
-    await pubsub.publish_cloudevent('node', {'op': op, 'id': str(obj.id)})
+    await pubsub.publish_cloudevent('node', {
+                                    'op': op, 'id': str(obj.id),
+                                    'name': str(obj.name),
+                                    'revision': str(obj.revision)}
+                                    )
```
gctucker, line 107: Please don't change the data sent in the event, the aim is only to filter which events are being sent. This is to reduce the number of events being sent, and also to simplify or remove the need for filtering on the client side.

api/main.py
```diff
@@ -115,16 +119,21 @@ async def put_node(node_id: str, node: Node, token: str = Depends(get_user)):
...
```
gctucker, line 126: Ditto

api/main.py
gctucker: However, we can pass a dictionary with the attributes and values from the object as an extra argument to the `publish_cloudevent()` method to use as the data for applying filters.

For example, the object JSON dump would contain something like `{"name": "checkout", "revision": {...}}` and that can then be used for filtering with each pub/sub subscription.

api/pubsub.py
```diff
@@ -82,8 +86,37 @@ async def unsubscribe(self, sub_id):
             if sub is None:
                 raise ValueError(f"Invalid subscription id: {sub_id}")
             self._subscriptions.pop(sub_id)
+            self._filters.pop(sub_id)
             await sub.unsubscribe()
 
+    def pubsub_event_filter(self, sub_id, msg):
```
gctucker, line 92: One issue is that the logic should not have knowledge of the schema, or it will need to be updated all the time to be kept in sync and that's not maintainable. Instead, the data used for filtering should be provided alongside the filters and the logic could go through it recursively.

gctucker (commented):
The main issue here is that:
* events should have just the operation type and node ID, as if we start sending full Node objects then they're not events any more but data being pushed to clients
* each subscriber has a different set of filters, but events are sent to global channels
* filtering in the `listen()` method can only be done using the fields in the event data

Some ideas about how to solve this (mutually exclusive ones):
* create a separate channel for each subscriber, and iterate over them when publishing events (with `{'node': [<list of actual Redis channels>]}`)
* pass the data filter in the event so it can be filtered in `listen()` but then probably drop it before actually sending it to the client
* logic could be put in kernelci-core to simplify client code, with all events still actually sent by the API

JenySadadia:
> The main issue here is that:
> 
> * events should have just the operation type and node ID, as if we start sending full Node objects then they're not events any more but data being pushed to clients
> * each subscriber has a different set of filters, but events are sent to global channels
> * filtering in the `listen()` method can only be done using the fields in the event data
> 
> Some ideas about how to solve this (mutually exclusive ones):
> 
> * create a separate channel for each subscriber, and iterate over them when publishing events (with `{'node': [<list of actual Redis channels>]}`)
> * pass the data filter in the event so it can be filtered in `listen()` but then probably drop it before actually sending it to the client
> * logic could be put in kernelci-core to simplify client code, with all events still actually sent by the API

I tried to solve the problem using ` create a separate channel for each subscriber, and iterate over them when publishing events (with `{'node': [<list of actual Redis channels>]}`)` approach.
```
async def publish(self, channel, message):
        """Publish a message on a channel

        Publish an arbitrary message asynchronously on a Pub/Sub channel.
        """
        for sub_id in self._subscriptions:
            event_filter_status = self.pubsub_event_filter(sub_id, message)
            if event_filter_status is False:
                continue
        await self._redis.publish(channel, message)
```
It turns out that `self._redis.publish` will send messages to every subscriber of the channel. We can not select a particular subscriber to send the message based on the filter.
Hence, we can not use this approach.

JenySadadia:
Hi @gctucker , I analyzed the way to solve the issue using `logic could be put in kernelci-core to simplify client code, with all events still actually sent by the API` approach.

For that we need to add pub sub filter logic in `kernelci_api.KernelCI_API.get_event` method of kenelci-core.
This method calls `listen` endpoint from kernelci-api.
```
def get_event(self, sub_id):
        path = '/'.join(['listen', str(sub_id)])
        resp = self._get(path)
        return from_json(resp.json().get('data'))
```

So ultimately we need to add logic in `PubSub.listen` of kernelci-api which I have already implemented in this PR.

To use pubsub filters in client scripts of kernelci-pipeline, we need to add filter argument in `kernelci_api.KernelCI_API.subscribe` method and pass the filter dict to `db.subscribe` in kernelci-pipeline scripts. This way `get_event` will only receive events matching with the provided filter.

gctucker:
> > The main issue here is that:
> > 
> > * events should have just the operation type and node ID, as if we start sending full Node objects then they're not events any more but data being pushed to clients
> > * each subscriber has a different set of filters, but events are sent to global channels
> > * filtering in the `listen()` method can only be done using the fields in the event data
> > 
> > Some ideas about how to solve this (mutually exclusive ones):
> > 
> > * create a separate channel for each subscriber, and iterate over them when publishing events (with `{'node': [<list of actual Redis channels>]}`)
> > * pass the data filter in the event so it can be filtered in `listen()` but then probably drop it before actually sending it to the client
> > * logic could be put in kernelci-core to simplify client code, with all events still actually sent by the API
> 
> I tried to solve the problem using `create a separate channel for each subscriber, and iterate over them when publishing events (with`{'node': []}`)` approach.
> 
> ```
> async def publish(self, channel, message):
>         """Publish a message on a channel
> 
>         Publish an arbitrary message asynchronously on a Pub/Sub channel.
>         """
>         for sub_id in self._subscriptions:
>             event_filter_status = self.pubsub_event_filter(sub_id, message)
>             if event_filter_status is False:
>                 continue
>         await self._redis.publish(channel, message)
> ```
> 
> It turns out that `self._redis.publish` will send messages to every subscriber of the channel. We can not select a particular subscriber to send the message based on the filter. According to the Redis documentation, publish does not work based on subscription id. Hence, we can not use this approach.

You need to use a different channel name for each subscriber. Then it will work.

gctucker:
> Hi @gctucker , I analyzed the way to solve the issue using `logic could be put in kernelci-core to simplify client code, with all events still actually sent by the API` approach.
> 
> For that we need to add pub sub filter logic in `kernelci_api.KernelCI_API.get_event` method of kenelci-core. This method calls `listen` endpoint from kernelci-api.
> 
> ```
> def get_event(self, sub_id):
>         path = '/'.join(['listen', str(sub_id)])
>         resp = self._get(path)
>         return from_json(resp.json().get('data'))
> ```
> 
> So ultimately we need to add logic in `PubSub.listen` of kernelci-api which I have already implemented in this PR.
> 
> To use pubsub filters in client scripts of kernelci-pipeline, we need to add filter argument in `kernelci_api.KernelCI_API.subscribe` method and pass the filter dict to `db.subscribe` in kernelci-pipeline scripts. This way `get_event` will only receive events matching with the provided filter.

We can have a method in kernelci-core to directly get nodes matching a particular criteria. It would need to be specific to nodes, but that's OK from a client library point of view. No changes would need to be made in the API, just the code we currently have in `runner.py` etc. would be generalised in kernelci-core. It would work like this:
```python
sub_id = db.subscribe_for_nodes(filters={'name': 'checkout'})
while True:
  node = db.receive_node(sub_id)
```
The filters would be kept in the `db` object, which would receive all events from the `nodes` channel then retrieve the Node object and apply the filters. If they match, then it returns the Node. Otherwise it keeps listening.

I think we can have this implemented in any case as it's convenient and most clients will be doing that. Then if we implement server-side event filters then we can change the implementation in kernelci-core to rely on it and stop keeping the filters in the `db` object.

JenySadadia:
> > > The main issue here is that:
> > > 
> > > * events should have just the operation type and node ID, as if we start sending full Node objects then they're not events any more but data being pushed to clients
> > > * each subscriber has a different set of filters, but events are sent to global channels
> > > * filtering in the `listen()` method can only be done using the fields in the event data
> > > 
> > > Some ideas about how to solve this (mutually exclusive ones):
> > > 
> > > * create a separate channel for each subscriber, and iterate over them when publishing events (with `{'node': [<list of actual Redis channels>]}`)
> > > * pass the data filter in the event so it can be filtered in `listen()` but then probably drop it before actually sending it to the client
> > > * logic could be put in kernelci-core to simplify client code, with all events still actually sent by the API
> > 
> > 
> > I tried to solve the problem using `create a separate channel for each subscriber, and iterate over them when publishing events (with`{'node': []}`)` approach.
> > ```
> > async def publish(self, channel, message):
> >         """Publish a message on a channel
> > 
> >         Publish an arbitrary message asynchronously on a Pub/Sub channel.
> >         """
> >         for sub_id in self._subscriptions:
> >             event_filter_status = self.pubsub_event_filter(sub_id, message)
> >             if event_filter_status is False:
> >                 continue
> >         await self._redis.publish(channel, message)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It turns out that `self._redis.publish` will send messages to every subscriber of the channel. We can not select a particular subscriber to send the message based on the filter. According to the Redis documentation, publish does not work based on subscription id. Hence, we can not use this approach.
> 
> You need to use a different channel name for each subscriber. Then it will work.

The channel name would be the same irrespective of the filter passed by a subscriber.
To have a different channel for every different filter does not look feasible.

gctucker:
> > > > The main issue here is that:
> > > > 
> > > > * events should have just the operation type and node ID, as if we start sending full Node objects then they're not events any more but data being pushed to clients
> > > > * each subscriber has a different set of filters, but events are sent to global channels
> > > > * filtering in the `listen()` method can only be done using the fields in the event data
> > > > 
> > > > Some ideas about how to solve this (mutually exclusive ones):
> > > > 
> > > > * create a separate channel for each subscriber, and iterate over them when publishing events (with `{'node': [<list of actual Redis channels>]}`)
> > > > * pass the data filter in the event so it can be filtered in `listen()` but then probably drop it before actually sending it to the client
> > > > * logic could be put in kernelci-core to simplify client code, with all events still actually sent by the API
> > > 
> > > 
> > > I tried to solve the problem using `create a separate channel for each subscriber, and iterate over them when publishing events (with`{'node': []}`)` approach.
> > > ```
> > > async def publish(self, channel, message):
> > >         """Publish a message on a channel
> > > 
> > >         Publish an arbitrary message asynchronously on a Pub/Sub channel.
> > >         """
> > >         for sub_id in self._subscriptions:
> > >             event_filter_status = self.pubsub_event_filter(sub_id, message)
> > >             if event_filter_status is False:
> > >                 continue
> > >         await self._redis.publish(channel, message)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > It turns out that `self._redis.publish` will send messages to every subscriber of the channel. We can not select a particular subscriber to send the message based on the filter. According to the Redis documentation, publish does not work based on subscription id. Hence, we can not use this approach.
> > 
> > 
> > You need to use a different channel name for each subscriber. Then it will work.
> 
> The channel name would be the same irrespective of the filter passed by a subscriber. To have a different channel for every different filter does not look feasible.

It can be done this way: https://github.com/kernelci/kernelci-api/pull/120/files

It's obviously not great to add an indirection layer on top of Redis, but at least its pub/sub feature still provides a queuing mechanism. Then on top of that it's possible to add filters in the `publish()` method by adding filter parameters to each `sub_data`.

JenySadadia:
Created https://github.com/kernelci/kernelci-core/pull/1047.
Closing this one.

gctucker:
OK, closing https://github.com/kernelci/kernelci-api/pull/120 as well since it was just a basic proof-of-concept.

PR #105, opened 2022-01-26
Pub/Sub: raise exceptions when invalid id

gctucker:
Raise a `ValueError` exception when the subscription id is not valid.

### DIFF
--- a/test/test_listen_handler.py
+++ b/test/test_listen_handler.py
@@ -53,11 +53,11 @@ def test_listen_endpoint(mock_get_current_user,
                 active=True)
     mock_get_current_user.return_value = user
     mock_listen.return_value = 'Listening for events on channel 1'
+
     with TestClient(app) as client:
         response = client.get(
             "/listen/1",
             headers={
-                "Accept": "application/json",
                 "Authorization": "Bearer "
                 "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9."
                 "eyJzdWIiOiJib2IifQ.ci1smeJeuX779PptTkuaG1S"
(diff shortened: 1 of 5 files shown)


JenySadadia (changes requested):
We also need to catch ValueError in `listen` endpoint.

```
@app.get('/listen/{sub_id}')
async def listen(sub_id: int, user: User = Depends(get_user)):
    try:
        msg = await pubsub.listen(sub_id)
        if msg is None:
            raise HTTPException(
                status_code=status.HTTP_404_NOT_FOUND,
                detail=f"Not subscribed to channel with id: {sub_id}"
            )
        return msg
    except ValueError as error:
        raise HTTPException(
                status_code=status.HTTP_204_NO_CONTENT,
                detail=str(error)
        )
```

gctucker:
> We also need to catch ValueError in `listen` endpoint.
> 
> ```
> @app.get('/listen/{sub_id}')
> async def listen(sub_id: int, user: User = Depends(get_user)):
>     try:
>         msg = await pubsub.listen(sub_id)
>         if msg is None:
>             raise HTTPException(
>                 status_code=status.HTTP_404_NOT_FOUND,
>                 detail=f"Not subscribed to channel with id: {sub_id}"
>             )
>         return msg
>     except ValueError as error:
>         raise HTTPException(
>                 status_code=status.HTTP_204_NO_CONTENT,
>                 detail=str(error)
>         )
> ```

OK I've added that and updated unit tests accordingly.

gctucker:
@JenySadadia Could you please take another look?

test/test_listen_handler.py
```diff
@@ -108,7 +110,8 @@ def test_listen_endpoint_without_token(mock_get_current_user,
                                 'xCZGmM8jWXUXJZ4K',
                 active=True)
     mock_get_current_user.return_value = user
-    mock_listen.return_value = 'Listening for events on channel 1'
+    mock_listen.return_value = None
```
JenySadadia, line 113: We can omit `mock_listen` from this.

```
def test_listen_endpoint_without_token(mock_get_current_user,
                                       mock_init_sub_id):
    """
    Test Case : Test KernelCI API GET /listen endpoint for the
    negative path
    Expected Result :
        HTTP Response Code 401 Unauthorized
        The request requires user authentication by token in header
    """
    user = User(username='bob',
                hashed_password='$2b$12$CpJZx5ooxM11bCFXT76/z.o6HWs2sPJy4iP8.'
                                'xCZGmM8jWXUXJZ4K',
                active=True)
    mock_get_current_user.return_value = user

    with TestClient(app) as client:
        response = client.get(
            "/listen/1",
            headers={
                "Accept": "application/json"
            },
        )
        assert response.status_code == 401
```

test/test_listen_handler.py
```diff
@@ -68,22 +68,24 @@ def test_listen_endpoint(mock_get_current_user,
 
 
 def test_listen_endpoint_not_found(mock_get_current_user,
```
JenySadadia, line 70: We can add another assert statement to endpoint `test_listen_endpoint_not_found`:
```
assert 'detail' in response.json()
```

JenySadadia (changes requested):

JenySadadia (approved):

PR #106, opened 2022-01-28
api.models.Node: add artifacts

gctucker:
Add optional artifacts dictionary to the Node model as a mapping
between names and URLs to download files from the storage.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -5,7 +5,7 @@
 
 """KernelCI API model definitions"""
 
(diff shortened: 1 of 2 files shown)


gctucker:
A follow-up PR will be to add some documentation and validation as the artifacts should be a dictionary with names mapping to URLs.  Merging this now so we can unblock progress with the rest of the pipeline as it's already functional.

PR #107, opened 2022-01-31
test/test_node_handler: Remove 'request_dict' from a test case

JenySadadia:
'request_dict' is not being used in test_get_node_by_id_endpoint_empty_response.
Remove unnecessary code from the test function.

Fixes c933034 api/test_api.py: Add test_get_node_by_id_endpoint_empty_response


gctucker:
Looks like it would be good to enable Pylint on the unit test modules too :)

JenySadadia:
> Looks like it would be good to enable Pylint on the unit test modules too :)

Yes, I will create separate PR for that.

mgalka (approved):

PR #108, opened 2022-02-01
docker-compose.yaml: bind-mount volume with tests

gctucker:
Add a bind-mount volume with the 'test' directory to the api container
so unit tests can be run in the same environment.  To run the unit
tests by hand, one can now do:

  $ REQUIREMENTS=requirements-dev.txt docker-compose build
  $ docker-compose run api pytest test


gctucker:
@JenySadadia @mgalka What do you think of this?

mgalka (approved):
I like the idea of having a simple way to run tests. This can be used during development or in the automated CI process.

PR #109, opened 2022-02-02
Add 'created' field to Node model

JenySadadia:


api/models.py
```diff
@@ -7,6 +7,7 @@
 
 from typing import Optional, Dict
 
+from datetime import datetime
```
gctucker, line 10:
```suggestion
from datetime import datetime
from typing import Optional, Dict

```
The imports should be grouped by category: first the ones from the standard Python library, then additional packages and finally local ones from the current package. We don't strictly need blank lines but it helps making the distinction. Also, keeping them in alphabetical order avoids adding duplicates and it's easier to maintain.

test/test_node_handler.py
```diff
@@ -124,6 +124,7 @@ def test_create_node_endpoint(mock_get_current_user, mock_init_sub_id,
             'parent',
             'revision',
             'status',
+            'created',
```
gctucker, line 127: I also like to put the fields in alphabetical order, just to keep things easier to maintain. But that's not a requirement.

gctucker (approved):

PR #110, opened 2022-02-03
test/test_pubsub.py: test publish_cloudevent json

alexandrasp:
Include a test to validate if the json is generated and its attributes are correct.

Signed-off-by: Alexandra Pereira <alexandra.pereira@collabora.com>

### DIFF
--- a/test/test_pubsub.py
+++ b/test/test_pubsub.py
@@ -2,11 +2,16 @@
 #
 # Copyright (C) 2022 Collabora Limited
 # Author: Michal Galka <michal.galka@collabora.com>
+# Author: Alexandra Pereira <alexandra.pereira@collabora.com>
 
 # pylint: disable=protected-access
 
 """Unit test functions for KernelCI API Pub/Sub"""
 
+from unittest import result
+from api.pubsub import PubSub
+import fakeredis.aioredis
+import json
 import pytest
 
 
@@ -80,3 +85,37 @@ async def test_unsubscribe_sub_id_not_exists(mock_pubsub_subscriptions):
         await mock_pubsub_subscriptions.unsubscribe(sub_id=2)
     assert len(mock_pubsub_subscriptions._subscriptions) == 1
     assert 1 in mock_pubsub_subscriptions._subscriptions
+
+
+@pytest.fixture()
+def mock_pubsub_publish(mocker):
+    pubsub = PubSub()
+    redis_mock = fakeredis.aioredis.FakeRedis()
+    mocker.patch.object(pubsub, '_redis', redis_mock)
+    mocker.patch.object(pubsub._redis, 'execute_command')
+    return pubsub
+
+
+@pytest.mark.asyncio
+async def test_pubsub_publish_couldevent(mock_pubsub_publish):
+    """
+    Test Case: Validate and check the json built by the cloud event
+    published in the channel by the redis publisher.
+
+    Expected Results:
+        Validate that a json is sent to the channel and assert the json values from
+        data and attributes parameters in Pubsub.publish_cloudevent(). There's no
+        return value, but a json to be published in a channel.
+    """
+
+    data = 'validate json'
+    attributes = { "specversion": "1.0",  "id": "6878b661-96dc-4e93-8c92-26eb9ff8db64", "source": "https://api.kernelci.org/", "type": "api.kernelci.org", "time": "2022-01-31T21:29:29.675593+00:00"}
+
+    await mock_pubsub_publish.publish_cloudevent('CHANNEL1', data, attributes)
+
+    expected_json = str.encode('{"specversion": "1.0", "id": "6878b661-96dc-4e93-8c92-26eb9ff8db64", "source": "https://api.kernelci.org/", "type": "api.kernelci.org", "time": "2022-01-31T21:29:29.675593+00:00", "data": "validate json"}')
+
+    json_arg = mock_pubsub_publish._redis.execute_command.call_args.args[2]
+
+    json.loads(json_arg)
+    assert json_arg == expected_json
--- a/test/test_pubsub.py
+++ b/test/test_pubsub.py
@@ -8,9 +8,6 @@
 
 """Unit test functions for KernelCI API Pub/Sub"""
 
-from unittest import result
-from api.pubsub import PubSub
-import fakeredis.aioredis
 import json
 import pytest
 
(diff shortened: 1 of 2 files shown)


test/test_pubsub.py
```diff
@@ -3,8 +3,15 @@
 # Copyright (C) 2022 Collabora Limited
 # Author: Michal Galka <michal.galka@collabora.com>
 
+# Copyright (C) 2022 Collabora Limited
+# Author: Alexandra Pereira <alexandra.pereira@collabora.com>
+
```
mgalka, line 8:
```suggestion
# Author: Alexandra Pereira <alexandra.pereira@collabora.com>
```

There is no need to add another `Copyright` header. Just add yourself as an author.

test/test_pubsub.py
```diff
@@ -3,8 +3,15 @@
...
+from unittest import result
+
```
mgalka, line 10: Remove the extra empty line and sort imports alphabetically.

test/test_pubsub.py
```diff
@@ -99,3 +106,41 @@ async def test_unsubscribe_sub_id_not_exists(mock_pubsub_subscriptions):
     assert len(mock_pubsub_subscriptions._subscriptions) == 1
     assert 1 in mock_pubsub_subscriptions._subscriptions
     assert result is False
+
+
+@pytest.fixture()
+def mock_pubsub_publish(mocker):
+    pubsub = PubSub()
+    redis_mock = fakeredis.aioredis.FakeRedis()
+    mocker.patch.object(pubsub, '_redis', redis_mock)
```
mgalka, line 115: These lines basically mock redis connection and are the same as ones in `mock_pubsub_subscriptions()`. Maybe we could extract them to a separate function and use in both cases?

test/test_pubsub.py
```diff
@@ -99,3 +106,41 @@ async def test_unsubscribe_sub_id_not_exists(mock_pubsub_subscriptions):
...
+    await mock_pubsub_publish.publish_cloudevent('CHANNEL1', data, attributes)
+
+    expected_json = str.encode('{"specversion": "1.0", "id": "6878b661-96dc-4e93-8c92-26eb9ff8db64", "source": "https://api.kernelci.org/", "type": "api.kernelci.org", "time": "2022-01-31T21:29:29.675593+00:00", "data": "validate json"}')
+
+    json_arg = mock_pubsub_publish._redis.execute_command.call_args.args[2]
+
+    try:
+        json.loads(json_arg)
+    except json.JSONDecodeError:
+        assert False
```
mgalka, line 144: I don't quite get the idea of `assert False` here.
If `json_arg` is supposed to be a valid JSON object there should be neither `try..except` block nor `assert False`. 
If for some reason `json.loads` raises an exception the test should end up with `Error` status while `assert False` will mark it as `Failed`.

mgalka (changes requested):

test/test_pubsub.py
alexandrasp: done

test/test_pubsub.py
alexandrasp: thanks for noticing this, you're right, so I removed the `try..except` block and now the test will just fail in case of any problem with the JSON.

test/test_pubsub.py
```diff
@@ -2,20 +2,28 @@
...
     redis_mock = fakeredis.aioredis.FakeRedis()
     mocker.patch.object(pubsub, '_redis', redis_mock)
     return pubsub
 
 
+@pytest.fixture()
+def mock_pubsub(mocker):
+    return prepare_pubsub_mock(mocker)
+
+
```
mgalka, line 26: Is there a reason why we need a `prepare_pubsub_mock()` function instead of keeping the code in the fixture?

alexandrasp: it makes sense, no special reason and fixed

test/test_pubsub.py
```diff
@@ -51,6 +53,13 @@ async def test_subscribe_multiple_channels(mock_pubsub):
     assert (1, 2, 3) == tuple(mock_pubsub._subscriptions.keys())
 
 
+@pytest.fixture()
+def mock_pubsub_subscriptions(mocker, mock_pubsub):
```
mgalka, line 57: Something to check. I looks that there is no need to inherit after `mocker here`

You could try to apply this change and see if it works.

```suggestion
def mock_pubsub_subscriptions(mock_pubsub):
```

test/test_pubsub.py
```diff
@@ -80,3 +89,39 @@ async def test_unsubscribe_sub_id_not_exists(mock_pubsub_subscriptions):
         await mock_pubsub_subscriptions.unsubscribe(sub_id=2)
     assert len(mock_pubsub_subscriptions._subscriptions) == 1
     assert 1 in mock_pubsub_subscriptions._subscriptions
+
+
+@pytest.fixture()
+def mock_pubsub_publish(mocker, mock_pubsub):
```
mgalka, line 95: same as previous comment

mgalka (approved):

test/test_pubsub.py
```diff
@@ -51,6 +53,13 @@ async def test_subscribe_multiple_channels(mock_pubsub):
     assert (1, 2, 3) == tuple(mock_pubsub._subscriptions.keys())
 
 
+@pytest.fixture()
+def mock_pubsub_subscriptions(mocker, mock_pubsub):
+    subscriptions_mock = dict({1: mock_pubsub._redis.pubsub()})
+    mocker.patch.object(mock_pubsub, '_subscriptions', subscriptions_mock)
+    return mock_pubsub
+
+
```
mgalka, line 62: Please move fixture to `conftest.py`. Please, mind that there is already a fixture named `mock_pubsub_subscriptions()`.

mgalka: Add the docstring

test/test_pubsub.py
```diff
@@ -80,3 +89,39 @@ async def test_unsubscribe_sub_id_not_exists(mock_pubsub_subscriptions):
         await mock_pubsub_subscriptions.unsubscribe(sub_id=2)
     assert len(mock_pubsub_subscriptions._subscriptions) == 1
     assert 1 in mock_pubsub_subscriptions._subscriptions
+
+
+@pytest.fixture()
+def mock_pubsub_publish(mocker, mock_pubsub):
+    mocker.patch.object(mock_pubsub._redis, 'execute_command')
+    return mock_pubsub
```
mgalka, line 97: Move the fixture to the `conftest.py`

mgalka: Add missing docstring

mgalka (changes requested):
Please fix the issues detected by pylint (see the comments)

mgalka (approved):

PR #111, opened 2022-02-03
Add 'updated' field to Node model

JenySadadia:

Depends on #109

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -87,3 +87,4 @@ class Node(ModelId):
     status: Optional[StatusValues] = StatusValues.PENDING
     artifacts: Optional[Dict]
     created: Optional[datetime] = Field(default_factory=datetime.utcnow)
+    updated: Optional[datetime] = Field(default_factory=datetime.utcnow)
--- a/api/models.py
+++ b/api/models.py
@@ -1,7 +1,8 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 #
-# Copyright (C) 2021 Collabora Limited
+# Copyright (C) 2021, 2022 Collabora Limited
 # Author: Guillaume Tucker <guillaume.tucker@collabora.com>
+# Author: Jeny Sadadia <jeny.sadadia@collabora.com>
 
 """KernelCI API model definitions"""
 
@@ -62,6 +63,12 @@ class ModelId(BaseModel):
 # Database models
 #
 
+class DatabaseModel(ModelId):
+    """Database model"""
+    def update(self):
+        """Method to update model"""
+
+
 class User(ModelId):
     """API user model"""
     username: str
@@ -78,7 +85,7 @@ class Revision(BaseModel):
     describe: Optional[str] = None
 
 
-class Node(ModelId):
+class Node(DatabaseModel):
     """KernelCI primitive node object model for generic test results"""
     kind: str = 'node'
     name: str
@@ -88,3 +95,6 @@ class Node(ModelId):
     artifacts: Optional[Dict]
     created: Optional[datetime] = Field(default_factory=datetime.utcnow)
     updated: Optional[datetime] = Field(default_factory=datetime.utcnow)
+
+    def update(self):
+        self.updated = datetime.utcnow()
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ $ curl -X 'POST' \
   "commit":"2a987e65025e2b79c6d453b78cb5985ac6e5eb26",
   "describe":"v5.16-rc4-31-g2a987e65025e"}
 }'
-{"_id":"61bda8f2eb1a63d2b7152418","kind":"node","name":"checkout","revision":{"tree":"mainline","url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git","branch":"master","commit":"2a987e65025e2b79c6d453b78cb5985ac6e5eb26","describe":"v5.16-rc4-31-g2a987e65025e"},"parent":null,"status":"pending", "created":"2022-02-02T11:23:03.157648"}
+{"_id":"61bda8f2eb1a63d2b7152418","kind":"node","name":"checkout","revision":{"tree":"mainline","url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git","branch":"master","commit":"2a987e65025e2b79c6d453b78cb5985ac6e5eb26","describe":"v5.16-rc4-31-g2a987e65025e"},"parent":null,"status":"pending", "created":"2022-02-02T11:23:03.157648", "updated":"2022-02-02T11:23:03.157648"}
 ```
 
 ### Getting Nodes back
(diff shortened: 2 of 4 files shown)


gctucker:
Please rebase now that the PR to add `created` has been merged.

JenySadadia:
> Please rebase now that the PR to add `created` has been merged.

Done.

gctucker:
This looks like a good start but there's nothing to actually change the `updated` time when the object is saved again in the database.

JenySadadia:
After the code changes, I observed that, 'updated' value of object returned by put_node and the value stored in DB differs. 
```
$ curl -X 'PUT'   'http://localhost:8001/node/621319d8a84ca10f0bffe3af'   -H 'accept: application/json'   -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJib2IifQ.tYsu5nu2njDzrkrf_sVK1BRzCL6pUuop-YQuFnZ2xLo'   -H 'Content-Type: application/json'   -d '{
  "name":"checkout","created":"2022-02-21T04:49:28.139000","updated":"2022-02-21T04:41:28.139000",
  "revision":{"tree":"mainline",
  "url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git",
  "branch":"master",
  "commit":"2a987e65025e2b79c6d453b78cb5985ac6e5eb26",
  "describe":"v5.16-rc4-31-g2a987e65025e"}
}'
{"_id":"621319d8a84ca10f0bffe3af","kind":"node","name":"checkout","revision":{"tree":"mainline","url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git","branch":"master","commit":"2a987e65025e2b79c6d453b78cb5985ac6e5eb26","describe":"v5.16-rc4-31-g2a987e65025e"},"parent":null,"status":null,"artifacts":null,"created":"2022-02-21T04:49:28.139000","updated":"2022-02-21T07:15:07.541032"}


$ curl http://localhost:8001/nodes
[{"_id":"621319d8a84ca10f0bffe3af","kind":"node","name":"checkout","revision":{"tree":"mainline","url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git","branch":"master","commit":"2a987e65025e2b79c6d453b78cb5985ac6e5eb26","describe":"v5.16-rc4-31-g2a987e65025e"},"parent":null,"status":null,"artifacts":null,"created":"2022-02-21T04:49:28.139000","updated":"2022-02-21T07:15:07.541000"}]
```

I am trying to debug the issue.

JenySadadia:
Added commit to return the actual object from DB after update operation.

JenySadadia:
> After the code changes, I observed that, 'updated' value of object returned by put_node and the value stored in DB differs.
> 
> ```
> $ curl -X 'PUT'   'http://localhost:8001/node/621319d8a84ca10f0bffe3af'   -H 'accept: application/json'   -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJib2IifQ.tYsu5nu2njDzrkrf_sVK1BRzCL6pUuop-YQuFnZ2xLo'   -H 'Content-Type: application/json'   -d '{
>   "name":"checkout","created":"2022-02-21T04:49:28.139000","updated":"2022-02-21T04:41:28.139000",
>   "revision":{"tree":"mainline",
>   "url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git",
>   "branch":"master",
>   "commit":"2a987e65025e2b79c6d453b78cb5985ac6e5eb26",
>   "describe":"v5.16-rc4-31-g2a987e65025e"}
> }'
> {"_id":"621319d8a84ca10f0bffe3af","kind":"node","name":"checkout","revision":{"tree":"mainline","url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git","branch":"master","commit":"2a987e65025e2b79c6d453b78cb5985ac6e5eb26","describe":"v5.16-rc4-31-g2a987e65025e"},"parent":null,"status":null,"artifacts":null,"created":"2022-02-21T04:49:28.139000","updated":"2022-02-21T07:15:07.541032"}
> 
> 
> $ curl http://localhost:8001/nodes
> [{"_id":"621319d8a84ca10f0bffe3af","kind":"node","name":"checkout","revision":{"tree":"mainline","url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git","branch":"master","commit":"2a987e65025e2b79c6d453b78cb5985ac6e5eb26","describe":"v5.16-rc4-31-g2a987e65025e"},"parent":null,"status":null,"artifacts":null,"created":"2022-02-21T04:49:28.139000","updated":"2022-02-21T07:15:07.541000"}]
> ```
> 
> I am trying to debug the issue.

I checked the value stored in the database.
It matches with the output from `/nodes` endpoint.
I think we should return the actual object from `update` method. @gctucker 
```
$ docker-compose exec db /bin/mongo kernelci --eval   "db.node.find()"
MongoDB shell version v5.0.5
connecting to: mongodb://127.0.0.1:27017/kernelci?compressors=disabled&gssapiServiceName=mongodb
Implicit session: session { "id" : UUID("498abc08-ef20-44d6-8cf6-0524f3925cfd") }
MongoDB server version: 5.0.5
{ "_id" : ObjectId("621319d8a84ca10f0bffe3af"), "kind" : "node", "name" : "checkout", "revision" : { "tree" : "mainline", "url" : "https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git", "branch" : "master", "commit" : "2a987e65025e2b79c6d453b78cb5985ac6e5eb26", "describe" : "v5.16-rc4-31-g2a987e65025e" }, "parent" : null, "status" : null, "artifacts" : null, "created" : ISODate("2022-02-21T04:49:28.139Z"), "updated" : ISODate("2022-02-21T07:15:07.541Z") }
```

gctucker:
Yes I think the returned data from `PUT` should be exactly the same as when doing a `GET`, or nothing at all.

There's actually a very long discussion about this here:
https://stackoverflow.com/questions/797834/should-a-restful-put-operation-return-something

RFC 2616 doesn't require any body in a response to `PUT` so I think the best solution here is probably to not return the updated node. If the client wants to get the node again, it can do that with a `GET` request.

api/models.py
```diff
@@ -68,7 +76,7 @@ class Revision(BaseModel):
     describe: Optional[str] = None
 
 
-class Node(ModelId):
+class Node(DatabaseModel):
```
gctucker, line 79: Looking at the documentation:
https://docs.python.org/3/library/abc.html

It seems that `.register()` should be used for concrete sub-classes.

I was originally thinking of using "old-style" abstract classes by just not defining an implementation:
```python
class DatabaseModel(ModelId):

    def update(self):
        pass
```
Then models can inherit from it without using the `abc` module, and we can just call `obj.update()` on any of them.

@mgalka Do you think using `abc` is useful in this case?

gctucker:
Adding skip label so we can focus on merging the status enum changes first.

api/models.py
JenySadadia: What do you suggest on this @mgalka ?

JenySadadia: > Looking at the documentation: https://docs.python.org/3/library/abc.html
> 
> It seems that `.register()` should be used for concrete sub-classes.
> 
> I was originally thinking of using "old-style" abstract classes by just not defining an implementation:
> 
> ```python
> class DatabaseModel(ModelId):
> 
>     def update(self):
>         pass
> ```
> 
> Then models can inherit from it without using the `abc` module, and we can just call `obj.update()` on any of them.
> 
> @mgalka Do you think using `abc` is useful in this case?

@gctucker The documentation https://docs.python.org/3/library/abc.html states that we will need to use `register()` method in case of virtual subclasses. On the contrary `Node` model explicitly inherits `DatabaseModel`. Hence, we do not need the method.

JenySadadia: If we go with the old-style implementation, we will be able to initiate DatabaseModel class. It won't be a pure abstract class then. It will be an inheritance instead. @mgalka @gctucker

mgalka: As far as I'm concerned, I'd keep the `ABC` mixin in the `DatabaseModel`. 

According to what I've read in the the previous comments there is an idea of having `Database.update()` method that'll call `update()` methods from models. Having `DatabaseModel` declared as an ABC will ensure that the `update()` method is actually implemented. Attempts to instantiate objects with unimplemented abstract methods can also be easily detected by static code checkers like `pylint` or `mypy`.
Although, having the "fully fledged" abstract class instead of an "old-style" one may seem a bit of an overhead as it has just one method it brings value to the project. I'd keep it.

@gctucker I'm not aware of a need of using `.register()` method unless we want to register our class as a sub-class of some unrelated class. It's more of a concept when regular inheritance doesn't seem quite right or e.g. we want to use a built-in class as a virtual class for our.

@JenySadadia I'd rename method `.set_updated()` to `.update()`. It looks more appropriate as the former suggest that it's implementation will be limited setting the `updated` class attribute and as far as I understand the idea behind the `.update()` method is to perform all kinds of actions that we need when updating the object in the database.

JenySadadia: Yes, that makes sense. @mgalka 
Renamed the method.

gctucker: Yes, actually my example is not really an abstract class since it has a default implementation (which does nothing).  A typical old-style abstract method with no implementation would raise `NotImplementedError`, like in some other places in kernelci-core.

mgalka (approved):

PR #119, opened 2022-02-17
gitignore: add ssh and storage data directories

gctucker:
Add the ssh user-data and storage data directories to .gitignore as
they contain data files which should not be kept under source control.


gctucker (commented):
Tested OK on staging.

Issue #118, opened 2022-02-16
Implement endpoint to get root node


JenySadadia:
In order to generate a test report, we need the root node's revision information.
We can fetch the root node by recursively calling `db.find_by_id` method using `Node.parent` attribute.

PR #121, opened 2022-02-22
api/main.py: implement endpoint to get root node

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/118

Add 'get_root_node/' endpoint to get root node
by providing node id to it.
It uses Node.parent attribute to fetch root node.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -10,7 +10,7 @@ from .db import Database
 from .models import Node, User
 from .pubsub import PubSub, Subscription
 from typing import List
-from bson import ObjectId
+from bson import ObjectId, errors
 
 app = FastAPI()
 db = Database()
@@ -88,6 +88,25 @@ async def get_node(node_id: str):
 @app.get('/nodes', response_model=List[Node])
 async def get_nodes(request: Request):
     return await db.find_by_attributes(Node, dict(request.query_params))
+
+
+@app.get('/get_root_node/{node_id}', response_model=Node)
+async def get_root_node(node_id: str):
+    while node_id:
+        try:
+            node = await db.find_by_id(Node, node_id)
+        except errors.InvalidId as error:
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail=str(error)
+            )
+        if node is None:
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail="Node not found with id: {}".format(node_id)
+            )
+        node_id = node.parent
+    return node
 
 
 @app.post('/node', response_model=Node)


gctucker:
LGTM. Have you tried using it in some client code, such as the test report generator?

We might need to add some unit test for this endpoint too.

JenySadadia:
I will need to create `KernelCI_API.get_root_node` in `kernelci-core` to test it in client code.
I will post the result here.

api/main.py
```diff
@@ -119,6 +119,18 @@ async def put_node(node_id: str, node: Node, token: str = Depends(get_user)):
...
 
+@app.get('/get_root_node/{node_id}', response_model=Node)
+async def get_root_node(node_id: str):
+    while True:
+        node = await db.find_by_id(Node, node_id)
+        if node is None:
+            break
+        node_id = node.parent
+        if node_id is None:
+            break
```
mgalka, line 130: I think we could change it something like that. It's just a bit more concise. Having said that, this code requires Python 3.8+


```suggestion
    while (node := await db.find_by_id(Node, node_id)) is not None:
        node_id = node.parent
        if node_id is None:
            break
```

gctucker: Shouldn't the loop be doing a `while node_id is not None`? That's the actual condition for ending the loop, then there's no parent so no node_id. If node is None, that's more like an error if an id is not valid.

JenySadadia: Yes, `if node_id is not None` is the condition until we need to loop through the nodes. In case of invalid `node_id`, we need to catch an `InvalidId` exception. When `node_id` is valid but no node matching with the id exists, we will get `null` output.

gctucker: OK, by "invalid id" I meant an id that doesn't match any node. That should not happen, unless there's an error in the database or a corrupt id. So it should be handled as an error case, but the actual logic is to keep looping as long as there is a node id. Something like:
```
while node_id:
  node = db.get_node(node_id)
  node_id = node.parent
```
and everything else is error handling (i.e. node being None). We could add a check in the middle to `break` if node is None, or catch exceptions around the loop to better separate the actual logic of the code from error handling.

api/main.py
```diff
@@ -90,6 +90,28 @@ async def get_nodes(request: Request):
...
+    except errors.InvalidId as error:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=str(error)
+        )
+    except ValueError as error:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=str(error)
+        )
```
mgalka, line 112: I'd rather limit the number of lines in the `try..except` block to the very minimum. 
I think there is no need to `raise` the `ValueError` and catch it immediately to raise the HTTPException. I think we can just raise the latter in case of an error.

```suggestion
        while node_id:
            try:
                node = await db.find_by_id(Node, node_id)
            except errors.InvalidId as error:
                raise HTTPException(
                    status_code=status.HTTP_400_BAD_REQUEST,
                    detail=str(error)
                )
            
            if node is None:
                raise HTTPException(
                    status_code=status.HTTP_400_BAD_REQUEST,
                    detail=('Node not found with id: {}'.format(node_id)
                )
            node_id = node.parent
        return node
```

gctucker:
OK I see this is being used in https://github.com/kernelci/kernelci-pipeline/pull/24

mgalka (approved):

PR #123, opened 2022-02-28
api/models.py: Modify Node.status schema

JenySadadia:
As we can have multiple values for Node.status like pending, pass,
and fail. Hence, we need to define status as Enum type.
Need to add 'use_enum_values' to ModelID to populate
the status field with 'value' property of Enums.

Fixes: f9bf5b9 ("api.models: add Node model")


gctucker (commented):
LGTM, this will get tested on staging then we'll be able to check the logs.

One small thing though, the commit subject "api/models.py: Modify Node.status schema" is a bit vague. How about "convert Node.status to enum"?

JenySadadia:
> 

Yes, it seems more precise.

gctucker:
Adding skip label as some client-side changes are still required.

PR #124, opened 2022-03-01
Unit tests for get_root_node/ endpoint

JenySadadia:


test/test_node_handler.py
```diff
@@ -308,3 +309,88 @@ def test_get_all_nodes_empty_response(mock_get_current_user,
...
+        HTTP Response Code 400 Bad Request
+        JSON with 'detail' key
+    """
+    mock_db_find_by_id.side_effect = errors.InvalidId
+
+    with TestClient(app) as client:
+        response = client.get("/get_root_node/61bda8f2eb1a63d2b71524")
+        print("response.json()", response.json())
+        assert response.status_code == 400
+        assert 'detail' in response.json()
```
mgalka, line 396: We could check if the response `detail` value has correct `node_id` included.

JenySadadia: In that case we need to pass error message in `mock_db_find_by_id.side_effect = errors.InvalidId`.
At this time we are getting `{'detail': ''}` response json as we have not provided any error string.

JenySadadia:
> Please, rebase the PR with the current main

Done rebasing.

PR #125, opened 2022-03-01
api.pubsub: add keep-alive periodic events

gctucker:
Send events periodically on all active channels to keep TCP
connections alive.  This is to avoid proxies and firewall closing them
due to inactivity.


gctucker:
Tested OK on staging.

PR #134, opened 2022-04-27
api.models: add Node.translate_fields() method

gctucker:
Rewrite Node.replace_parent() method with a more generic one to
translate all applicable fields to objects such as database IDs and
Enum values.  Update how it's used in api.main accordingly to keep the
code in related to the model's fields in api.models.


gctucker:
@JenySadadia Actually I don't see any code in the pipeline or any unit test making use of this.  The only API calls currently handled by the `/nodes` endpoint use `revision.commit` from the `.get_nodes_by_commit_hash()` method in `kernelci-core`. It might be a good idea to add a unit test for it so it doesn't break without anyone noticing.  I think it will be used at some point though.

JenySadadia (commented):
Request: `curl 'http://localhost:8001/nodes?status=pending'`

Need to fix the error:
```
kernelci-api | bson.errors.InvalidDocument: cannot encode object: <StatusValues.PENDING: 'pending'>, of type: <enum 'StatusValues'>
```

JenySadadia:
> @JenySadadia Actually I don't see any code in the pipeline or any unit test making use of this. The only API calls currently handled by the `/nodes` endpoint use `revision.commit` from the `.get_nodes_by_commit_hash()` method in `kernelci-core`. It might be a good idea to add a unit test for it so it doesn't break without anyone noticing. I think it will be used at some point though.

I have already added `test_get_node_by_attributes_endpoint` in test_node_handler.py file.

JenySadadia:
> Request: `curl 'http://localhost:8001/nodes?status=pending'`
> 
> Need to fix the error:
> 
> ```
> kernelci-api | bson.errors.InvalidDocument: cannot encode object: <StatusValues.PENDING: 'pending'>, of type: <enum 'StatusValues'>
> ```

The issue was only for `parent` attribute. As StatusValues have string enum values, we don't need any external type conversion for `status`. The `nodes/status=pending` was working before.

gctucker:
> > Request: `curl 'http://localhost:8001/nodes?status=pending'`
> > Need to fix the error:
> > ```
> > kernelci-api | bson.errors.InvalidDocument: cannot encode object: <StatusValues.PENDING: 'pending'>, of type: <enum 'StatusValues'>
> > ```
> 
> The issue was only for `parent` attribute. As StatusValues have string enum values, we don't need any external type conversion for `status`. The `nodes/status=pending` was working before.

OK thanks for testing that.  One thing it does is validating the status string, so if the query has an invalid status it would cause an error.  But I guess we should have another method for that e.g. `.validate()`.

I've updated this PR to drop the part about `StatusValues`.  I think the unit test should be extended to try other fields like `status` etc.

api/main.py
```diff
@@ -87,10 +87,9 @@ async def get_node(node_id: str):
 
 @app.get('/nodes', response_model=List[Node])
 async def get_nodes(request: Request):
-    query_param_dict = dict(request.query_params)
-    if 'parent' in query_param_dict:
-        query_param_dict = Node.modify_parent(query_param_dict)
-    return await db.find_by_attributes(Node, query_param_dict)
+    query_params = dict(request.query_params)
+    translated_params = Node.translate_fields(query_params)
```
JenySadadia, line 91: To reduce the LoC, I would suggest,

translated_params = Node.translate_fields(dict(request.query_params))

api/models.py
```diff
@@ -109,7 +109,16 @@ def update(self):
...
 
     @classmethod
-    def modify_parent(cls, query_param_dict):
-        """Modify query parameter dictionary for parent"""
-        query_param_dict['parent'] = ObjectId(query_param_dict['parent'])
-        return query_param_dict
+    def translate_fields(cls, params: dict):
+        """Translate fields in `params` into objects as applicable
+
+        Translate fields reprensted by strings in the `params` dictionary into
```
JenySadadia, line 115: Need to fix a minor typo.
```suggestion
        Translate fields represented by strings in the `params` dictionary into
```

api/models.py
```diff
@@ -109,7 +109,16 @@ def update(self):
...
+        objects that match the model.  For example, database IDs are converted
+        to ObjectId.  Return a new dictionary with the translated values
+        replaced.
+        """
+        translated = params.copy()
```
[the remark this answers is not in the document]
gctucker: This is to avoid modifying the input argument. I've stopped making functions that do that as they can lead to very subtle issues. Modifying data in-place should be done by methods of an object instead.

api/main.py
gctucker: I split this into 2 to make it easier to read and keep the line short.  Sometimes a smaller number of lines actually makes the code harder to read or maintain.

PR #136, opened 2022-05-02
T35091 Add method to validate request parameters

JenySadadia:
api/models.py: add 'validate_params' method
api/main.py: validate params in '/nodes' endpoint

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -9,7 +9,7 @@
 from datetime import datetime
 from typing import Optional, Dict
 import enum
-from bson import ObjectId
+from bson import ObjectId, errors
 from pydantic import BaseModel, Field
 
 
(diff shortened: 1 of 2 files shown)


api/models.py
```diff
@@ -113,3 +113,19 @@ def modify_parent(cls, query_param_dict):
         """Modify query parameter dictionary for parent"""
         query_param_dict['parent'] = ObjectId(query_param_dict['parent'])
         return query_param_dict
+
+    @classmethod
+    def validate_params(cls, params: dict):
+        """Validate Node parameters"""
+        status = params.get('status')
+        if status:
+            if status not in list(status.value for status in StatusValues):
```
mgalka, line 122: maybe we could use `and` operator instead of a nested `if`?

mgalka (commented):
Looks good, one minor issue that may be worth fixing.

api/models.py
JenySadadia: Yes, also replaced `list( )` with `[ ]` as it affects the execution speed.

mgalka (approved):
LGTM

PR #138, opened 2022-05-03
T35149 Update API docs & README.md

JenySadadia:
Keep minimal instructions in README.md and move other documentation to kernelci.org/docs.


doc/getting-started.md
```diff
@@ -139,6 +139,122 @@ $ curl -X 'GET' \
 {"_id":"615f30020eb7c3c6616e5ac3","username":"bob","hashed_password":"$2b$12$VtfVij6zz20F/Qr0Ri18O.11.0LJMMXyJxAJAHQbKU0jC96eo2fr.","active":true}
 ```
 
+### Create a Node
```
gctucker, line 142: I think this should be put into a separate section about Node object as it's not really something users have to know in order to get started.  A typical starting point is to have the API and pipeline containers running rather than using `curl` with the raw API.

doc/getting-started.md
```diff
@@ -139,6 +139,122 @@ $ curl -X 'GET' \
...
+  "url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git",
+  "branch":"master",
+  "commit":"2a987e65025e2b79c6d453b78cb5985ac6e5eb26",
+  "describe":"v5.16-rc4-31-g2a987e65025e"},
+  "created":"2022-02-02T11:23:03.157648"
+}'
+{"_id":"61bda8f2eb1a63d2b7152418","kind":"node","name":"checkout-test","revision":{"tree":"mainline","url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git","branch":"master","commit":"2a987e65025e2b79c6d453b78cb5985ac6e5eb26","describe":"v5.16-rc4-31-g2a987e65025e"},"parent":null,"status":"pending", "created":"2022-02-02T11:23:03.157648", "updated":"2022-02-02T12:23:03.157648"}
+```
+
+### Listen & Publish CloudEvent
```
gctucker, line 211: Same idea, this could be moved to a pub/sub section in the documentation.

README.md
```diff
@@ -72,124 +72,17 @@ use it with other tools: http://localhost:8001/openapi.json
 As a proof-of-concept, an object model called `Node` is defined in this API.
 It's possible to create new objects and retrieve them via the API.
 
-### Creating a Node
```
gctucker, line 75: We can probably just drop the Node and Pub/Sub parts from the README as it's already pretty specialised. If we mention this here we would need to mention all the aspects of the API as a table of contents, which is something we'll have in the main docs anyway.

JenySadadia: Agreed. Updated the README file.

doc/getting-started.md
JenySadadia: Added separate sections. Thanks.

gctucker:
It's still in the "getting started" section:
https://static.staging.kernelci.org/docs/api/getting-started/#nodes

You've added a title for nodes and pub/sub but I was thinking it could have its own separate page to cover details about the API.

JenySadadia:
> It's still in the "getting started" section: https://static.staging.kernelci.org/docs/api/getting-started/#nodes
> 
> You've added a title for nodes and pub/sub but I was thinking it could have its own separate page to cover details about the API.

I have added new page `API details`.

doc/getting-started.md
```diff
@@ -38,6 +38,16 @@ $ echo SECRET_KEY=$(openssl rand -hex 32) >> .env
...
+#### Set ALGORITHM and ACCESS_TOKEN_EXPIRE_MINUTES in environment file
+
+We need to specify an algorithm for JWT token encoding and decoding. ALGORITHM variable needs to be passed in the parameter for that.
+ALGORITHM is set default to HS256.
+We have used ACCESS_TOKEN_EXPIRE_MINUTES variable to set expiry time on generated jwt access token.
+ACCESS_TOKEN_EXPIRE_MINUTES is set default to None.
+If a user wants to change any of the above variables, they should be added to the .env file.
+
+Please find the [env.sample](https://github.com/kernelci/kernelci-api/blob/main/env.sample) file for the reference.
+
```
gctucker, line 50: This is not actually required to get started, which is why I didn't include it here.  It's something that could be added in the full API documentation under a section about environment variables.

JenySadadia: Okay, I'll add it to `API details` page.

gctucker (approved):
OK https://static.staging.kernelci.org/docs/api/api-details/

PR #139, opened 2022-05-06
T35229 Add field definitions to data model

### LINKED ISSUE #41: Add detailed definition of each field in the data models
The data models currently just have a field name and a type, and `Optional()` when appropriate.  Adding some description of what each field is about, and some more specific type such as URL validation rather than a plain string for URLs would make it easier to use from the point of view of API clients.  This should also automatically be included in the generated documentation.

For example, URLs: https://pydantic-docs.helpmanual.io/usage/types/#urls

A model using the `Field()` constructor to pass extra information:
https://pydantic-docs.helpmanual.io/usage/schema/

See also:
https://fastapi.tiangolo.com/tutorial/body-fields/?h=field#declare-model-attributes
```python
class Item(BaseModel):
    name: str
    description: Optional[str] = Field(
        None, title="The description of the item", max_length=300
    )
    price: float = Field(..., gt=0, description="The price must be greater than zero")
    tax: Optional[float] = None
```

JenySadadia:
Fixes #41

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -10,7 +10,7 @@ from datetime import datetime
 from typing import Optional, Dict
 import enum
 from bson import ObjectId, errors
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, SecretStr, HttpUrl
 
 
 class PyObjectId(ObjectId):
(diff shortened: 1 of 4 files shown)


api/models.py
```diff
@@ -81,29 +81,37 @@ def update(self):
...
-    url: str
-    branch: str
-    commit: str
-    describe: Optional[str] = None
+    tree: str = Field(description='git tree of the revision')
+    url: HttpUrl = Field(description='git URL of the revision')
+    branch: str = Field(description='git branch of the revision')
+    commit: str = Field(description='git commit SHA of the revision')
+    describe: Optional[str] = Field(None, description='git describe of \
+the revision')
```
gctucker, line 96: Please break long lines between variables to avoid splitting strings:
```suggestion
    describe: Optional[str] = Field(
        default=None, description='git describe of the revision'
    )
```
Searching the code works better when strings aren't broken, and having code aligned to the left makes it harder to read.

JenySadadia: Done with the changes.

api/models.py
```diff
@@ -81,29 +83,63 @@ def update(self):
...
+    )
+    parent: Optional[PyObjectId] = Field(
+        description='Parent commit SHA'
+    )
+    status: Optional[StatusValues] = Field(
+        default=StatusValues.PENDING,
+        description='Status of test result'
+    )
+    artifacts: Optional[Dict] = Field(
+        description='Dictionary with names mapping to URLs'
```
gctucker, line 133: It's worth mentioning that the URLs are for artifacts associated with the node (binaries, logs...).

gctucker (commented):
Just one minor comment.

Looks OK on staging with auto-generated documentation: https://staging.kernelci.org:9000/docs#/

JenySadadia:
Another thing is documentation is not displaying `auth.TokenData` class in the schemas section.
Need to check on that one.

JenySadadia:
> > Another thing is documentation is not displaying `auth.TokenData` class in the schemas section. Need to check on that one.
> 
> I believe this is because the `TokenData` model isn't used by the API, it's only an internal model. There's no entry point using this model to exchange data with the client, as far as I can see. In fact we might even be able to drop `TokenData` if we don't make any other use of it since we're just using it temporarily to pass the `username` string in `get_current_user()`.

Okay. Nope, we can not drop `TokenData` as we will be using it to store scopes for user permissions.

gctucker (approved):

PR #141, opened 2022-05-11
T35295 Use OAuth2 scopes for user permissions

JenySadadia:
Use OAuth2 scopes to declare available scopes.
Get user scopes in `\token` endpoint and map it with JWT token.
Use the token scopes for authorization.


JenySadadia:
Token scopes can be added using request:

```
curl -X 'POST'   'http://localhost:8001/token'   -H 'accept: application/json'   -H 'Content-Type: application/x-www-form-urlencoded'   -d 'grant_type=&username=test&password=test&scope=user'
```

Multiple scopes can be provided using space-separated values.
```
curl -X 'POST'   'http://localhost:8001/token'   -H 'accept: application/json'   -H 'Content-Type: application/x-www-form-urlencoded'   -d 'grant_type=&username=test&password=test&scope=admin user'
```

gctucker:
Please rebase now that #139 has been merged.

JenySadadia:
> Please rebase now that #139 has been merged.

Done.

api/auth.py
```diff
@@ -94,10 +94,12 @@ async def get_current_user(self, token):
...
-                return None
+                return None, None
             token_data = TokenData(username=username)
         except JWTError:
-            return None
+            return None, None
 
-        return await self._db.find_one(User, username=token_data.username)
+        return await self._db.find_one(
+            User, username=token_data.username), token_scopes
```
gctucker, line 105: This would seem easier to read:
```python
user = await self._db.find_one(User, username=token_data.username)
return user, token_scopes
```

api/main.py
```diff
@@ -40,22 +40,37 @@ async def pubsub_startup():
...
-            headers={"WWW-Authenticate": "Bearer"},
+            headers={"WWW-Authenticate": authenticate_value},
         )
+    for scope in security_scopes.scopes:
+        if scope not in token_scopes:
+            raise HTTPException(
+                status_code=status.HTTP_401_UNAUTHORIZED,
+                detail="Access denied",
+                headers={"WWW-Authenticate": authenticate_value},
+            )
```
gctucker, line 73: Similarly, the `get_current_user()` method could be checking the scopes and return a `User` object only if there is a matching one that meets the scopes criteria rather than doing the check on the caller's side.

api/models.py
```diff
@@ -88,6 +88,10 @@ class User(ModelId):
         default=True,
         description='To check if user is active or not'
     )
+    isAdmin: bool = Field(
```
gctucker, line 91: I think the convention is to use underscores:
```python
is_admin: bool = Field(
```

api/main.py
```diff
@@ -106,6 +106,13 @@ async def login_for_access_token(
                 detail=f"Invalid scope: {scope}"
             )
 
+    if 'admin' in form_data.scopes:
```
gctucker, line 109: Maybe yet another thing that could be handled by `api.auth`.

gctucker (commented):
Looks great with a logical series of commits to add each part of this new feature.  I just put some comments about maybe moving some code to `api.auth` to simplify the logic in `api.main` and keep things self-contained around authentication handling.  But if it's not practical and causes some unnecessary complexity then we should be able to merge things as-is.

api/main.py
```diff
@@ -19,7 +19,12 @@
 
 app = FastAPI()
 db = Database()
-oauth2_scheme = OAuth2PasswordBearer(tokenUrl="token")
+
+user_scopes = {"admin": "Superusers", "users": "Regular users"}
+oauth2_scheme = OAuth2PasswordBearer(
+    tokenUrl="token",
+    scopes=user_scopes
+)
```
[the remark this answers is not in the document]
JenySadadia: Done.

api/auth.py
JenySadadia: Agreed.

api/main.py
JenySadadia: I think this should be added to a method in the User model.

api/main.py
```diff
@@ -79,7 +83,18 @@ async def login_for_access_token(
...
             headers={"WWW-Authenticate": "Bearer"},
         )
-    access_token = auth.create_access_token(data={"sub": user.username})
+
+    for scope in form_data.scopes:
+        if scope not in user_scopes:
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail=f"Invalid scope: {scope}"
+            )
```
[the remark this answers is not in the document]
JenySadadia: Added `auth.validate_scopes()`.

api/auth.py
```diff
@@ -6,14 +6,21 @@
...
 from .db import Database
 from .models import User
 
 
+user_scopes = {"admin": "Superusers", "users": "Regular users"}
+oauth2_scheme = OAuth2PasswordBearer(
+    tokenUrl="token",
+    scopes=user_scopes
+)
+
```
gctucker, line 22: The scopes could be part of the `Authentication` class. That would also probably solve the pylint error.

JenySadadia: Yes.

gctucker:
Please rebase one more time now that #142 has been merged.

api/main.py
```diff
@@ -19,7 +19,6 @@
 
 app = FastAPI()
 db = Database()
-oauth2_scheme = OAuth2PasswordBearer(tokenUrl="token")
 auth = Authentication(db)
```
gctucker, line 25: Maybe we could have:
```python
auth = Authentication(db)
oauth2_scheme = auth.get_scheme('token')
```
to avoid having `"token"` defined in `api.auth`.
Or:
```python
auth = Autnehtication(db, token_url='token')
```
and then `token: str = Depends(auth.scheme)` with internally scheme being `self._oauth2_scheme` and a `@property` to get it.

These are minor tweaks, but I think it's better to address issues related to implementation layers as soon as they appear to avoid creating some spaghetti code later on when the code base starts growing.

JenySadadia: Yes. Also added `user_scopes` to the constructor.

api/main.py
```diff
@@ -10,7 +10,10 @@
 
 from typing import List
 from fastapi import Depends, FastAPI, HTTPException, status, Request
-from fastapi.security import OAuth2PasswordRequestForm
+from fastapi.security import (
+    OAuth2PasswordRequestForm,
+    SecurityScopes
+)
```
gctucker, line 16: `SecurityScopes` is not used in `api.main` so the original imports can be left unchanged.

test/conftest.py
```diff
@@ -80,10 +80,10 @@ def mock_get_current_user(mocker):
     user = User(username='bob',
                 hashed_password='$2b$12$CpJZx5ooxM11bCFXT76/z.o6HWs2sPJy4iP8.'
                                 'xCZGmM8jWXUXJZ4K',
-                active=True)
+                active=True, is_admin=False)
```
gctucker, line 83: I think this commit should have been after the one to add `is_admin` or it'll always fail.  But that's a minor thing, the tests are passing at the end of the series of commits.

api/main.py
gctucker: OK let's leave it like this for now.

gctucker (commented):
A couple more minor tweaks

api/main.py
JenySadadia: It has been used by `main.get_current_user`.

```
async def get_current_user(
        security_scopes: SecurityScopes,
        token: str = Depends(auth.oauth2_scheme)):
```

test/conftest.py
JenySadadia: I have reordered the commits.

api/main.py
gctucker: Ah OK, I think that change is in a separate commit than the one where the import is being added.

JenySadadia: I have updated the commits. Thanks.

gctucker (approved):
Thanks, I've checked the pipeline run on staging and there's no side-effects so this is all working fine.

PR #140, opened 2022-05-10
T35275 api/main.py: implement /user endpoint

JenySadadia:
Depends on https://github.com/kernelci/kernelci-api/pull/141

Develop an endpoint to create user with a
POST request.

Signed-off-by: Jeny Sadadia <jeny.sadadia@collabora.com>

PR #145, opened 2022-05-17
T35399 kci_admin.py: Create utility script for admin operations

JenySadadia:
Depends on https://github.com/kernelci/kernelci-api/pull/140

Command-line utility for admin operations.

### DIFF
--- a/admin/_sample_settings.py
+++ b/admin/_sample_settings.py
@@ -1,0 +1,13 @@
+#!/usr/bin/env python3
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2022 Collabora Limited
+# Author: Jeny Sadadia <jeny.sadadia@collabora.com>
+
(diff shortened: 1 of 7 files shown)


JenySadadia:
> This is recreating some things that are already supported by kernelci-core, such as the settings file with what is essentially a `--db-config` setting. A better way to implement this would probably be to have methods in the `kernelci_api` module in kernelci-core to create / manage users, and `kci_data` in `kernelci-core` could have commands for that. This is something that will also need to be reworked as I think we should start separating the `kernelci_backend` and `kernelci_api` support in `kernelci-core` since ultimately we'll be dropping support for `kernelci_backend`. So in the meantime, a solution using `kci_data` may not be ideal but at least it's putting functionality at the correct layers in the stack I think.

Created PR: https://github.com/kernelci/kernelci-core/pull/1204

JenySadadia:
Created PR in kernelci-core. Hence, closing this one.

PR #146, opened 2022-05-26
T35622 API: Add documentation for admin user

JenySadadia:
Add instructions for adding users using `/user` endpoint.
Also, add description for getting token for an admin user.

### DIFF
--- a/doc/getting-started.md
+++ b/doc/getting-started.md
@@ -141,6 +141,9 @@ $ curl -X 'GET' \
 {"_id":"615f30020eb7c3c6616e5ac3","username":"bob","hashed_password":"$2b$12$VtfVij6zz20F/Qr0Ri18O.11.0LJMMXyJxAJAHQbKU0jC96eo2fr.","active":true}
 ```
 
+The token for an admin user should be created with admin scope so that other users can be created using kernelci-core command line tools or `/user` endpoint. More details are described [here](https://kernelci.org/docs/api/api-details/#create-an-API-token-with-security-scopes).
+
+
 ## Setting up a Pipeline instance
 
 The pipeline can perform a minimal set of tests using solely the API and its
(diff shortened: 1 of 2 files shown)


doc/getting-started.md
```diff
@@ -114,6 +114,17 @@ MongoDB server version: 5.0.3
 WriteResult({ "nInserted" : 1 })
 ```
 
+To create an admin user from a terminal, provide `is_admin: 1` to the request:
```
gctucker, line 117: I think we can simplify the documentation by saying that an admin user needs to be created first, and then additional users can be created with the kernelci-core command line tools.

Ideally we should have all the related changes in kernelci-core merged first, so either we adjust this PR now and wait for kernelci-core changes to be merged or we make a follow-up PR in the documentation later.

doc/getting-started.md
```diff
@@ -139,6 +150,46 @@ $ curl -X 'GET' \
 {"_id":"615f30020eb7c3c6616e5ac3","username":"bob","hashed_password":"$2b$12$VtfVij6zz20F/Qr0Ri18O.11.0LJMMXyJxAJAHQbKU0jC96eo2fr.","active":true}
 ```
 
+We can associate available security scopes with an API token. Currently available scopes are 'admin' (admin user permissions) and 'users' (regular user permissions).
```
gctucker, line 153: Same comment, the first token should be created with admin scope so that other users and anything else can then be done with the command line tools rather than the raw API.

We may detail parts of the raw API in the full API documentation (in fact that's already generated thanks to the OpenAPI format), but as this is a "getting started" guide it should be as simple as possible with just the minimal steps required to get a system up and running.

JenySadadia: I have added the detailed description to `api-details` and summary to `getting-started`.

doc/getting-started.md
JenySadadia: I think we need to merge this and https://github.com/kernelci/kernelci-core/pull/1210 together as both use URLs from each other's documentation.

gctucker (approved):

PR #144, opened 2022-05-16
T35393 Add unit test for /token endpoint for Admin

JenySadadia:
Depends on https://github.com/kernelci/kernelci-api/pull/143

Unit test for /token endpoint to get token with admin scope for Admin user.

PR #148, opened 2022-06-03
T35078: Trigger 'completed' event

JenySadadia:
Depends on https://github.com/kernelci/kernelci-api/pull/144

Trigger a 'completed' event when all child nodes are resolved (either with a pass, fail, or timeout status).


JenySadadia:
I think the return value of the endpoint will need to be modified once we use this in the pipeline.

api/main.py
```diff
@@ -186,6 +188,48 @@ async def get_root_node(node_id: str):
...
+                if ret:
+                    await db.update(node)
+                else:
+                    return {"message": "Nodes are not completed yet"}
+
+        operation = 'completed'
+        await pubsub.publish_cloudevent('node', {'op': operation,
+                                                 'id': str(node_id),
+                                        'nodes': jsonable_encoder(nodes)})
+        return {"message": "'Event triggered"}
```
mgalka, line 225: Is it possible to narrow down the block which is in the `try..except` block? I'm not entirely sure that we need to keep the whole function body in it.

api/main.py
```diff
@@ -186,6 +188,48 @@ async def get_root_node(node_id: str):
...
+async def trigger_completed_event(node_id: str, wait_time_hours: int = 0,
+                                  wait_time_minutes: int = 0,
+                                  wait_time_seconds: int = 0):
+    """Trigger an event when all child nodes are completed of
+       a given node"""
+    try:
+        nodes = await db.find_by_attributes(Node,
+                                            {"parent": ObjectId(node_id)})
+
+        if len(nodes) == 0:
```
mgalka, line 201: Wouldn't it be better to have here?:

```suggestion
        if not nodes:
```

api/models.py
```diff
@@ -145,6 +147,13 @@ class Node(DatabaseModel):
         default_factory=datetime.utcnow,
         description='Timestamp when node was last updated'
     )
+    max_wait_time: Optional[float] = Field(
+        default=24,
+        description='Maximum time in hours to wait for node to get it \
+completed',
+        ge=0,
+        le=24
```
mgalka, line 155: Shouldn't these be floats if the type hint says `Optional[float]`? E.g. `default=24.0` or `default=24.`

api/models.py
```diff
@@ -179,3 +188,19 @@ def translate_fields(cls, params: dict):
         if parent:
             translated['parent'] = ObjectId(parent)
         return translated
+
+    def set_timeout_status(self, current_time):
```
mgalka, line 192: Why do we need to pass `current_time` as a parameter? Wouldn't it be better to get time inside the method?

mgalka (commented):
Please, have a look at the comments/

api/models.py
JenySadadia: Yes, agreed.

api/models.py
JenySadadia: Because the value of `datetime.utcnow()` will be differed when the execution will reach this function. And we have already got the current time in `main` to pass it to `wait_for_node` function. I think the current time should be the same for both the functions.

api/main.py
JenySadadia: Yes, I think it is more readable.

api/main.py
JenySadadia: Yes, I narrowed it down.

api/models.py
mgalka: I'm trying to understand why we'd like to have time of some moment in the past rather than current time.  Could you help me with that?

Is there any reason why we need `wait_for_node` needs time taken from `main` instead of using current time? 
Why do we need `wait_for_node` and `set_timeout_status` use the same time?

mgalka (commented):
In general it looks good, but I'd like to understand why we want to pass time value from main to  check for timeout.

api/models.py
JenySadadia: One reason being is, that we can lower the execution time if we get the current time once, and use it in both functions. I also thought that if we get the current time in `set_timeout_status` it will have the sleep time of `wait_for_node` added to it. This will raise the chances of Node status being changed to `timeout`. But I think that is the correct thing to do logically. I will change the implementation to get the current time in both functions.

mgalka (approved):
LGTM

gctucker:
I think there are several design issues with this.  The events should either be generated by the API service itself, or managed by a dedicated client service. What I'm seeing here is a bit of both, and I think this is going to lead to some problems.

One thing that needs to be done either way is for clients to add either a timeout duration or an expiry timestamp to new node objects. I think something is done in this PR to cover that, but I don't quite understand the methods for them. I'll do a second review tomorrow.

Then about the 2 implementations, it's critical to have a single place in charge of transitioning nodes into their "timeout" status without splitting the logic between client and server.

Option 1 is to do this on the server (so in kernelci-api), in which case the clients shouldn't have to do anything special. A task could be run periodically to look for all nodes in some pending state and check if the current time is beyond their expiry time, and update their states to timeout if they are.

Option 2 is to do this on the client side (in kernelci-pipeline), whereby a client would request all the nodes in pending state and do the same logic, but then push the nodes with their state set to "timeout" when they have expired. I don't think we need any changes to the API at all to do that, except probably to be able to handle the "timed out" node state. But I fail to see why a new entry point or any logic would be needed in kernelci-api with this implementation. Also, doing this client-side comes with some limitations too: it's likely to be a bit slower to update each node that has timed out via the API but more importantly it could be a problem if multiple instances of the service were running. I think it's fine to start with, and we can move some features into kernelci-api later on as an optimisation.

JenySadadia:
I think Option 2 is preferable here. 

I will add one client script in the pipeline to set node status to 'timeout' after maximum wait time.
This will have `Node.max_wait_time` unchanged and will move the method `set_timeout_status` to pipeline.

PR #149, opened 2022-06-14
T35960 Fix authorization issue

JenySadadia:
Fix `CryptContext.verify` failing with the correct username and password. Need to use `get_secret_value()`
to get password in plain text while getting hash from password of type `SecretStr`.


mgalka (approved):
This looks good to me. It can be merged when #148 is merged.

PR #150, opened 2022-06-14
T35921 Command line utility to create an admin user

JenySadadia:
A python script, used to create a very first admin user by providing username, password, and optional API host URL from the command line.
After that a token can be retrieved to create other users using `/user` endpoint.

### DIFF
--- a/create_admin_user
+++ b/create_admin_user
@@ -1,0 +1,47 @@
+#!/usr/bin/env python3
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2022 Collabora Limited
+# Author: Jeny Sadadia <jeny.sadadia@collabora.com>
+
+"""Command line utility for creating an admin user"""
+
+import argparse
+import subprocess
+import json
+import urllib
+import re
+import requests
+
+
+DEFAULT_URL = 'http://localhost:8001'
+
+
+def get_hashed_password(password, url):
+    """Method to get hashed password"""
+    url = urllib.parse.urljoin(url, '/hash')
+    data = {'password': password}
+    res = requests.post(url, data=json.dumps(data))
+    return res.json()
+
+
+def create_admin(args):
+    """
+    Create an admin user
+    """
+    password = input("Please enter the password: ")
+    hashed_password = get_hashed_password(password, args.url)
+    subprocess.call(f"docker-compose exec db /bin/mongo kernelci --eval \
+\"db.user.insert({{username: '{args.username}', \
+hashed_password: '{re.escape(hashed_password)}', \
+active: true, is_admin: 1}})\"", shell=True)
+
+
+if __name__ == "__main__":
+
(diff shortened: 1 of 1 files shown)


alexandrasp (commented):
@JenySadadia it seems that python file is missing the `.py`

JenySadadia:
> @JenySadadia it seems that the python file is missing the `.py`

Hi @alexandrasp, here I have maintained the standards from kernelci-core. It has all `kci_*` python scripts just like this.

create_admin_user
```diff
@@ -0,0 +1,60 @@
...
+hashed_password: '{re.escape(hashed_password)}', \
+active: true, is_admin: 1}})\"")
+    print(ret)
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser("KernelCI API Admin")
+    parser.set_defaults(func=create_admin)
+    parser.add_argument('username', help="User name")
```
gctucker, line 53: This should be `required=True`

create_admin_user
```diff
@@ -0,0 +1,60 @@
...
+    parser.add_argument('password', help="User password")
```
gctucker, line 54: It would be better to interactively get this from the user to avoid passing it on the command line.

create_admin_user
```diff
@@ -0,0 +1,60 @@
...
+    parser.add_argument('--url', default=DEFAULT_URL, help="API host URL")
+    args = parser.parse_args()
+    try:
+        args.func(args)
+    except AttributeError as err:
+        parser.error("too few arguments")
```
gctucker, line 60: I don't think this is necessary, `ArgumentParser` will print an error if the arguments aren't correct.

Also, how about just calling `create_admin(args)` rather than storing the function in the args?

create_admin_user
JenySadadia: Yes.

create_admin_user
gctucker: Ah sorry, it's a keyword rather than an option.

create_admin_user
```diff
@@ -0,0 +1,56 @@
...
+    """
+    Create an admin user
+    """
+    password = input("Please enter the password: ")
+    hashed_password = get_hashed_password(password, args.url)
+    ret = shell_cmd(f"docker-compose exec db /bin/mongo kernelci --eval \
+\"db.user.insert({{username: '{args.username}', \
+hashed_password: '{re.escape(hashed_password)}', \
+active: true, is_admin: 1}})\"")
+    print(ret)
```
gctucker, line 47: This can simply be replaced with `subprocess.call("""<command>""")` as it will let the subprocess print to stdout. The `shell_cmd()` function isn't necessary here as the script only does one call, it's a helper copied from other tools in `kernelci-core` where it does simplify the syntax a bit.

JenySadadia: Yes, I used it from the core tools.
Updated the code.

gctucker (approved):
Tested OK locally.

PR #151, opened 2022-06-20
T35922 Regression tracking

JenySadadia:
Implement models and endpoints for regression tracking.

### DIFF
--- a/api/db.py
+++ b/api/db.py
@@ -8,7 +8,7 @@
 
 from bson import ObjectId
 from motor import motor_asyncio
-from .models import Node, User
+from .models import Node, User, Regression
 
 
 class Database:
@@ -22,6 +22,7 @@ class Database:
     COLLECTIONS = {
         User: 'user',
         Node: 'node',
+        Regression: 'regression',
     }
 
     def __init__(self, host='db', db_name='kernelci'):
--- a/api/main.py
+++ b/api/main.py
@@ -19,7 +19,7 @@ from fastapi.security import (
 from bson import ObjectId, errors
 from .auth import Authentication, Token
 from .db import Database
-from .models import Node, User, Password
+from .models import Node, Regression, User, Password
 from .pubsub import PubSub, Subscription
 
 app = FastAPI()
@@ -308,3 +308,23 @@ async def publish(raw: dict, channel: str, user: User = Depends(get_user)):
     attributes = dict(raw)
     data = attributes.pop('data')
     await pubsub.publish_cloudevent(channel, data, attributes)
+
+
+# -----------------------------------------------------------------------------
+# Regression
+
+@app.post('/regression')
+async def post_regression(regression: Regression,
+                          token: str = Depends(get_user)):
+    """Create a new regression"""
+    try:
+        obj = await db.create(regression)
+        operation = 'created'
+    except ValueError as error:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=str(error)
+        ) from error
+    await pubsub.publish_cloudevent('regression', {'op': operation,
+                                                   'id': str(obj.id)})
+    return obj
--- a/api/main.py
+++ b/api/main.py
@@ -9,7 +9,7 @@
 """KernelCI API main module"""
 
 from datetime import timedelta
-from typing import List
+from typing import List, Union
 from fastapi import Depends, FastAPI, HTTPException, status, Request, Security
 from fastapi.encoders import jsonable_encoder
 from fastapi.security import (
@@ -19,7 +19,7 @@ from fastapi.security import (
 from bson import ObjectId, errors
 from .auth import Authentication, Token
 from .db import Database
-from .models import Node, Regression, User, Password
+from .models import Node, Regression, User, Password, get_model_from_kind
 from .pubsub import PubSub, Subscription
 
 app = FastAPI()
@@ -147,10 +147,22 @@ def get_password_hash(password: Password):
 # -----------------------------------------------------------------------------
 # Nodes
 
-@app.get('/node/{node_id}', response_model=Node)
-async def get_node(node_id: str):
+@app.get('/node/{node_id}', response_model=Union[Regression, Node])
+async def get_node(node_id: str, kind: str = "node"):
     """Get node information from the provided node id"""
-    return await db.find_by_id(Node, node_id)
+    try:
+        model = get_model_from_kind(kind)
+        if model is None:
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail=f"Invalid kind: {kind}"
+            )
+        return await db.find_by_id(model, node_id)
+    except errors.InvalidId as error:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=str(error)
+        ) from error
 
 
 @app.get('/nodes', response_model=List[Node])
--- a/api/main.py
+++ b/api/main.py
@@ -347,3 +347,21 @@ async def post_regression(regression: Regression,
     await pubsub.publish_cloudevent('regression', {'op': operation,
                                                    'id': str(obj.id)})
     return obj
+
+
+@app.put('/regression/{regression_id}', response_model=Regression)
+async def put_regression(regression_id: str, regression: Regression,
+                         token: str = Depends(get_user)):
+    """Update an already added regression"""
+    try:
+        regression.id = ObjectId(regression_id)
+        obj = await db.update(regression)
+        operation = 'updated'
+    except ValueError as error:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=str(error)
+        ) from error
+    await pubsub.publish_cloudevent('regression', {'op': operation,
+                                                   'id': str(obj.id)})
+    return obj
--- a/api/main.py
+++ b/api/main.py
@@ -165,19 +165,26 @@ async def get_node(node_id: str, kind: str = "node"):
         ) from error
 
 
-@app.get('/nodes', response_model=List[Node])
-async def get_nodes(request: Request):
+@app.get('/nodes', response_model=List[Union[Regression, Node]])
+async def get_nodes(request: Request, kind: str = "node"):
     """Get all the nodes if no request parameters have passed.
        Get all the matching nodes otherwise."""
+    model = get_model_from_kind(kind)
+    if model is None:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=f"Invalid kind: {kind}"
+        )
+
     query_params = dict(request.query_params)
-    is_valid, msg = Node.validate_params(query_params)
+    is_valid, msg = model.validate_params(query_params)
     if not is_valid:
         raise HTTPException(
                 status_code=status.HTTP_400_BAD_REQUEST,
                 detail=f"Invalid request parameters: {msg}"
             )
-    translated_params = Node.translate_fields(query_params)
-    return await db.find_by_attributes(Node, translated_params)
+    translated_params = model.translate_fields(query_params)
+    return await db.find_by_attributes(model, translated_params)
 
 
 @app.get('/get_root_node/{node_id}', response_model=Node)
--- a/api/models.py
+++ b/api/models.py
@@ -260,3 +260,15 @@ class Regression(Node):
                 return False, str(error)
 
         return True, "Validated successfully"
+
+
+def get_model_from_kind(kind: str):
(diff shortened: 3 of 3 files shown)


gctucker:
Please remove the commit that adds the `create_admin_user` script as it's not related to this PR.

JenySadadia:
> Please remove the commit that adds the `create_admin_user` script as it's not related to this PR.

Done.

api/main.py
```diff
@@ -328,3 +328,15 @@ async def post_regression(regression: Regression,
     await pubsub.publish_cloudevent('regression', {'op': operation,
                                                    'id': str(obj.id)})
     return obj
+
+
+@app.get('/regression/{regression_id}', response_model=Regression)
+async def get_regression(regression_id: str):
```
gctucker, line 334: If Regressions are derived from Node, then maybe we could use the nodes API entry points to find them using `kind` set to `regression` and for example `parent_id` or other Node fields to find some particular ones.

api/models.py
```diff
@@ -206,3 +206,75 @@ async def wait_for_node(cls, timeout):
...
+
+    revision: Revision = Field(
+        description='Git revision of regression object'
+    )
+    status: StatusValues = Field(
+        description='Regression object status'
+    )
+
+
+class Regression(DatabaseModel):
```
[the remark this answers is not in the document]
JenySadadia: The reason I didn't inherit the `Node` model is, we don't need all the fields of it here e.g. `name`(as we already have node id), `max_wait_time`, `artifacts`, `updated`.

api/main.py
JenySadadia: I don't think we can use `Node` endpoints for `Regression` because `db.find_by_id` and `db.find_by_attributes` will take a model name as the parameter that will differ for `Node` and `Regression` models.

api/models.py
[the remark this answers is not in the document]
JenySadadia: Updated `Regression` model.

api/main.py
gctucker: At the moment we have:
```python
async def get_node(node_id: str):
    """Get node information from the provided node id"""
    return await db.find_by_id(Node, node_id)
```

We could have:
```
async def get_node(node_id: str, kind: str):
    """Get node information from the provided node id"""
    model = model_from_kind(kind)
    return await db.find_by_id(model, node_id)
```

api/main.py
```diff
@@ -340,3 +340,19 @@ async def get_regression(regression_id: str):
             status_code=status.HTTP_400_BAD_REQUEST,
             detail=str(error)
         ) from error
+
+
+@app.get('/regressions', response_model=List[Regression])
+async def get_regressions(request: Request):
```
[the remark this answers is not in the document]
JenySadadia: Updated the code for `get_node` and `get_nodes` endpoint.

api/main.py
```diff
@@ -308,3 +326,41 @@ async def publish(raw: dict, channel: str, user: User = Depends(get_user)):
     attributes = dict(raw)
     data = attributes.pop('data')
     await pubsub.publish_cloudevent(channel, data, attributes)
+
+
+# -----------------------------------------------------------------------------
+# Regression
+
+@app.post('/regression')
```
gctucker, line 334: Note: This means we can create or update regression objects with the `/regression` endpoint but only get them with the `/nodes` endpoint. I think this is fine and as far as I know there is no general expectation for an API to provide a `GET` for every endpoint. If we needed to implement it later we could easily do that by using the same logic as in the `/node` `GET` handler but with a hard-coded value for `kind`.

JenySadadia: Fair enough.

api/models.py
```diff
@@ -206,3 +206,43 @@ async def wait_for_node(cls, timeout):
...
+        description='Regression object status'
+    )
+
+
+class Regression(Node):
+    """API model for regression tracking"""
+    kind: str = Field(
+        default='regression',
+        description='Type of the object'
+    )
```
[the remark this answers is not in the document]
JenySadadia: Agreed.

api/models.py
```diff
@@ -206,3 +206,45 @@ async def wait_for_node(cls, timeout):
...
+class Regression(Node):
+    """API model for regression tracking"""
+
+    regression_data: List[RegressionData] = Field(
+        description='Regression details'
+    )
+
+    def __init__(self, **data):
+        if data.get('kind') is None:
+            data['kind'] = 'regression'
```
gctucker, line 235: Would that work?
```suggestion
    def __init__(self, kind='regression', **data):
```

api/models.py
```diff
@@ -206,3 +206,45 @@ async def wait_for_node(cls, timeout):
...
+    def validate_params(cls, params: dict):
+        """Validate regression parameters"""
+        ret, msg = Node.validate_params(params)
+        if not ret:
+            return ret, msg
+
+        status = params.get('regression_data.status')
+        if status and status not in [status.value
+                                     for status in StatusValues]:
+            return False, f"Invalid status value '{status}'"
```
gctucker, line 248: I think this is already covered by the Pydantic model, right? At least when data is being posted to the API endpoint, if there's an invalid field then it should return an HTTP error before it reaches this point.

api/db.py
```diff
@@ -68,6 +68,13 @@ async def find_by_attributes(self, model, attributes):
         data = await col.find(attributes).to_list(None)
         return list(model(**obj) for obj in data)
 
+    def get_model_from_kind(self, kind: str):
+        """Get model from kind parameter"""
+        for key, value in self.COLLECTIONS.items():
```
gctucker, line 73: Alternatively we could have a reversed dictionary created directly in the class, to avoid having to iterate through it every time. For example:
```python
    COLLECTIONS = {
        'foo': 'user',
        'bar': 'node',
        'baz': 'regression',
    }

    KINDS = {
        kind: collection
        for collection, kind in COLLECTIONS.items()
    }
```

Also, there's a slight problem here as the `COLLECTIONS` really are the name of the database collections. They may not always be the same as the models defined in `api.models`. So maybe this functionality should be implemented in `api.models` instead.

I think the whole model / collection / kind design needs to be reconsidered in the light of our current understanding of how this is meant to work in order to avoid potential issues or confusion in the semantics in the future.

api/main.py
```diff
@@ -147,10 +147,22 @@ def get_password_hash(password: Password):
 # -----------------------------------------------------------------------------
 # Nodes
 
-@app.get('/node/{node_id}', response_model=Node)
-async def get_node(node_id: str):
+@app.get('/node/{node_id}')
+async def get_node(node_id: str, kind: str = "node"):
     """Get node information from the provided node id"""
-    return await db.find_by_id(Node, node_id)
```
gctucker, line 153: It is a bit of a shame to be dropping `response_model=None` but I'm not sure what's best here.

One option would be to have some generic `data` field in the Node model so that other kinds of nodes could add things while still being a Node. That means the extra things would not be going through Pydantic validation, but it's more flexible in some ways.

Another option is to return a base `Node` object, but then we would be missing the extra fields from sub-class such as `Regression`.

I guess this is also something to take into account while reviewing how models / collections / kinds are supposed to be implemented and use by the API.

api/main.py
```diff
@@ -147,10 +147,22 @@ def get_password_hash(password: Password):
...
+    try:
+        model = db.get_model_from_kind(kind)
```
gctucker, line 153: If `kind` was a type with a finite number of possibilities (basically an `Enum`) then I think we could skip this validation as it would be done by Pydantic directly.

api/main.py
```diff
@@ -165,19 +164,26 @@ async def get_node(node_id: str, kind: str = "node"):
         ) from error
 
 
-@app.get('/nodes', response_model=List[Node])
```
gctucker, line 156: Maybe the response type can still be a `List` without specifying the type of its elements?  Also it might still be valid if all the elements are sub-classes of `Node` but I'm not sure how FastAPI would deal with that.

api/models.py
JenySadadia: If we want to use that signature the function becomes:
```
def __init__(self, kind='regression', **data):
        if data.get('kind') is None:
            data['kind'] = kind
        super().__init__(**data)
```
I don't see much difference. But I think this one is more readable and simplified.

api/models.py
JenySadadia: Yes, in the case of POST regression or node, Pydantic will automatically raise an error.
We have added this for GET nodes or regressions. In case the user passes invalid status or result attribute to the `/nodes` endpoint.

api/main.py
JenySadadia: If we use `Node`, it will not return fields only specific to `Regression` model.
We also can not use only `List` as then it fails to serialize response:
```
kernelci-api |   File "/home/kernelci/.local/lib/python3.10/site-packages/fastapi/encoders.py", line 144, in jsonable_encoder
kernelci-api |     raise ValueError(errors)
kernelci-api | ValueError: [TypeError("'ObjectId' object is not iterable"), TypeError('vars() argument must have __dict__ attribute')]
```

api/db.py
JenySadadia: Added `get_model_from_kind` to `api.models`.

api/main.py
JenySadadia: Yes, having `kind` as `Enum` seems fair.

api/main.py
JenySadadia: The better approach is to use `Union`.

api/models.py
gctucker: WIth `kind='regression'` you know 'kind' will not be in `**data` because that's the keyword arguments. So this should work then:
```python
def __init__(self, kind='regression', **data):
        super().__init__(kind=kind, **data)
```

api/main.py
gctucker: Well Python can handle that like any object-oriented language, if all the objects in the list are based on `Node`. For example:

```python
from typing import List


class A:
    name = 'something'


class B(A):
    name = "derived"
    foo = 'bar'


def main():
    l: List(A) = [A(), B()]
    print(l[0].name)
    print(l[1].name)
    print(l[1].foo)


if __name__ == '__main__':
    main()
```

But my question was whether Pydantic and FastAPI can handle that, by validating a `Regression` object and storing it in a list of `Node` object. I think it's worth double-checking.

JenySadadia: > > We also can not use only `List` as then it fails to serialize response
> 
> How did that work before when the returned type was a `List[Node]`?

I meant we can not use it without specifying any datatype i.e. `response_model=List[]`

JenySadadia: > Well Python can handle that like any object-oriented language, if all the objects in the list are based on `Node`. For example:
> 
> ```python
> from typing import List
> 
> 
> class A:
>     name = 'something'
> 
> 
> class B(A):
>     name = "derived"
>     foo = 'bar'
> 
> 
> def main():
>     l: List[A] = [A(), B()]
>     print(l[0].name)
>     print(l[1].name)
>     print(l[1].foo)
> 
> 
> if __name__ == '__main__':
>     main()
> ```
> 
> But my question was whether Pydantic and FastAPI can handle that, by validating a `Regression` object and storing it in a list of `Node` object. I think it's worth double-checking.

Yes, I did check. Here is the documentation URL from FastAPI: https://fastapi.tiangolo.com/tutorial/extra-models/#union-or-anyof 
The code with `response_model=List[Union[Regression,Node]]` is working as expected for Node and regression models.

api/models.py
JenySadadia: Yes.

api/main.py
JenySadadia: What if a user has stored a Node object with `kind=regression`? Shouldn't we put the validation?

api/main.py
gctucker: Erm, Union should be about data structures with variants. Here we have a plain object inheritance with just some added fields to the child class.

api/main.py
gctucker: Yes I think each model should have only one valid `kind` value. It's a way to "downcast" when going through a series of Node objects. In fact, maybe we shouldn't allow users to submit a value for `kind` and it should always be set with a fixed value directly in each model (like the `created` field I believe).

JenySadadia: Yes, the fixed value will resolve the issue.

api/models.py
gctucker: I thought the regression would contain a list of nodes, each of them having their own status already:
```python
regression_data: List[Node]
```
That way we can go through the history of results for a particular "test" (any kind of Node) through different revisions.

Did you have something else in mind to implement this?

JenySadadia: I had in mind to store a list of only regression specific information for a node. As we want a track of revision and status details when we record a new regression. Other details are stored in the Regression model.

gctucker (approved):

PR #152, opened 2022-06-22
T35078 api: drop Node.set_timeout_status, trigger_completed_event handler

JenySadadia:
Need to remove method and logic to set node status
to timeout in '/trigger_completed_event' endpoint
as a separate client script will be used
to do so in pipeline.

Fixes: 2e3c506 ("api.main: Add 'trigger_completed_event' endpoint")

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -8,10 +8,8 @@
 
 """KernelCI API main module"""
 
-from datetime import timedelta
 from typing import List, Union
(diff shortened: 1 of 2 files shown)


JenySadadia:
Removed 'trigger_completed_event' endpoint as we can use 'updated' event to mark node as completed and use it to generate test report.

gctucker (approved):
Tested OK on staging.

PR #153, opened 2022-06-24
doc: create admin user in getting-started section

gctucker:
Update the getting-started section to start by creating an admin user.
Then an extra step could be added as a follow-up to create regular
users via kci_data.

The tools to create API tokens and to create the initial user should
probably be improved first to simplify these initial steps before
adding more instructions (tool to just ask for admin password and
create entry + API token automatically etc.).

### DIFF
--- a/doc/api-details.md
+++ b/doc/api-details.md
@@ -27,8 +27,8 @@ If a user wants to change any of the above variables, they should be added to th
 
 ### Create an admin user
 
-The very first admin user needs to be created from terminal.
-To create an admin user from a terminal, provide `is_admin: 1` to the request:
+The very first admin user needs to be created from terminal.  To create an
+admin user from a terminal, provide `is_admin: 1` to the request:
 
 ```
 sudo docker-compose exec db /bin/mongo kernelci --eval   "db.user.insert({username: 'admin', hashed_password: '\$2b\$12\$VtfVij6zz20F/Qr0Ri18O.11.0LJMMXyJxAJAHQbKU0jC96eo2fr.', active: true, is_admin: 1})"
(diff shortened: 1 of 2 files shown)


doc/api-details.md
```diff
@@ -27,8 +27,8 @@ If a user wants to change any of the above variables, they should be added to th
 
 ### Create an admin user
```
JenySadadia, line 28: I think this whole section is already covered on the `getting-started` page https://github.com/gctucker/kernelci-api/blob/doc-admin-user/doc/getting-started.md#create-an-admin-user-account.
We can drop this here to avoid duplication.

doc/getting-started.md
```diff
@@ -92,46 +93,51 @@ curl \
...
 MongoDB shell version v5.0.3
 connecting to: mongodb://127.0.0.1:27017/kernelci?compressors=disabled&gssapiServiceName=mongodb
 Implicit session: session { "id" : UUID("84a62d1b-4b06-4631-8227-413964826100") }
 MongoDB server version: 5.0.3
 WriteResult({ "nInserted" : 1 })
 ```
 
-Even though after creating the above user, we would be able to run API, ideally, an admin user needs to be created first (using terminal), and then additional users can be created with the `/user` endpoint or kernelci-core command line tools (`kci_data`). The instructions are described [here](https://kernelci.org/docs/api/api-details/#users).
+> **Note** For more details about how to create users via the raw API, see the
+> [API documentation](/docs/api/api-details/#users)
```
JenySadadia, line 118: If we drop the `Create an admin user` section from the `api-details`, we need to add the below line at the end:

```Regular users can also be created this way without the `is_admin` attribute.```

JenySadadia (changes requested):

doc/api-details.md
gctucker: Well, maybe we'll need to update the API documentation. But I think everything about the API should be covered in the API documentation, whereas the "Getting Started" page provides step-by-step instructions to get a system working. So some things might be explained twice, with a focus on a particular use-case in the "Getting Started" section.

doc/getting-started.md
gctucker: OK, but I don't think we should drop the section from the API documentation.

JenySadadia: If we don't want to drop the whole section from `api-details`, we can remove the description of creating an admin user from it and add a link to the `getting-started` page. In case we want to add some details in the future, we can use the same section.

gctucker: OK, but this is about the detailed API documentation. This PR is to update the "getting started" section which needs to have instructions about how to create an initial admin user. So if the changes in this PR look fine we can have them merged and then rework the detailed API documentation as a follow-up.

JenySadadia: Yes, makes sense. Let me merge this.

JenySadadia (approved):

PR #155, opened 2022-07-06
T36372 Update Node model with status and result

JenySadadia:
Need to update Node model to accommodate status and result separately.
Update unit tests and documentation for the same.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -175,6 +175,11 @@ completed',
                                      for status in StatusValues]:
             return False, f"Invalid status value '{status}'"
 
+        result = params.get('result')
+        if result and result not in [result.value
+                                     for result in ResultValues]:
+            return False, f"Invalid result value '{result}'"
+
         parent = params.get('parent')
         if parent:
             try:
--- a/api/models.py
+++ b/api/models.py
@@ -36,10 +36,16 @@ class PyObjectId(ObjectId):
 class StatusValues(enum.Enum):
     """Enumeration to declare values to be used for Node.status"""
 
+    COMPLETED = "completed"
     PENDING = "pending"
+    TIMEOUT = "timeout"
+
+
+class ResultValues(enum.Enum):
+    """Enumeration to declare values to be used for Node.result"""
+
     PASS = "pass"
     FAIL = "fail"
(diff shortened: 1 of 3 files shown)


api/models.py
```diff
@@ -133,7 +139,10 @@ class Node(DatabaseModel):
     )
     status: Optional[StatusValues] = Field(
         default=StatusValues.PENDING,
-        description='Status of test result'
+        description='Status of test'
```
gctucker, line 142: It's actually not a test any more, but status of the node. This is a detail, we may not need to change it now but it'll be interesting to see how things evolve when we have actual test nodes and other kinds of nodes.

test/test_node_handler.py
```diff
@@ -367,6 +378,7 @@ def test_get_root_node_endpoint(mock_db_find_by_id, mock_init_sub_id):
             'name',
             'parent',
             'revision',
+            'result',
```
gctucker, line 381: I think it's a bit better to keep following the alphabetical order:
```suggestion
            'result',
            'revision',
```

gctucker (approved):
LGTM, just some minor things that might be tweaked.

gctucker:
In a follow-up PR it would be good to document the state machine for the `status` field with `pending`, `pass`, `fail` and `timeout`.  Also the transitions can be enforced in the model.

test/test_node_handler.py
JenySadadia: Updated.

api/models.py
JenySadadia: Updated the description for `status` and `result`.

JenySadadia:
> In a follow-up PR it would be good to document the state machine for the `status` field with `pending`, `pass`, `fail` and `timeout`. Also the transitions can be enforced in the model.

Can you please elaborate?

gctucker:
Sorry I mixed up status and result as they're now separate. I meant, the status can only start as `pending` and then it can only transition to `complete` or `timeout`. Then it can't go back to `pending` or change between `complete` and `timeout`. We may be able to add some code to validate these state changes and return an error if the client tries to make an "illegal" state change.

gctucker (approved):

PR #156, opened 2022-07-12
T36501 Declare Node.kind to be constant

JenySadadia:
Allow only default value for 'Node.kind' i.e 'node'.
This is required to get all the node objects based on the kind specified e.g regression nodes will have kind 'regression'.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -126,7 +126,8 @@ class Node(DatabaseModel):
     """KernelCI primitive node object model for generic test results"""
     kind: str = Field(
         default='node',
-        description='Type of the object'
+        description='Type of the object',
+        const=True
     )
     name: str = Field(
         description='Name of the node object'


gctucker:
How about also setting it to their fixed value in the model constructor?  Then I think we might not even need to actually store it in the database, but again I don't know how FastAPI and Pydantic would let us do that.

JenySadadia:
> How about also setting it to their fixed value in the model constructor? Then I think we might not even need to actually store it in the database, but again I don't know how FastAPI and Pydantic would let us do that.

If the field is `const` then the default value will always be set automatically. Then why do we want to specify it in the constructor? Also, if a user sends the value of Node.kind something other than the default, pydantic will automatically raise an error. I think it is better than handling that manually from the constructor.

gctucker:
OK, then subclasses will have to redefine the field with a different value. By going through the constructor, they could be passing a value to `super()` but I guess that may not fit well with the Pydantic model framework.

I'm not sure how const fields can be populated from user data anyway, I'll need to investigate that. I don't know if a const field can be initialised with a user value that matches the one in the field declaration or if the user just can't provide any value at all for a const field.

JenySadadia:
> OK, then subclasses will have to redefine the field with a different value. By going through the constructor, they could be passing a value to `super()` but I guess that may not fit well with the Pydantic model framework.
> 
We can not provide any other value than the default one(i.e. 'node' in Node) to `super()`. 

> I'm not sure how const fields can be populated from user data anyway, I'll need to investigate that. I don't know if a const field can be initialised with a user value that matches the one in the field declaration or if the user just can't provide any value at all for a const field.

Here, the user can provide the value for `kind` field but it should match with the default one otherwise the pydantic will raise an error.

JenySadadia:
@gctucker Does this look OK to you?

PR #157, opened 2022-07-13
doc/pipeline-details: add pipeline detailed flow

JenySadadia:
Add diagram and description of pipeline
flow in detail with each client script.


gctucker (approved):
OK I think this is fine as an initial step.  We can merge this now and then update it as we make further progress on the API and the pipeline gains more features.

PR #158, opened 2022-07-13
T36525 doc/getting-started: Add documentation for SSH

JenySadadia:
doc/getting-started: Add instruction to generate SSH keys to use SSH container. Also, provide instructions specific to WSL.
Dropped them from README.md


doc/getting-started.md
```diff
@@ -147,6 +147,62 @@ $ curl -X 'GET' \
 {"_id":"615f30020eb7c3c6616e5ac3","username":"bob","hashed_password":"$2b$12$VtfVij6zz20F/Qr0Ri18O.11.0LJMMXyJxAJAHQbKU0jC96eo2fr.","active":true}
 ```
 
+### Setup SSH keys
+
+We need to generate SSH keys to use SSH docker container.
```
gctucker, line 152: It's probably worth adding a line about the fact that this container is used for storage (even though it wasn't mentioned on the README, just some minor improvement we could make now) and that the files are then available for download via HTTP from the Nginx container.

JenySadadia: Updated.

mgalka (approved):

PR #159, opened 2022-07-19
T36483 Validate node transition

JenySadadia:
Validate values of `Node.status` and `Node.result` before updating the node object.


gctucker:
@JenySadadia I think it's time to rebase and update this one now that we have the new node states implemented.

JenySadadia:
> @JenySadadia I think it's time to rebase and update this one now that we have the new node states implemented.

Updated and tested OK locally.

gctucker (approved):
OK thanks.  I've removed the skip label, let's check it works as expected on staging.

mgalka (approved):

PR #160, opened 2022-07-19
T35078 Set root node status to 'completed'

JenySadadia:
Depends on https://github.com/kernelci/kernelci-api/pull/152

A few changes in the `trigger_completed_event` endpoint.
When all child nodes are completed, set root node status to 'completed' from the endpoint. Also, trigger the event.
Moved logic to wait for the node in `trigger_completed` pipeline script.


JenySadadia:
The endpoint is not needed anymore.
Hence, closing this one.

PR #161, opened 2022-07-21
doc: update getting started page with create_admin_user

gctucker:
Update the "Getting Started" page with a sample command using the new
create_admin_user script to create an initial admin user account
instead of raw HTTP queries.

### DIFF
--- a/doc/getting-started.md
+++ b/doc/getting-started.md
@@ -78,38 +78,20 @@ which enables them to create new user accounts using the
 [kci_data](/docs/core/kci_data#creating-new-api-user-experimental) command line
 tool.
 
-So let's start by creating the initial admin user account.  First, get an
-encrypted hash for your password with the `/hash` API endpoint.  This uses a
-`POST` method for security reasons.  It doesn't require any authentication and
-it doesn't make any changes to the database.  For example, if the password is
-`hello`:
+So let's start by creating the initial admin user account.  This can be done
+with the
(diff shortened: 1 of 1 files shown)


doc/getting-started.md
```diff
@@ -78,38 +78,18 @@ which enables them to create new user accounts using the
 [kci_data](/docs/core/kci_data#creating-new-api-user-experimental) command line
 tool.
 
-So let's start by creating the initial admin user account.  First, get an
-encrypted hash for your password with the `/hash` API endpoint.  This uses a
-`POST` method for security reasons.  It doesn't require any authentication and
-it doesn't make any changes to the database.  For example, if the password is
-`hello`:
+So let's start by creating the initial admin user account.  This can be done
+with the `create_admin_user` tool provided in the `kernelci-api` repository.
```
JenySadadia, line 82:
```suggestion
with the [`create_admin_user`](https://github.com/kernelci/kernelci-api/blob/main/create_admin_user) tool provided in the `kernelci-api` repository.
```

JenySadadia (changes requested):
A link to the script can be added here.

doc/getting-started.md
gctucker: The `kernelci-api` repository needs to be checked out locally anyway, but yes I guess we could have a link.  At some point the script should be installed in the standard path when things start shaping up a bit more and the tools have been consolidated between core, pipeline and api.

JenySadadia (approved):

PR #163, opened 2022-08-01
T36848 api.models: rename Node status to 'complete'

JenySadadia:
Node.status value 'complete' is more appropriate
than 'completed' to denote completed node.


gctucker (approved):
OK this change is trivial so let's merge it now and update the pipeline accordingly.

We can consider renaming other status names later (e.g. "timeout" to "expired") once all the pending pipeline changes have been merged.

PR #164, opened 2022-08-05
T36916 Implement pagination for '/nodes' endpoint

JenySadadia:
Add 'skip' and 'limit' request query parameters to get a desired set of nodes in response from the endpoint.
Need to drop them from 'query_params' dictionary to match the fields with the model. It is required to get objects matching model attributes from the database.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -234,6 +234,14 @@ completed',
             translated['parent'] = ObjectId(parent)
         return translated
 
+    @classmethod
+    def filter_params(cls, params: dict):
+        """Drop parameters not matching model fields from the `params`
+        dictionary"""
+        for param in list(params):
+            if param not in cls.__fields__:
+                del params[param]
+
 
 class Regression(Node):
     """API model for regression tracking"""
--- a/api/main.py
+++ b/api/main.py
@@ -8,17 +8,33 @@
 
 """KernelCI API main module"""
 
-from typing import List, Union
-from fastapi import Depends, FastAPI, HTTPException, status, Request, Security
+from typing import Union
+from fastapi import (
+    Depends,
+    FastAPI,
+    HTTPException,
+    status,
+    Request,
+    Security,
+    Query
+)
 from fastapi.security import (
     OAuth2PasswordRequestForm,
     SecurityScopes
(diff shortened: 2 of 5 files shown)


gctucker (commented):
How about reusing this?
https://pypi.org/project/fastapi-pagination/

JenySadadia:
> How about reusing this? https://pypi.org/project/fastapi-pagination/

The reason being is FastAPI official documentation doesn't mention this package for pagination. 
Instead, it mentioned using query parameters to implement it the simpler way. https://fastapi.tiangolo.com/tutorial/query-params/
Also, I am unsure if the https://pypi.org/project/fastapi-pagination/ is thoroughly tested.

gctucker:
> 

Looking at the GitHub page it runs over 500 unit tests on 4 Python revisions and also does pylint.  It's actively maintained so I think it's worth considering it.  Of course the FastAPI documentation doesn't mention external packages, but that doesn't mean they can't be used.

JenySadadia:
> Looking at the GitHub page it runs over 500 unit tests on 4 Python revisions and also does pylint. It's actively maintained so I think it's worth considering it. Of course the FastAPI documentation doesn't mention external packages, but that doesn't mean they can't be used.

Okay. I think I'll give it a try.

api/models.py
```diff
@@ -66,6 +67,18 @@ class Config:
...
+class PageModel(LimitOffsetPage[TypeVar("T")]):
+    """Model for pagination
+
+    This model is required to serialize paginated model data response"""
+
+    class Config:
+        """Configuration attributes for PageNode"""
+        json_encoders = {
+            ObjectId: str,
+        }
```
mgalka, line 79: I'd think of moving this model to different module. `models.py` seems to be a place for data models. These are connected with the application logic, while `PageModel` is rather connected with data presentation.

mgalka (commented):
Small remarks

api/models.py
JenySadadia: Makes sense. Added `paginator_models.py`

mgalka (approved):
LGTM

PR #165, opened 2022-08-05
T35078 api.models: remove Node.wait_for_node method

JenySadadia:
Need to drop the method to wait for node
to get completed as it is not required
anymore as per the new design changes.


gctucker (approved):
Tested OK on staging.  This method is not used any more.

PR #166, opened 2022-08-15
api.main: fix user authentication

JenySadadia:
Use `main.get_password_hash` instead of
`auth.get_password_hash` directly to convert
password from SecretStr to plain text first.

Fixes: 7a67489 ("api/main.py: implement /user endpoint")

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -75,7 +75,8 @@ async def post_user(
         current_user: User = Security(get_user, scopes=["admin"])):
     """Create new user"""
     try:
-        hashed_password = auth.get_password_hash(str(password.password))
+        hashed_password = auth.get_password_hash(
+                                password.password.get_secret_value())
         obj = await db.create(User(
                                 username=username,
                                 hashed_password=hashed_password,


mgalka:
@JenySadadia Do you have this change tested?

JenySadadia:
> @JenySadadia Do you have this change tested?

Yes, tested OK on local.

api/main.py
```diff
@@ -75,7 +75,7 @@ async def post_user(
         current_user: User = Security(get_user, scopes=["admin"])):
     """Create new user"""
     try:
-        hashed_password = auth.get_password_hash(str(password.password))
+        hashed_password = get_password_hash(password)
```
gctucker, line 78: The issue is that `get_password_hash()` is an API entry point function, so I don't think it's a good idea to call it from within another entry point function.

This should work too and would stay within the current layer:
```python
hashed_password = auth.get_password_hash(password.password.get_secret_value())
```

JenySadadia: Done.

PR #167, opened 2022-08-19
api.models: add Node.group attribute to group nodes together

gctucker:
Add a Node.group attribute to be able to query nodes that belong to a
same group.  For example, this can be used to get all the nodes with
test cases from a single test suite.  The group name is optional and
should be set to the name of the top-level name of the group.


api/models.py
```diff
@@ -162,6 +162,9 @@ class Node(DatabaseModel):
     name: str = Field(
         description='Name of the node object'
     )
+    group: Optional[str] = Field(
+        descriptoin='Name of a group this node belongs too'
```
JenySadadia, line 166:
```suggestion
        descriptoin='Name of a group this node belongs to'
```

api/models.py
```diff
@@ -162,6 +162,9 @@ class Node(DatabaseModel):
     name: str = Field(
         description='Name of the node object'
     )
+    group: Optional[str] = Field(
+        descriptoin='Name of a group this node belongs to'
```
JenySadadia, line 166: Need to fix a minor typo.
```suggestion
        description='Name of a group this node belongs to'
```

JenySadadia (approved):

PR #168, opened 2022-08-19
api.models: add Node.path with list of node names to root

gctucker:
Add Node.path attribute to provide a list of strings with the name of
all the parent nodes from the root.  This can be used to name a test
case with dotted syntax, for example kunit.overflow.test_u8_zero.

Depends on #167


api/models.py
```diff
@@ -162,6 +162,12 @@ class Node(DatabaseModel):
     name: str = Field(
         description='Name of the node object'
     )
+    path: List[str] = Field(
+        description='Full path with node names from the top-level node'
+    )
+    group: Optional[str] = Field(
```
JenySadadia, line 168: I see the `group` field is `optional` here.
Shouldn't every test belong to one of the test groups?

gctucker: No, the checkout and tarball nodes aren't part of any group.

test/test_node_handler.py
```diff
@@ -358,6 +373,7 @@ def test_get_root_node_endpoint(mock_db_find_by_id, mock_init_sub_id):
             _id="61bda8f2eb1a63d2b7152410",
             kind="node",
             name="root",
+            path=["parent", "something", "root"],
```
JenySadadia, line 376: `checkout` node is itself the root node. Hence the `node_obj` and `root_node_obj` should be the same.
Or else we can change the `node_obj` to be different and `root_node_obj` will be the `checkout` node.

gctucker:
@JenySadadia Hello, does this look OK to you now?

JenySadadia:
> @JenySadadia Hello, does this look OK to you now?

Yes, looks good to me.

JenySadadia (approved):

PR #170, opened 2022-08-23
Add Node state diagram

gctucker:
Add a state diagram for Node objects.  Also clean up and improve some parts of the api-details page.

### DIFF
--- a/doc/api-details.md
+++ b/doc/api-details.md
@@ -1,11 +1,12 @@
 ---
 title: "API details"
-date: 2022-05-09
+date: 2022-08-23
 description: "KernelCI API building blocks"
 weight: 3
 ---
 
-This guide describes KernelCI API building blocks such as Node model and PubSub interface. It also includes details about environment variables.
+This guide describes KernelCI API building blocks such as Node model and PubSub
+interface. It also includes details about environment variables.
 
 
 ## Environment Variables
@@ -14,11 +15,12 @@ The instructions about environment file is described [here].(https://kernelci.or
 
 ### Set ALGORITHM and ACCESS_TOKEN_EXPIRE_MINUTES in environment file
 
-We need to specify an algorithm for JWT token encoding and decoding. ALGORITHM variable needs to be passed in the parameter for that.
-ALGORITHM is set default to HS256.
-We have used ACCESS_TOKEN_EXPIRE_MINUTES variable to set expiry time on generated jwt access token.
-ACCESS_TOKEN_EXPIRE_MINUTES is set default to None.
-If a user wants to change any of the above variables, they should be added to the .env file.
+We need to specify an algorithm for JWT token encoding and decoding. ALGORITHM
+variable needs to be passed in the parameter for that.  ALGORITHM is set
+default to HS256.  We have used ACCESS_TOKEN_EXPIRE_MINUTES variable to set
+expiry time on generated jwt access token.  ACCESS_TOKEN_EXPIRE_MINUTES is set
+default to None.  If a user wants to change any of the above variables, they
+should be added to the .env file.
 
 
 ## Users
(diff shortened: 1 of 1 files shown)


doc/api-details.md
```diff
@@ -164,6 +176,61 @@ $ curl -X 'PUT' \
...
+Ready
+: The job was successful.  Child nodes that depend on the job's success can now
+  be created.
+
+Pending
+: The holdoff time has been reached, the node is now waiting for any pending
+  child nodes to reach the Done state.
+
+Done
+: The node has reached its final state.
```
JenySadadia, line 210: I would suggest using Bullet List here.

gctucker: This is actually the Markdown syntax for definitions

JenySadadia: Oh, okay. I was unaware of that.

doc/api-details.md
```diff
@@ -164,6 +176,61 @@ $ curl -X 'PUT' \
...
+different state values are described below:
+
+Running
+: The job has been scheduled and there is no result yet.
+
+Ready
+: The job was successful.  Child nodes that depend on the job's success can now
+  be created.
+
+Pending
```
JenySadadia, line 205: How about renaming it to "awaiting" as we are waiting for the child nodes to get completed?
```suggestion
Awaiting
```

doc/api-details.md
```diff
@@ -164,6 +176,61 @@ $ curl -X 'PUT' \
...
+> be used (for example, runtime tests can't be scheduled if a kernel build
+> failed).
+
+There are two fields that can cause time-driven state transitions:
+
+holdoff
+: This is for the node to remain in the Ready state for a minimum amount of
+  time and allow other nodes that depend on its success to be created.  Without
+  this time constraint, there would be a race condition when the job is
+  complete as without any child nodes it would go directly to Done
```
JenySadadia, line 227:
```suggestion
  complete as without any child nodes it would go directly to Done.
```

doc/api-details.md
gctucker: I'm not sure, because it could be "awaiting job completion" or anything.  It doesn't really mean it's awaiting child nodes to complete.

I think it's better if the names are about the node state seen from the node's "point of view".  So "running" and "done" are fine, but "ready" means it's waiting for any potential child nodes to get scheduled.  The node itself is not really more ready when it's in this state than when it's "running".  The "pending" or "awaiting" state has the same issue I think.

So here are the definitions of these states, let's call them 2 and 3 (1 being running and 4 being done):

2: The job has succeeded, any other job that depend on it can now get scheduled (e.g. tests depending on a source tarball).  The "holdoff" time has not been reached yet, so the node won't move to the Done state even if all child nodes are already Done.

3: The "holdoff" time has been reached, so as soon as all child nodes are Done the node will also be moved to the Done state.

gctucker: I guess state 2 could be called Available (instead of Ready) even though that's arguably not really from the node's point of view, it's available for other jobs to use it.

For state 3, maybe Closing would be a better fit.  Like, it's time to get out of a shop or something before it's too late (timeout).  And that's really about the parent node, it's closing so child nodes have to complete in time.

A more poetic way to call these states could be Dawn, Day, Dusk and Night.  I think we could use that for a presentation slide to illustrate the states rather than actually call them like that in the API as it's probably a bit too vague.

JenySadadia: When I think from the node's point of view this makes more sense to me:
```
Running --> Available --> Closing/Completing/Terminating/Ending --> Done
```

2. `Available` is preferable here as it denotes that a node is available for creating child nodes.
3. Here we are trying to say is the node is about to complete or end as the hold off has been reached. I would rather go for one of `Closing/Completing/Terminating/Ending` states to denote it. The `pending` will not convey the actual point.

gctucker: OK thanks.  So I think we agree on Running, Available, Closing, Done

JenySadadia (approved):

gctucker:
Waiting for this to be deployed on the staging static website to confirm it's looking as expected.

gctucker:
OK: https://static.staging.kernelci.org/docs/api/api-details/#state-diagram

PR #171, opened 2022-08-23
api.models: implement new Node.state values

gctucker:
Replace Node.status with Node.state to make it clear it's about the
state of the Node and not about any pass/fail result.  Update the list
of states to match the new specs.

### DIFF
--- a/test/test_node_handler.py
+++ b/test/test_node_handler.py
@@ -44,7 +44,7 @@ def test_create_node_endpoint(mock_get_current_user, mock_init_sub_id,
             group="debug",
             revision=revision_obj,
             parent=None,
-            status="pending",
+            state="closing",
             result=None,
         )
     mock_db_create.return_value = node_obj
@@ -82,7 +82,7 @@ def test_create_node_endpoint(mock_get_current_user, mock_init_sub_id,
             'parent',
             'result',
             'revision',
-            'status',
+            'state',
             'updated',
         }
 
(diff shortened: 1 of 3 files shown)


gctucker:
Adding `staging-skip` until the pipeline services have been updated to use the new state names.  Also existing nodes in the database will have to be removed (or migrated...).

test/test_node_handler.py
```diff
@@ -351,7 +351,7 @@ def test_get_root_node_endpoint(mock_db_find_by_id, mock_init_sub_id):
             name="checkout",
             revision=revision_obj,
             parent="61bda8f2eb1a63d2b7152410",
-            status="pending",
+            state="pending",
```
JenySadadia, line 354: Although the state value doesn't make much of a difference in these tests, the node state value should be changed to the default one.
```suggestion
            state="running",
```
This should be applied to all the nodes in tests.

gctucker: Why does it need to be the default value?

JenySadadia: I think, If we are setting the random value to node state for tests, the default one should be preferred.
I don't see why we should set it to 'pending'.

gctucker: Well I think we should be testing all the values in principle, and I don't see why the default one is preferred to other ones.  The only reason we might test the default one is if we don't set a value we can check the default one is used when the node is created.

JenySadadia (approved):

test/test_node_handler.py
JenySadadia: Makes sense.

PR #173, opened 2022-08-23
pipeline-details: add new node fields and client script changes

JenySadadia:
Add new node fields as per the latest schema to
flow diagram. Also, modify diagram and description
as per the new changes to client scripts.


doc/pipeline-details.md
```diff
@@ -15,35 +15,45 @@ flowchart
...
-        test_report --> email_report[Generate and <br />email test report]
+        test_report[Test Report] --> email_report[Generate and <br />email test report <br />for tarball]
     end
     subgraph set_timeout_service[Set Timeout Service]
-        set_timeout[Set Timeout] --> max_wait_time{Maximum wait <br/>time over?} --> |Yes| set_status[Set node status <br />to 'timeout']
-        max_wait_time --> |No| set_timeout[Set Timeout]
-        set_status --> |event: <br />updated| trigger_completed
+        set_timeout[Set Timeout] --> get_incomplete_nodes
+        get_incomplete_nodes[Get nodes <br /> with status=pending] --> max_wait_time
+        max_wait_time{Maximum wait <br/>time 'node.timeout' over?} --> |Yes| set_incomplete[Set parent and pending child nodes <br /> status=done, result=incomplete]
```
gctucker, line 50: I think `max_wait_time` is now just `timeout`, right?

JenySadadia: > I think `max_wait_time` is now just `timeout`, right?

Yes.

gctucker:
The diagram can be seen here: https://static.staging.kernelci.org/docs/api/pipeline-details/

I think it's a bit too much on one diagram, not sure how to make it easier to read though. Maybe the holdoff and timeout parts could have their own diagram.

There seems to be an issue with the flowchart as there's an arrow between the "completed" and "timeout" services, but the only way they can communicate is via pub/sub events which implicitly happens when a node changes state.  So there should be a node state change caused by one service (from Ready to Pending), then the timeout service would move nodes from Pending to Done.

JenySadadia:
> The diagram can be seen here: https://static.staging.kernelci.org/docs/api/pipeline-details/
> 
> I think it's a bit too much on one diagram, not sure how to make it easier to read though. Maybe the holdoff and timeout parts could have their own diagram.
> 
> There seems to be an issue with the flowchart as there's an arrow between the "completed" and "timeout" services, but the only way they can communicate is via pub/sub events which implicitly happens when a node changes state. So there should be a node state change caused by one service (from Ready to Pending), then the timeout service would move nodes from Pending to Done.

I notice a difference in the diagram at https://github.com/JenySadadia/kernelci-api/blob/document-schema-changes/doc/pipeline-details.md and https://static.staging.kernelci.org/docs/api/pipeline-details/.
The conditional block 'All child nodes completed ?' of set_completed service is in set_timeout flow instead.

About the service communication issue, I will update the diagram to set the node state to 'pending' if holdoff has been reached and child nodes are still pending.

gctucker (approved):
OK that's great, I think we can merge it now and tweak a few things as follow-ups.

The checkout and tarball nodes are getting merged together with just checkout nodes, so that will need to be reflected in the diagram.

Then I think it might be a good idea to split it into smaller diagrams, but that's just some suggestion for improving readability.

PR #174, opened 2022-09-02
Add /nodes/{node_id} entry point for submitting hierarchy

gctucker:
Add a Hierarchy model and an entry point for submitting a hierarchy of nodes to be added with an existing root node as the top-level parent.  Implement a `Database.create_hierarchy()` method to create all the node objects in the database accordingly.


gctucker (commented):
Tested OK on staging, merging now to unblock progress.

PR #175, opened 2022-09-06
api.paginator_models: add model to redefine limit

JenySadadia:
Add 'CustomLimitOffsetParams' class to redefine
limit parameter with custom validations.

### DIFF
--- a/api/paginator_models.py
+++ b/api/paginator_models.py
@@ -6,14 +6,26 @@
 """Model definitions for pagination"""
 
 from typing import TypeVar
-from fastapi_pagination import LimitOffsetPage
+from fastapi import Query
+from fastapi_pagination import LimitOffsetPage, LimitOffsetParams
 from bson import ObjectId
+
+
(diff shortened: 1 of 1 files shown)


gctucker:
Is it really worth the additional code though?  Maybe a limit of 100 is just fine.  The real issue is that it adds complexity on the caller's side to always handle paging when in some cases a full list of nodes is required, and for this we can implement something in `kernelci-core` to iterate over the pages and return a flat list.

gctucker:
@JenySadadia Why add `staging-skip` label?

JenySadadia:
> Is it actually possible to not set a limit? I thought that dropping the maximum means the client could set a very high limit, but if there's no limit from the client then the default would be used instead.

Your understanding is correct here. I have dropped the validation on the highest limit, so the client would be able to set any limit. Otherwise, the default will be used.

gctucker (approved):

Issue #176, opened 2022-09-06
Fix pagination integration with Mongo DB


gctucker:
The pagination for returning Node objects is currently returning a slice of the whole data set retrieved from Mongo DB.  Instead, the limit and offset arguments should be used in the Mongo DB query to only retrieve the data that's needed by the client and return it directly.

See the Motor integration for FastAPI Pagination:

https://github.com/uriyyo/fastapi-pagination/blob/main/examples/pagination_motor.py

PR #182, opened 2022-09-12
T37711 Fix pagination integration with Mongo DB

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/176

api.db: use pagination in `Database.find_by_attributes`
api.main: Drop the call to `paginate` method from `get_nodes`

### DIFF
--- a/api/db.py
+++ b/api/db.py
@@ -7,6 +7,7 @@
 """Database abstraction"""
 
 from bson import ObjectId
+from fastapi_pagination.ext.motor import paginate
 from motor import motor_asyncio
 from .models import Hierarchy, Node, User, Regression
 
@@ -68,11 +69,13 @@ class Database:
         """Find objects with matching attributes
 
         Find all objects with attributes matching the key/value pairs in the
-        attributes dictionary and return them as a list.
+        attributes dictionary using pagination and return the paginated
+        response.
+        The response dictionary will include 'items', 'total', 'limit',
+        and 'offset' keys.
         """
         col = self._get_collection(model)
-        data = await col.find(attributes).to_list(None)
-        return list(model(**obj) for obj in data)
+        return await paginate(collection=col, query_filter=attributes)
 
     async def create(self, obj):
         """Create a database document from a model object
--- a/api/main.py
+++ b/api/main.py
@@ -16,7 +16,6 @@ from fastapi import (
     status,
     Request,
     Security,
-    Query
 )
 from fastapi.security import (
     OAuth2PasswordRequestForm,
@@ -194,9 +193,7 @@ async def get_node(node_id: str, kind: str = "node"):
 
 
 @app.get('/nodes', response_model=PageModel)
-async def get_nodes(request: Request, kind: str = "node",
-                    offset: int = Query(default=0, ge=0),
-                    limit: int = Query(default=50, ge=1)):
+async def get_nodes(request: Request, kind: str = "node"):
     """Get all the nodes if no request parameters have passed.
        Get all the matching nodes otherwise, within the pagination limit."""
     model = get_model_from_kind(kind)
--- a/api/main.py
+++ b/api/main.py
@@ -22,7 +22,7 @@ from fastapi.security import (
     OAuth2PasswordRequestForm,
     SecurityScopes
 )
-from fastapi_pagination import add_pagination, paginate
+from fastapi_pagination import add_pagination
 from bson import ObjectId, errors
 from pymongo.errors import DuplicateKeyError
 from .auth import Authentication, Token
@@ -220,7 +220,7 @@ async def get_nodes(request: Request, kind: str = "node",
         )
 
     translated_params = model.translate_fields(query_params)
-    return paginate(await db.find_by_attributes(model, translated_params))
+    return await db.find_by_attributes(model, translated_params)
 
 add_pagination(app)
 
--- a/test/test_node_handler.py
+++ b/test/test_node_handler.py
@@ -136,7 +136,12 @@ def test_get_nodes_by_attributes_endpoint(mock_get_current_user,
             state="closing",
             result=None,
         )
-    mock_db_find_by_attributes.return_value = [node_obj_1, node_obj_2]
+    mock_db_find_by_attributes.return_value = {
+        'items': [node_obj_1, node_obj_2],
+        'total': 2,
+        'limit': 50,
+        'offset': 0
+    }
 
     params = {
         "name": "checkout",
@@ -152,7 +157,7 @@ def test_get_nodes_by_attributes_endpoint(mock_get_current_user,
             )
         print("response.json()", response.json())
         assert response.status_code == 200
-        assert len(response.json()) > 0
+        assert len(response.json()['items']) > 0
 
 
 def test_get_nodes_by_attributes_endpoint_node_not_found(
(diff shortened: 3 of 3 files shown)


api/db.py
```diff
@@ -68,11 +69,13 @@ async def find_by_attributes(self, model, attributes):
         """Find objects with matching attributes
 
         Find all objects with attributes matching the key/value pairs in the
-        attributes dictionary and return them as a list.
+        attributes dictionary using pagination and return the paginated
+        response.
+        The response dictionary will include 'items', 'total', 'limit',
+        and 'offset' keys.
         """
-        col = self._get_collection(model)
```
gctucker, line 73: Keep this on a separate line probably still helps with readability rather than in the `paginate()` call.

api/db.py
```diff
@@ -68,11 +69,13 @@ async def find_by_attributes(self, model, attributes):
...
-        data = await col.find(attributes).to_list(None)
-        return list(model(**obj) for obj in data)
+        return await paginate(collection=self._get_collection(model),
+                              query_filter=attributes)
```
gctucker, line 78: Looks good. It would be interesting to get some stats from Mongo DB about the number of objects looked up or the time it took to complete the query, or to trace the actual operations it's doing.  Overall I guess it should already speed up how the email report is reading all the nodes currently.

api/main.py
```diff
@@ -220,7 +220,7 @@ async def get_nodes(request: Request, kind: str = "node",
         )
 
     translated_params = model.translate_fields(query_params)
-    return paginate(await db.find_by_attributes(model, translated_params))
+    return await db.find_by_attributes(model, translated_params)
```
gctucker, line 223: The `limit` and `offset` arguments are dropped from `query_params`, so they won't be in `translated_params` either.  So I believe they're not passed to `find_by_attributes()` and pagination is still not happening at the database level.

Are you actually returning chunks of the expected length?  I wouldn't be surprised if the returned data was the whole set with no pagination enabled.

JenySadadia: Even if we do not provide values of `limit` and `offset` to the method,  `paginate` is able to get those values and uses it to only get the required chunk from the database. So, pagination is working at the db level.

The `paginate` method is fetching the variables using context variable.
Below is the piece of code:
```
@dataclass
class RawParams:
    limit: int
    offset: int


class AbstractParams(ABC):
    @abstractmethod
    def to_raw_params(self) -> RawParams:
        pass
```

I also tried to print the output of `await paginate(collection=col, query_filter=attributes)`. The output has the actual value of `limit` and `offset` passed to the endpoint.

gctucker: So how are the limit and offset values passed to the paginate function if they're not in the parameters?

gctucker: I guess this is some implicit FastAPI feature to provide the query parameters while bypassing the endpoint handler?  It would be good to understand this completely.  Maybe you've already found the documentation for this?

JenySadadia: I haven't found any related documentation. I studied the `fastapi-pagination` code. As per my understanding below is the scenario here:

In `add_pagination` function, there is `_update_route`. In which query parameters are determined using `_create_params_dependency` and somehow stored in the context variables.
Please check the code here: https://github.com/uriyyo/fastapi-pagination/blob/main/fastapi_pagination/api.py#L161

The parameters are being fetched using `__params_type__` of the mentioned `response_model`. We have `PageModel` in our case.
If you add any test variable in `PageModel.__params_type__`, it will be reflected in the list of parameters on http://localhost:8001/docs#/default/get_nodes_nodes_get
 
So, the /nodes handler function does not need to externally specify `limit` and `offset` in the signature.
`async def get_nodes(request: Request, kind: str = "node"):` would work too.

gctucker: I see.  So maybe there's some potential simplification to be done there.

api/db.py
JenySadadia: This is the implementation of `paginate` method:
```
async def paginate(
    collection: AsyncIOMotorCollection,
    query_filter: Optional[Dict[Any, Any]] = None,
    params: Optional[AbstractParams] = None,
    **kwargs: Any,
) -> AbstractPage:
    params = resolve_params(params)
    query_filter = query_filter or {}

    raw_params = params.to_raw_params()
    total = await collection.count_documents(query_filter)
    cursor = collection.find(query_filter, skip=raw_params.offset, limit=raw_params.limit, **kwargs)
    items = await cursor.to_list(length=raw_params.limit)

    return create_page(items, total, params)
```
Basically, it uses a `find` query with `skip` and `limit` parameters at the backend.

JenySadadia: Below are some statistics:

1. Without pagination:
```
data = await col.find(attributes, skip, limit).to_list(None)
list(model(**obj) for obj in data)
```
- Execution time: 
total nodes = 50                :0.009965658187866211 sec
total nodes=50, offset=2   :0.010911703109741211 sec
total nodes=50, limit=35   :0.013278484344482422 sec

2. Using `paginate`:
```
await paginate(collection=col, query_filter=attributes)
```
- Execution time:
total nodes = 50                :0.004123210906982422 sec
total nodes=50, offset=2   :0.004777431488037109 sec
total nodes=50, limit=35   :0.004402637481689453 sec

gctucker: Nice.  I guess the time would be less when fetching a smaller number of nodes? (either lower limit or last page)

JenySadadia: Yes, below are the numbers:
**1. Without pagination:**

- total:50 limit:10
  execution time: 0.007180929183959961

- total:50 offset:48
  execution time: 0.008078813552856445	
	
- total:92 offset:50 limit=50
  execution time: 0.009499788284301758

**2.  With pagination:**

- total:50 limit:10
  execution time 0.003947257995605469
	
- total:50 offset:48 limit=50
  execution time: 0.0029668807983398438
	
- total:92 offset:50 limit=50
  execution time: 0.005255937576293945

gctucker: OK I think all these numbers are very small and the difference based on the number of nodes is actually noise compared to the rest of the time to perform a request.  Well I think we'll need some end-to-end testing to really test entire features like that, at least it seems to be working as expected.

gctucker: With "no pagination", I guess that means without the Mongo DB paginated queries?  I see there's still the limit / offset arguments but I guess they were just passed to the API and not the DB query?

api/db.py
```diff
@@ -6,6 +6,7 @@
 
 """Database abstraction"""
 
+from fastapi_pagination.ext.motor import paginate
```
gctucker, line 9: Maybe keep the imports in alphabetical order?  It's not a strict requirement nor a `pylint` check but it helps avoid duplicating lines or finding existing imports when the list grows in general.

test/test_node_handler.py
```diff
@@ -152,7 +157,7 @@ def test_get_nodes_by_attributes_endpoint(mock_get_current_user,
             )
         print("response.json()", response.json())
         assert response.status_code == 200
-        assert len(response.json()) > 0
+        assert len(response.json().get('items')) > 0
```
gctucker, line 160:
```suggestion
        assert len(response.json()['items']) > 0
```

api/db.py
JenySadadia: At the moment, I only have 50 to 100 nodes in the DB, that's why we can not get a massive difference in the performance.

gctucker: Yes, we'll get some more realistic results with the staging instance.

api/db.py
JenySadadia: Yes, I updated the import order.

gctucker (approved):

PR #186, opened 2022-09-27
Password prompt

mgalka:
- Use getpass() instead of input() to hide password user input
- Return Unix exit codes
Update docs "Getting started" section accordingly


create_admin_user
```diff
@@ -26,22 +29,26 @@ def get_hashed_password(password, url):
...
 
 
 if __name__ == "__main__":
-
     parser = argparse.ArgumentParser("KernelCI API Admin")
     parser.add_argument('username', help="User name")
     parser.add_argument('--url', default=DEFAULT_URL, help="API host URL")
     args = parser.parse_args()
-    create_admin(args)
+    sys.exit(0 if main(args) else 1)
```
JenySadadia, line 54: Need to add a blank line at the end.

mgalka: good point

mgalka:
tested OK locally

PR #188, opened 2022-09-28
api.models: add a Node.debug attribute for development purposes

hardboprobot:
Add a Node.debug attribute (a dict) to enable debug options and flags used in certain development scenarios to enable node-specific debug features.

Signed-off-by: Ricardo Cañuelo <ricardo.canuelo@collabora.com>


JenySadadia (changes requested):
Need to fix the `pytest` errors.

test/test_node_handler.py
```diff
@@ -87,6 +87,7 @@ def test_create_node_endpoint(mock_get_current_user, mock_init_sub_id,
             'state',
             'timeout',
             'updated',
+            'debug',
```
JenySadadia, line 90: we prefer to maintain alphabetical order here.

hardboprobot: Ack, I'll fix it and rebase

api/models.py
```diff
@@ -201,6 +201,10 @@ class Node(DatabaseModel):
         ge=0.0,
         le=24.0
     )
+    debug: Optional[Dict] = Field(
+        description='Additional debug fields. Can be ignored for all \
```
JenySadadia, line 205: Need to mention what kind of debug information will be stored here e.g any kind of URL or logs.

hardboprobot: The idea is that this could be pretty open. Right now I think it could be useful for implementing alternative workflows for development, so it's not intended to be used for debug info per se, but to enable debug flags specific to some pipeline steps.

PR #189, opened 2022-09-30
T38051 Update "pipeline-details" for timeout and tarball

JenySadadia:
- Update pipeline flow diagram for timeout service.
- Since we have merged checkout and tarball node into a single checkout node, need to update the documentation accordingly.

### DIFF
--- a/doc/pipeline-details.md
+++ b/doc/pipeline-details.md
@@ -15,16 +15,15 @@ flowchart
     subgraph trigger_service[Trigger Service]
         trigger[Trigger] --> kernel_revision{New kernel <br/>revision ?}
         kernel_revision --> |No| trigger
-        kernel_revision --> |Yes| checkout[Create 'checkout' node <br />state=available, result=None, set holdoff]
+        kernel_revision --> |Yes| checkout[Create 'checkout' node <br />state=running, result=None]
     end
     subgraph tarball_service[Tarball Service]
-        checkout --> |event: <br />checkout created, state=available| tarball[Tarball]
-        tarball --> tarball_node[Create 'tarball' node <br />state=running, result=None, holdoff=None <br />Update 'checkout' node <br />with describe and artifacts]
-        tarball_node --> upload_tarball[Create and upload tarball to the storage <br />state=available, result=None, set holdoff]
+        upload_tarball[Create and upload tarball to the storage]
+        checkout --> |event: <br />checkout created, state=running| upload_tarball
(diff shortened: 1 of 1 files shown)


JenySadadia:
Documentation on staging:
https://static.staging.kernelci.org/docs/api/pipeline-details/

doc/pipeline-details.md
```diff
@@ -15,15 +15,14 @@ flowchart
     subgraph trigger_service[Trigger Service]
         trigger[Trigger] --> kernel_revision{New kernel <br/>revision ?}
         kernel_revision --> |No| trigger
```
gctucker, line 17:
```
|No| trigger
```
What does `trigger` mean here?  I think it should be something more like `sleep`.

doc/pipeline-details.md
```diff
@@ -15,15 +15,14 @@ flowchart
...
-        kernel_revision --> |Yes| checkout[Create 'checkout' node <br />state=available, result=None, set holdoff]
+        kernel_revision --> |Yes| checkout[Create 'checkout' node <br />state=running, result=None]
     end
     subgraph tarball_service[Tarball Service]
-        checkout --> |event: <br />checkout created, state=available| tarball[Tarball]
-        tarball --> tarball_node[Create 'tarball' node <br />state=running, result=None, holdoff=None <br />Update 'checkout' node <br />with describe and artifacts]
-        tarball_node --> upload_tarball[Create and upload tarball to the storage <br />state=available, result=None, set holdoff]
+        checkout --> |event: <br />checkout created, state=running| tarball[Tarball]
```
gctucker, line 21: What does `tarball[Tarball]` mean here?  I think the block from the `runner` service about uploading the tarball should be moved here instead as that's what the tarball service does.

JenySadadia: Definitely.
Those are some glitches that may be introduced when I renamed `tarball` node.
Need to fix this.

JenySadadia: I have created it with mermaid live editor and the block you mentioned is in `tarball` service itself.
I wonder why it is different on staging.
Attaching the flowchart from https://mermaid.live

image

doc/pipeline-details.md
JenySadadia: Yes, will fix this.

doc/pipeline-details.md
JenySadadia: Even https://github.com/kernelci/kernelci-api/blob/staging.kernelci.org/doc/pipeline-details.md matches with the flowchart I attached.

gctucker (approved):

PR #190, opened 2022-10-05
T38131 api.models: fix node state validation

JenySadadia:
Need to handle a case where node update operation
do not change the node state.

Fixes: 1f96e91 ("api.models: implement `Node.validate_node_transition`")

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -287,6 +287,11 @@ async def put_node(node_id: str, node: Node, token: str = Depends(get_user)):
     try:
         node.id = ObjectId(node_id)
         node_from_id = await get_node(node.id)
+        if not node_from_id:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail=f"Node not found with id: {node.id}"
+            )
         is_valid, message = node_from_id.validate_node_state_transition(
             node.state)
         if not is_valid:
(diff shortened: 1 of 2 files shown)


api/main.py
```diff
@@ -286,16 +286,22 @@ async def put_node(node_id: str, node: Node, token: str = Depends(get_user)):
     """Update an already added node"""
     try:
         node.id = ObjectId(node_id)
-        node_from_id = await get_node(node.id)
-        is_valid, message = node_from_id.validate_node_transition(node.state)
+        node_from_id = await db.find_by_id(Node, node_id)
+        if not node_from_id:
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail=f"Node not found with id: {node.id}"
```
mgalka, line 293: I think we should use here HTTP 404. Even the detailed message says that node hasn't been found.

api/models.py
```diff
@@ -239,8 +239,10 @@ def translate_fields(cls, params: dict):
             translated['parent'] = ObjectId(parent)
         return translated
 
-    def validate_node_transition(self, new_state):
+    def validate_node_state_transition(self, new_state):
         """Validate Node.state transitions"""
+        if new_state == self.state:
+            return True, "No validation is required"
```
mgalka, line 245: Maybe it'd be better to give more informative message like: 
```python
f"Transition to the same state: { new_state }. No validation required."
```

mgalka (commented):
minor remarks

api/main.py
JenySadadia: Definitely. It was a blunder.

mgalka (approved):
LGTM

PR #191, opened 2022-10-06
T38061 Validate parent in POST node handler

JenySadadia:
Implement a method to check if the node is available to create child nodes. This is needed to avoid the creation of child nodes if the parent is still running or closing.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -269,6 +269,19 @@ async def get_root_node(node_id: str):
 async def post_node(node: Node, token: str = Depends(get_user)):
     """Create a new node"""
     try:
+        if node.parent:
+            parent = await db.find_by_id(Node, node.parent)
+            if not parent:
+                raise HTTPException(
+                    status_code=status.HTTP_404_NOT_FOUND,
+                    detail=f"Parent not found with id: {node.parent}"
+                )
+            is_valid, message = parent.validate_parent()
+            if not is_valid:
+                raise HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST,
+                    detail=message
+                )
         obj = await db.create(node)
         operation = 'created'
     except ValueError as error:
--- a/api/models.py
+++ b/api/models.py
@@ -256,6 +256,13 @@ available state'
             return False, f"Transition not allowed with state: {new_state}"
         return True, "Transition validated successfully"
 
+    def validate_parent(self):
+        """Validate the parent node's state before creating child nodes"""
+        if self.state != 'available':
+            return False, f"The node is unavailable to create child node: \
+{self.id}"
+        return True, f"The node is available to create child node: {self.id}"
+
 
 class Hierarchy(BaseModel):
     """Hierarchy of nodes with child nodes"""


api/main.py
```diff
@@ -269,6 +269,14 @@ async def get_root_node(node_id: str):
 async def post_node(node: Node, token: str = Depends(get_user)):
     """Create a new node"""
     try:
+        if node.parent:
+            parent = await get_node(node.parent)
+            is_valid, message = parent.validate_parent()
```
hardboprobot, line 274: This fails when trying to post a node with a "parent" field containing an ObjectID that points nowhere:

```
kernelci-api | INFO:     172.20.0.1:59906 - "POST /node HTTP/1.1" 500 Internal Server Error
kernelci-api | ERROR:    Exception in ASGI application
kernelci-api | Traceback (most recent call last):
kernelci-api |   File "/home/kernelci/.local/lib/python3.10/site-packages/uvicorn/protocols/http/httptools_impl.py", line 398, in run_asgi
kernelci-api |     result = await app(self.scope, self.receive, self.send)
kernelci-api |   File "/home/kernelci/.local/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
kernelci-api |     return await self.app(scope, receive, send)
kernelci-api |   File "/home/kernelci/.local/lib/python3.10/site-packages/fastapi/applications.py", line 208, in __call__
kernelci-api |     await super().__call__(scope, receive, send)
kernelci-api |   File "/home/kernelci/.local/lib/python3.10/site-packages/starlette/applications.py", line 112, in __call__
kernelci-api |     await self.middleware_stack(scope, receive, send)
kernelci-api |   File "/home/kernelci/.local/lib/python3.10/site-packages/starlette/middleware/errors.py", line 181, in __call__
kernelci-api |     raise exc from None
kernelci-api |   File "/home/kernelci/.local/lib/python3.10/site-packages/starlette/middleware/errors.py", line 159, in __call__
kernelci-api |     await self.app(scope, receive, _send)
kernelci-api |   File "/home/kernelci/.local/lib/python3.10/site-packages/starlette/exceptions.py", line 82, in __call__
kernelci-api |     raise exc from None
kernelci-api |   File "/home/kernelci/.local/lib/python3.10/site-packages/starlette/exceptions.py", line 71, in __call__
kernelci-api |     await self.app(scope, receive, sender)
kernelci-api |   File "/home/kernelci/.local/lib/python3.10/site-packages/starlette/routing.py", line 580, in __call__
kernelci-api |     await route.handle(scope, receive, send)
kernelci-api |   File "/home/kernelci/.local/lib/python3.10/site-packages/starlette/routing.py", line 241, in handle
kernelci-api |     await self.app(scope, receive, send)
kernelci-api |   File "/home/kernelci/.local/lib/python3.10/site-packages/starlette/routing.py", line 52, in app
kernelci-api |     response = await func(request)
kernelci-api |   File "/home/kernelci/.local/lib/python3.10/site-packages/fastapi/routing.py", line 226, in app
kernelci-api |     raw_response = await run_endpoint_function(
kernelci-api |   File "/home/kernelci/.local/lib/python3.10/site-packages/fastapi/routing.py", line 159, in run_endpoint_function
kernelci-api |     return await dependant.call(**values)
kernelci-api |   File "/home/kernelci/./api/main.py", line 275, in post_node
kernelci-api |     is_valid, message = parent.validate_parent()
kernelci-api | AttributeError: 'NoneType' object has no attribute 'validate_parent'
```

Shouldn't there be an additional check after `get_node` to continue only if `parent` is actually a Node?

hardboprobot (changes requested):

api/main.py
JenySadadia: Yes, I missed that check. Thanks for testing and pointing that out.
I fixed it.

hardboprobot (approved):

JenySadadia:
> Approved, but see the review of [kernelci/kernelci-pipeline#161](https://github.com/kernelci/kernelci-pipeline/pull/161) as well. TL;DR: why do we want to restrict the creation of child nodes to nodes that are "available" only?

As a part of the node schema design, we have introduced different node states to moderate its life cycle.
For instance, a checkout node will create a source tarball while being in a `running` state. When the tarball upload is complete, the node will be moved to `available` state. Now the tests can be run using the checkout node. It can not run tests (have child nodes) without having the source tarball ready. That is why we need a restriction like this to check if the node is available to have further child nodes.
Please refer to this if you have a moment: https://kernelci.org/docs/api/api-details/#state-diagram

alexandrasp (approved):
LGTM

api/main.py
```diff
@@ -269,6 +269,19 @@ async def get_root_node(node_id: str):
 async def post_node(node: Node, token: str = Depends(get_user)):
     """Create a new node"""
     try:
+        if node.parent:
+            parent = await db.find_by_id(Node, node.parent)
+            if not parent:
+                raise HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST,
```
mgalka, line 276: I think it should rather get HTTP `404` status than `400`. Is there a reason we want to return `Bad Request` here?

mgalka (commented):
One small comment. Actually pretty similar to the one from #190

mgalka (approved):

Issue #196, opened 2022-11-09
Enable support to filter nodes based on timestamp fields


JenySadadia:
Since we have various node fields related to timestamps i.e. `created`, `updated`, `timeout`, and `holdoff`, need to enable support to query nodes with them.

For example, we should be able to get nodes matching created value:
`curl 'http://localhost:8001/nodes?created=2022-11-07T11:11:27.873000'`

Also, we need comparison operator (gt, lt, lte, gte) support.
For instance, regression tracker will need to query nodes less than provided created timestamp to detect regressions.
`curl 'http://localhost:8001/nodes?created__lt=2022-11-07T11:11:27.873000'`

We need to enable support for the below syntax in request params:
`field__operator`

PR #197, opened 2022-11-09
T38872 Enable support to filter nodes based on timestamp fields

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/196

- api.models: validate timestamp filter
- api.models: translate params with comparison operator
- api.main: check for successful request param translation


JenySadadia:
Tested OK on staging:

```
$ curl 'https://staging.kernelci.org:9000/nodes?created=2022-11-09T22:18:53.320000&limit=1'
{"items":[{"_id":"636c274dd51784bc3c9cfe8d","kind":"node","name":"test_big_hole_assigned_dynamic_all","path":["checkout","kunit","stackinit","test_big_hole_assigned_dynamic_all"],"group":"kunit","revision":{"tree":"mainline","url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git","branch":"master","commit":"f67dd6ce0723ad013395f20a3f79d8a437d3f455","describe":"v6.1-rc4-20-gf67dd6ce0723","version":{"version":6,"patchlevel":1,"sublevel":null,"extra":"-rc4-20-gf67dd6ce0723","name":null}},"parent":"636c274dd51784bc3c9cfe5e","state":"done","result":null,"artifacts":null,"created":"2022-11-09T22:18:53.320000","updated":"2022-11-09T22:18:53.320000","timeout":"2022-11-10T22:18:53.320000","holdoff":null}],"total":11,"limit":1,"offset":0}
```

```
$ curl 'https://staging.kernelci.org:9000/nodes?created=2022-11-09T22:18:53.320000&timeout=2022-11-10T22:18:53.320000&limit=1'
{"items":[{"_id":"636c274dd51784bc3c9cfe8d","kind":"node","name":"test_big_hole_assigned_dynamic_all","path":["checkout","kunit","stackinit","test_big_hole_assigned_dynamic_all"],"group":"kunit","revision":{"tree":"mainline","url":"https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git","branch":"master","commit":"f67dd6ce0723ad013395f20a3f79d8a437d3f455","describe":"v6.1-rc4-20-gf67dd6ce0723","version":{"version":6,"patchlevel":1,"sublevel":null,"extra":"-rc4-20-gf67dd6ce0723","name":null}},"parent":"636c274dd51784bc3c9cfe5e","state":"done","result":null,"artifacts":null,"created":"2022-11-09T22:18:53.320000","updated":"2022-11-09T22:18:53.320000","timeout":"2022-11-10T22:18:53.320000","holdoff":null}],"total":11,"limit":1,"offset":0}
```

```
$ curl 'https://staging.kernelci.org:9000/nodes?timeout=2022-11-10T22:18:53.32000'
{"detail":"Invalid request parameters: Invalid isoformat string '2022-11-10T22:18:53.32000'"}
```

```
$ curl 'https://staging.kernelci.org:9000/nodes?timeout__gt=2022-11-10T22:18:53.32000'
{"detail":"Failed to translate request parameters: Invalid isoformat string '2022-11-10T22:18:53.32000'"}
```

```
$ curl 'https://staging.kernelci.org:9000/nodes?timeout__lt=2022-11-10T22:18:53.320000&limit=2'
{"items":[{"_id":"63636116ad8e3c87022a6ccc","kind":"node","name":"checkout","path":["checkout"],"group":null,"revision":{"tree":"kernelci","url":"https://github.com/kernelci/linux.git","branch":"staging-mainline","commit":"a050d367bd20b276e5dbe15a31cee16cafba5f2a","describe":"staging-mainline-20221102.0","version":{"version":6,"patchlevel":1,"sublevel":null,"extra":"-rc3-47-ga050d367bd20","name":null}},"parent":null,"state":"done","result":"pass","artifacts":{"tarball":"http://172.17.0.1:8002/linux-kernelci-staging-mainline-staging-mainline-20221102.0.tar.gz"},"created":"2022-11-03T06:35:02.051000","updated":"2022-11-04T09:28:47.267000","timeout":"2022-11-04T06:35:02.051000","holdoff":"2022-11-03T06:45:34.246000"},{"_id":"63636117ad8e3c87022a6ccd","kind":"node","name":"checkout","path":["checkout"],"group":null,"revision":{"tree":"kernelci","url":"https://github.com/kernelci/linux.git","branch":"staging-next","commit":"41278526a3eea44f63f8e623e753356328af1c8c","describe":"staging-next-20221102.0","version":{"version":6,"patchlevel":1,"sublevel":null,"extra":"-rc3-4837-g41278526a3ee","name":null}},"parent":null,"state":"done","result":null,"artifacts":{"tarball":"http://172.17.0.1:8002/linux-kernelci-staging-next-staging-next-20221102.0.tar.gz"},"created":"2022-11-03T06:35:03.100000","updated":"2022-11-03T06:46:15.942000","timeout":"2022-11-04T06:35:03.100000","holdoff":"2022-11-03T06:46:08.965000"}],"total":47,"limit":2,"offset":0}
```

mgalka:
@JenySadadia Pydantic has build in [validation mechanism](https://pydantic-docs.helpmanual.io/usage/validators/) is there a reason why we're not using it?

JenySadadia:
> @JenySadadia Pydantic has build in [validation mechanism](https://pydantic-docs.helpmanual.io/usage/validators/) is there a reason why we're not using it?

@mgalka The validators you have mentioned can be used to validate Pydantic object fields.
In this case, `validate_params` is implemented to validate request query parameters (in case of GET requests).

api/models.py
```diff
@@ -215,6 +215,19 @@ def set_timeout(self, hours: int = 24):
     def update(self):
         self.updated = datetime.utcnow()
 
+    @classmethod
```
gctucker, line 218: Is that not done already by pydantic?

api/main.py
```diff
@@ -216,7 +216,13 @@ async def get_nodes(request: Request, kind: str = "node"):
                 detail=f"Invalid request parameters: {msg}"
         )
 
-    translated_params = model.translate_fields(query_params)
+    is_translated, translated_params = model.translate_fields(query_params)
+    if not is_translated:
```
gctucker, line 220: It seems like there's a lot of overhead with throwing exceptions, catching exceptions and returning status.  Maybe it would be good to review the whole error propagation and handling code.  Ideally, I think exceptions should only be caught in the endpoint implementation in `main.py` to convert them to HTTP responses.

api/models.py
```diff
@@ -247,7 +265,37 @@ def translate_fields(cls, params: dict):
         parent = params.get('parent')
         if parent:
             translated['parent'] = ObjectId(parent)
-        return translated
+
+        timestamp_fields = ('created', 'updated', 'timeout', 'holdoff')
+
+        operator_map = {
```
gctucker, line 271: I wonder if we could make this more generic to allow any field to be compared with a value rather than just timestamps.

api/models.py
```diff
@@ -247,7 +265,37 @@ def translate_fields(cls, params: dict):
...
+            'lte': '$lte',
+            'gt': '$gt',
+            'gte': '$gte',
+        }
+
+        for key in params.keys():
+            if key in timestamp_fields:
+                translated[key] = datetime.fromisoformat(translated[key])
+            if '__' in key:
+                field, operator = key.split('__')
```
gctucker, line 282: If `__` is twice in the key then it won't really work.  I think it's better to just call `split` without an `if` and then check the returned value.

api/models.py
gctucker: Also these are MongoDB operators, but the models defined in this module shouldn't have any knowledge of the database engine being used.  So this operator translation should be done in `Database` instead.

api/models.py
gctucker: Sounds like the same question @mgalka asked a while ago, and I still don't understand the rationale for not using pydantic's standard datetime validation.

JenySadadia: As I already explained, pydantic validators defined with `validator` decorator will be invoked when an object of a model is instantiated. It can be used while creating or updating documents e.g.
```
try:
    Node(name="kunit-test-case", path=["checkout","kunit","kunit-test-case"],...)
except ValidationError as e:
    print(e)
```
On the other hand, `validate_params` is defined to validate request query parameters received with GET request.
These request parameters are being used as attributes to get required documents from the database and not to create any documents.
So, this is the validation of request query params and not object fields.

gctucker: When receiving a request in `main.py`, and the handler function specifies a pydantic model, then validation will be made on the incoming data.  For example:
```
@app.post('/node', response_model=Node)
```
The input data will need to produce a valid `Node` object, if the pydantic validation doesn't accept it then an HTTP response is directly sent to the client.

gctucker: I think the issue here is that some use-case from the endpoint handler is being "leaked" into the model.  The model shouldn't know about how it's being used in other parts of the code.  In any case, there's some code in pydantic to validate fields such as timestamps and I still don't see any good reason why we would not be using it.  The actual problem is how to validate parameters for GET requests, and maybe that should be done outside the model.

JenySadadia: That's exactly what I am trying to say.
We can use the pydantic validators in POST and PUT requests but in this case, we need to validate request params manually.
I am not sure if we should move out this validation from the model, as these validations are model specific and would be different for every model.

api/models.py
gctucker: This should be a simple rework, do you see any issue with it?

api/models.py
JenySadadia: I agree, the operator translation should be in DB layer.
I am working on it.

api/models.py
JenySadadia: Yes, I'll fix this.

api/models.py
JenySadadia: I updated operator mapping to be more generic to apply it to all attributes rather than just timestamp fields.

api/main.py
JenySadadia: I think that can be a follow-up PR.

gctucker: OK

api/models.py
JenySadadia: Yes, if we want to use [FastAPI validation](https://fastapi.tiangolo.com/tutorial/query-params-str-validations/#add-more-validations) for query params, we need to mention all the params explicitly which is not feasible in our case as we have so many fields.

gctucker: Yes but pydantic is separate from FastAPI so we might be able to use pydantic validation directly on the fields.  I'm not sure, just a thought.

JenySadadia: Currently, data type validations are already being handled by default for POST and PUT requests.
I don't think we need any extra validations atm.

gctucker: I mean, doing something like `pydantic.validate(type=datetime, input_value)` (making things up here...) could avoid us having to write some custom validation logic.  But I think we can park that for now, it's not a blocker for this PR.

gctucker (approved):
OK the rework with the `operator_map` looks good.  Let's just check this is working as expected on staging and it should be ready to be merged.

JenySadadia:
> OK the rework with the `operator_map` looks good. Let's just check this is working as expected on staging and it should be ready to be merged.

Tested OK on staging:

```
$ curl 'https://staging.kernelci.org:9000/nodes?timeout__gt=2022-09-28T11:07:32.543000&limit=1'
{"items":[{"_id":"6332d9741a45d41c279e7a41","kind":"node","name":"example","path":["checkout","kunit","example"],"group":"kunit","revision":{"tree":"kernelci","url":"https://github.com/kernelci/linux.git","branch":"staging-mainline","commit":"7f036eb8d7a5ff2f655c5d949343bac6a2928bce","describe":"staging-mainline-20220927.0","version":{"version":6,"patchlevel":0,"sublevel":null,"extra":"-rc7-36-g7f036eb8d7a5","name":null}},"parent":"6332d92f1a45d41c279e7a06","state":"done","result":null,"artifacts":null,"created":"2022-09-27T11:07:32.544000","updated":"2022-09-27T11:07:32.544000","timeout":"2022-09-28T11:07:32.544000","holdoff":null}],"total":65248,"limit":1,"offset":0}
```

```
$ curl 'https://staging.kernelci.org:9000/nodes?created__gt=2022-09-28T11:07:32.43000&limit=1'
{"detail":"Invalid request parameters: Invalid isoformat string '2022-09-28T11:07:32.43000'"}
```

```
$ curl 'https://staging.kernelci.org:9000/nodes?created__g=2022-09-28T11:07:32.543000&limit=1'
{"items":[],"total":0,"limit":1,"offset":0}
```

gctucker:
> > OK the rework with the `operator_map` looks good. Let's just check this is working as expected on staging and it should be ready to be merged.
> 
> Tested OK on staging:
> 
> ```
> $ curl 'https://staging.kernelci.org:9000/nodes?timeout__gt=2022-09-28T11:07:32.543000&limit=1'
> {"items":[{"_id":"6332d9741a45d41c279e7a41","kind":"node","name":"example","path":["checkout","kunit","example"],"group":"kunit","revision":{"tree":"kernelci","url":"https://github.com/kernelci/linux.git","branch":"staging-mainline","commit":"7f036eb8d7a5ff2f655c5d949343bac6a2928bce","describe":"staging-mainline-20220927.0","version":{"version":6,"patchlevel":0,"sublevel":null,"extra":"-rc7-36-g7f036eb8d7a5","name":null}},"parent":"6332d92f1a45d41c279e7a06","state":"done","result":null,"artifacts":null,"created":"2022-09-27T11:07:32.544000","updated":"2022-09-27T11:07:32.544000","timeout":"2022-09-28T11:07:32.544000","holdoff":null}],"total":65248,"limit":1,"offset":0}
> ```
Nice.  I would still like to see the jobs running on staging, we'll have it in a few hours.

> ```
> $ curl 'https://staging.kernelci.org:9000/nodes?created__gt=2022-09-28T11:07:32.43000&limit=1'
> {"detail":"Invalid request parameters: Invalid isoformat string '2022-09-28T11:07:32.43000'"}
> ```
So the error here was a missing `0`, right?

> ```
> $ curl 'https://staging.kernelci.org:9000/nodes?created__g=2022-09-28T11:07:32.543000&limit=1'
> {"items":[],"total":0,"limit":1,"offset":0}
> ```
There seems to be a problem with `created__g`, should it be `created__gt` instead for "greater than"?  Or if that's an invalid operator, should it be returning an HTTP 400 instead?

JenySadadia:
> > > OK the rework with the `operator_map` looks good. Let's just check this is working as expected on staging and it should be ready to be merged.
> > 
> > 
> > Tested OK on staging:
> > ```
> > $ curl 'https://staging.kernelci.org:9000/nodes?timeout__gt=2022-09-28T11:07:32.543000&limit=1'
> > {"items":[{"_id":"6332d9741a45d41c279e7a41","kind":"node","name":"example","path":["checkout","kunit","example"],"group":"kunit","revision":{"tree":"kernelci","url":"https://github.com/kernelci/linux.git","branch":"staging-mainline","commit":"7f036eb8d7a5ff2f655c5d949343bac6a2928bce","describe":"staging-mainline-20220927.0","version":{"version":6,"patchlevel":0,"sublevel":null,"extra":"-rc7-36-g7f036eb8d7a5","name":null}},"parent":"6332d92f1a45d41c279e7a06","state":"done","result":null,"artifacts":null,"created":"2022-09-27T11:07:32.544000","updated":"2022-09-27T11:07:32.544000","timeout":"2022-09-28T11:07:32.544000","holdoff":null}],"total":65248,"limit":1,"offset":0}
> > ```
> 
> Nice. I would still like to see the jobs running on staging, we'll have it in a few hours.
> 
> > ```
> > $ curl 'https://staging.kernelci.org:9000/nodes?created__gt=2022-09-28T11:07:32.43000&limit=1'
> > {"detail":"Invalid request parameters: Invalid isoformat string '2022-09-28T11:07:32.43000'"}
> > ```
> 
> So the error here was a missing `0`, right?
> 
Exactly.
> > ```
> > $ curl 'https://staging.kernelci.org:9000/nodes?created__g=2022-09-28T11:07:32.543000&limit=1'
> > {"items":[],"total":0,"limit":1,"offset":0}
> > ```
> 
> There seems to be a problem with `created__g`, should it be `created__gt` instead for "greater than"? Or if that's an invalid operator, should it be returning an HTTP 400 instead?

This is the case when the operator is not available on the specified map. I haven't raised 400 to have scope for node fields with `__` in the name.

gctucker:
OK, or we can add to the design docs that fields shouldn't have __ in them. The field names are part of the models, not user input so we can decide rules for them.

api/db.py
```diff
@@ -26,6 +26,13 @@ class Database:
         Regression: 'regression',
     }
 
+    operator_map = {
```
gctucker, line 29: The universal convention for "constant" variables is to use capitalised names.  Not a real blocker, but see `COLLECTIONS` and not `collections` etc.  So keeping this consistent would probably be a good idea.

api/models.py
```diff
@@ -215,6 +215,19 @@ def set_timeout(self, hours: int = 24):
     def update(self):
         self.updated = datetime.utcnow()
 
+    @classmethod
+    def validate_timestamp(cls, params: dict):
+        """Validate timestamp fields in `params` to be valid isoformat
+        string"""
+        timestamp_fields = ('created', 'updated', 'timeout', 'holdoff')
+        for key in params.keys():
+            if any(field in key for field in timestamp_fields):
```
gctucker, line 224: Can we not do something like this instead?
```python
for key, value in params.items():
  if key in timestamp_fields:
    datetime.fromisoformat(value)
```

api/db.py
```diff
@@ -65,6 +72,20 @@ async def find_by_id(self, model, obj_id):
         obj = await col.find_one({'_id': ObjectId(obj_id)})
         return model(**obj) if obj else None
 
+    def _operator_translation(self, attributes):
+        translated = attributes.copy()
```
gctucker, line 76: Instead of starting with a copy and then iterating through the attributed to edit it, it would seem easier to follow the implementation if it started with an empty dictionary and then arguments would be added to it either as-is if they're not operators or translated if they are.

api/db.py
```diff
@@ -65,6 +72,20 @@ async def find_by_id(self, model, obj_id):
...
+        for key in attributes.keys():
+            field = key.split('__')
+            if len(field) == 2:
+                if not self.operator_map.get(field[1]):
+                    raise ValueError(f"No operator found matching \
+'{field[1]}'")
```
gctucker, line 82: It's much better to avoid wrapping strings, it's harder to read when the line goes back to column 0 and it makes it harder to search for messages in the code.  This should pass `pycodestyle` too:
```python
                    raise ValueError(
                        f"No operator found matching '{field[1]}'"
                    )
```

gctucker (commented):
A few things that I think could be improved.  Still they're probably not blockers so we could have this PR merged if it's working as expected on staging to have the feature completely available and then some follow-up PRs could be made to clean a few things up.

api/db.py
JenySadadia: I think this is just another way of doing it.

api/models.py
JenySadadia: I have dropped `validate_timestamp` method and moved `translate_fields` to try-except block to catch exceptions when invalid isoformat string will be received.

JenySadadia: However, `if key in timestamp_fields:` will not be able to evaluate correctly when a field is received with an operator e.g. `'created__lt' in ('created', 'updated', 'timeout', 'holdoff')` will be `False`.

api/db.py
gctucker: Yes but I mean the way I'm suggesting it has the advantage of defining / constructing the output only once.  It could even be turned into a comprehension I believe, not that it's necessarily always an improvement in Python.

api/models.py
gctucker: Ah I see.  Maybe the operators should be parsed first so just the field names appear here, but I'm not sure what's the best way to implement that.

api/main.py
JenySadadia: Added a follow-up PR for this https://github.com/kernelci/kernelci-api/pull/200

api/main.py
```diff
@@ -216,8 +216,14 @@ async def get_nodes(request: Request, kind: str = "node"):
                 detail=f"Invalid request parameters: {msg}"
         )
 
-    translated_params = model.translate_fields(query_params)
-    return await db.find_by_attributes(model, translated_params)
+    try:
+        translated_params = model.translate_fields(query_params)
+        return await db.find_by_attributes(model, translated_params)
```
gctucker, line 221: Shouldn't this line be outside of the try-catch block?  The `ValueError` exceptions appear to only be meant for catching invalid input in other endpoint handlers.

api/main.py
```diff
@@ -241,8 +247,15 @@ async def get_nodes_count(request: Request, kind: str = "node"):
                 status_code=status.HTTP_400_BAD_REQUEST,
                 detail=f"Invalid request parameters: {msg}"
             )
-    translated_params = model.translate_fields(query_params)
-    return await db.count(model, translated_params)
+
+    try:
+        translated_params = model.translate_fields(query_params)
```
gctucker, line 252: I've lost track of why this can't be done in `validate_params()`, but looking at the code here it doesn't seem right to be mixing the actual code handling logic and input validation logic.

api/main.py
JenySadadia: Operator mapping in the `db` can result in `ValueError` if the operator is not found.
That's why I added it to try-catch block.

api/main.py
JenySadadia: If it is by default raising `ValueError` for the conversion `datetime.fromisoformat`, why check it manually in `validate_params`?

api/models.py
JenySadadia: Yes, that can be done too. Let me see.

JenySadadia: Updated it. @gctucker

api/main.py
gctucker: Right, but it's happening after validation.  So the validation is getting mixed up with the implementation of the logic to handle the API query.

I think this ValueError raised by the database class is not a validation issue, it's just an exception that happens while executing the code.  We could add some actual operator validation instead.  I think it's not a step in the right direction to be bundling validation with logic here.  But also I don't think we should be blocking this PR based on that, the whole design of exceptions and error propagation should be reviewed now we have more code and new use-cases that didn't exist when the endpoint handlers were first written.

api/main.py
gctucker: See my previous comment.

gctucker:
By the way, staging runs are stuck again with this error:
```
fatal: Unable to create '/home/kernelci/data/src/linux/.git/index.lock': File exists.
```
I guess we should do something to prevent it by ensuring the git repo is put back in a good working state when the tarball job runs.

Regardless, I think this feature is only used in the regression service so I don't think we would see anything in the staging logs yet anyway.  So it seems ready to be merged now.

PR #198, opened 2022-11-16
Fix timeout value handling in API

gctucker:
Rather than implicitly setting the timeout to a fixed value in the `Database` class, set a default value of 6h in the Node base model and allow the user to set it to any arbitrary value.  Also drop the `Optional` hints for mandatory fields including `timeout`.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -158,6 +158,28 @@ class Revision(BaseModel):
     )
 
 
+class DefaultTimeout:
+    """Helper to create default values for timeout fields
+
+    The `hours` and `minutes` provided are used to create a `timedelta` object
+    available in the `.delta` attribute.  This can then be used to get a
+    timeout value used as a default when defining a non-optional field in a
+    model with the `.get_timeout()` method.
+    """
+
+    def __init__(self, hours=0, minutes=0):
+        self._delta = timedelta(hours=hours, minutes=minutes)
+
+    @property
+    def delta(self):
+        """Get the timedelta set in this object"""
+        return self._delta
+
+    def get_timeout(self):
+        """Get a timeout timestamp with current time and delta"""
+        return datetime.utcnow() + self.delta
+
+
 class Node(DatabaseModel):
     """KernelCI primitive node object model for generic test results"""
     kind: str = Field(
--- a/api/models.py
+++ b/api/models.py
@@ -221,7 +221,8 @@ URLs (e.g. URL to binaries or logs)'
         default_factory=datetime.utcnow,
         description='Timestamp when node was last updated'
     )
-    timeout: Optional[datetime] = Field(
+    timeout: datetime = Field(
+        default_factory=DefaultTimeout(hours=6).get_timeout,
         description='Node expiry timestamp'
     )
     holdoff: Optional[datetime] = Field(
--- a/api/models.py
+++ b/api/models.py
@@ -93,9 +93,6 @@ class DatabaseModel(ModelId):
     def create_indexes(cls, collection):
         """Method to create indexes"""
 
-    def set_timeout(self, hours: int):
-        """Method to set model timeout"""
-
 
 class User(DatabaseModel):
     """API user model"""
@@ -229,14 +226,6 @@ URLs (e.g. URL to binaries or logs)'
         description='Holdoff expiry timestamp for node to be in \
 available state'
     )
-
-    def set_timeout(self, hours: int = 24):
-        """Set Node timeout (expiry timestamp)
-
-        After the node's creation, it will timeout after the number
-        of hours (default: 24) provided.
-        """
-        self.timeout = self.created + timedelta(hours=hours)
 
     def update(self):
         self.updated = datetime.utcnow()
--- a/api/models.py
+++ b/api/models.py
@@ -180,7 +180,7 @@ class Node(DatabaseModel):
     parent: Optional[PyObjectId] = Field(
         description='Parent commit SHA'
     )
-    state: Optional[StateValues] = Field(
+    state: StateValues = Field(
         default=StateValues.RUNNING,
         description='State of the node'
     )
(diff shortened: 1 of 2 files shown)


api/models.py
```diff
@@ -191,30 +209,22 @@ class Node(DatabaseModel):
...
         description='Timestamp of node creation'
     )
-    updated: Optional[datetime] = Field(
+    updated: datetime = Field(
         default_factory=datetime.utcnow,
         description='Timestamp when node was last updated'
     )
-    timeout: Optional[datetime] = Field(
+    timeout: datetime = Field(
+        default_factory=DefaultTimeout(hours=6).get_timeout,
```
JenySadadia, line 221: Is there any specific reason to set 6 hrs as default?

api/models.py
```diff
@@ -158,6 +155,27 @@ class Revision(BaseModel):
...
+        self._delta = delta or timedelta(hours=hours, minutes=minutes)
+
+    @property
+    def delta(self):
+        """Get the timedelta set in this object"""
+        return self._delta
+
+    def get_timeout(self):
+        """Get a timeout timestamp with current time and delta"""
+        return datetime.utcnow() + self.delta
```
JenySadadia, line 176: Shouldn't we use `created` timestamp to calculate timeout?

api/models.py
gctucker: Because less than that is too short and longer than that is too long...  Ideally the default should never be used, but if it's used then that's the kind of timeframe someone would expect results to arrive after a new kernel revision has been pushed to git.

api/models.py
gctucker: I'm not sure we can enforce `created` to be initialised first, and for similar reasons we can't pass it to the default_factory callable in the class definition.  In practice, it'll still be "now" between when the two fields are initialised.  If we ever get to the point where there's more than one minute difference between created and the time used to calculate the timeout then I think we'll have much more serious issues to address.

In case the client is providing the `created` time but no timeout while submitting a new node, then the timeout will most likely be more than 6h after creation time.  But I don't think that's an issue, and it's kind of expected.

gctucker: I've added a `start_time` argument so the caller could pass the `created` time or another start time if that's needed.  If not it will keep using `.utcnow()` by default.

gctucker: Actually I'm not sure we'll ever be able to use `start_time` with the original intention to use this class when defining models as we can't pass any values to it.  I was thinking we might be able to use it for `holdoff` when the node goes to `Available` state and the client hasn't specified a holdoff timeout, if we need to set a default one too.  If we don't need to set any default then I'll revert to the previous implementation with no `start_time`.

api/models.py
```diff
@@ -158,6 +155,31 @@ class Revision(BaseModel):
     )
 
 
+class DefaultTimeout:
+    """Helper to create default values for timeout fields"""
+
+    def __init__(self, hours=0, minutes=0, delta=None):
```
JenySadadia, line 161:
```suggestion
    def __init__(self, hours=6, minutes=0, delta=None):
```

api/models.py
```diff
@@ -191,30 +213,22 @@ class Node(DatabaseModel):
...
         description='Timestamp of node creation'
     )
-    updated: Optional[datetime] = Field(
+    updated: datetime = Field(
         default_factory=datetime.utcnow,
         description='Timestamp when node was last updated'
     )
-    timeout: Optional[datetime] = Field(
+    timeout: datetime = Field(
+        default_factory=DefaultTimeout(hours=6).get_timeout,
```
JenySadadia, line 225:
```suggestion
        default_factory=DefaultTimeout().get_timeout,
```

JenySadadia (changes requested):
The default hours should be directly set in the constructor.
Also, if the default is 0, and `delta` is None, the start time would be the timeout itself and the node will be timed-out immediately.

api/models.py
gctucker: I really don't like this, the default timeout value should be explicitly stated in the field definition.  Otherwise, changing the default value in the DefaultTimeout class will implicitly change the defaults in the models which would be very hard to track.  Also it means it's hard to figure out the default value when looking at a model definition as you need to look up the DefaultTimeout class implementation etc.

api/models.py
gctucker: The class could just have `delta`, the `hours` and `minutes` are to make it easier to use.  I guess I could just remove `delta` if it makes things harder to understand and we'll only have `hours` and `minutes` granularity.

But yes if nothing is provided then of course the timeout will be now.  However, the whole point of this class is to provide a default value.

api/models.py
JenySadadia: I am not able to think of any use case when a caller would need to pass the `start_time`.
Even how will the `delta` be dynamic with this implementation?
We need to receive it from the request parameter to make it client specific.

api/models.py
JenySadadia: Yes, I think it is better to drop `delta` from this.

api/models.py
JenySadadia: Makes sense.

api/models.py
gctucker: Right, so I'll revert to the previous implementation and always rely on `.utcnow()` as the start time.

api/models.py
gctucker: Done

JenySadadia (approved):
Looks good to me.
I think we can merge this and have a follow-up PR to make the time delta configurable from the client.

gctucker:
The time delta is already configurable from the client.  I've made the pipeline changes to make use of that already, will send some PRs soon (just had too many piling up in kernelci-pipeline lately).

gctucker:
@JenySadadia I think you could have merged it, right?

PR #202, opened 2022-12-08
T39454 doc/api-details: document comparison operators for request query params

JenySadadia:
Add information about request query params with comparison operator for `/nodes` and `/count` endpoints.
Also, add a note about field name restrictions.

### DIFF
--- a/doc/api-details.md
+++ b/doc/api-details.md
@@ -209,6 +209,12 @@ $ curl 'http://localhost:8001/count?name=checkout&artifacts.tarball=http://172.1
 1
 ```
 
+Same as `/nodes`, the `/count` endpoint also supports comparison operators for request query parameters.
+```
+$ curl 'http://localhost:8001/count?name=checkout&created__lt=2022-12-06T04:59:08.102000'
+3
+```
+
 ### State diagram
 
 The Node objects are governed by the following state machine:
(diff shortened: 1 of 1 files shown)


gctucker (commented):
I wonder how hard it would be to make these operators work for any request, rather than just `/nodes` and `/count`.  Have you considered doing this?

JenySadadia:
> I wonder how hard it would be to make these operators work for any request, rather than just `/nodes` and `/count`. Have you considered doing this?

I don't think we need to enable support for other requests, as comparison operators are only useful for matching values for GET requests.

gctucker:
I meant for _any_ GET request.  So I guess the answer to my question is no, and that's fine.  We might want to revisit things later as we keep adding endpoints that would also benefit from these operators without having to add a lot of boiler-plate code for each of them.

JenySadadia:
> I meant for _any_ GET request. So I guess the answer to my question is no, and that's fine. We might want to revisit things later as we keep adding endpoints that would also benefit from these operators without having to add a lot of boiler-plate code for each of them.

In case of new GET requests, we can easily enable support for comparison operators by using `model.translate_fields`  in the endpoint handler and `Database.operator_translation` for the DB operation. Hence, it will be using available methods rather than reworking.

JenySadadia:
Documentation on the staging: https://static.staging.kernelci.org/docs/api/api-details/#getting-nodes-back

mgalka (approved):
LGTM

Issue #205, opened 2022-12-23
Enable passing null as a request parameter for node parent


alexandrasp:
After the experimental work to connect the API to the old dashboard, it could be interesting to provide a way to filter by the root parent in case the parent reference is NULL. I just got all nodes in time-filtered, using `gte` feature, and on the front-end side, I filter considering parent information to show jobs by branch, tree and kernel version.

JenySadadia:
Yes, we can enable passing "null" as a request parameter value if it helps to reduce post-processing on the front-end side.
In fact, it should be implemented for all `Node` fields and not just for `parent`.

gctucker:
I think the title of this issue should be changed as it's not very descriptive of what it's really about.

alexandrasp:
I updated the title of the issue.

PR #206, opened 2022-12-29
T39710 Query nodes with null field values

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/205

Enable support for allowing `null` as the request query param value to `/nodes` and `/count` endpoints for GET requests.

e.g. Request to get root nodes from the database will be http://localhost:8001/nodes?parent=null

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -265,6 +265,8 @@ URLs (e.g. URL to binaries or logs)'
         timestamp_fields = ('created', 'updated', 'timeout', 'holdoff')
         # Split params with `__` in key and translate timestamp fields
         for key in params.keys():
+            if translated[key] is None:
+                continue
             if key in timestamp_fields:
                 translated[key] = datetime.fromisoformat(translated[key])
             field = key.split('__')
--- a/api/main.py
+++ b/api/main.py
@@ -218,6 +218,7 @@ async def get_nodes(request: Request, kind: str = "node"):
     for pg_key in ['limit', 'offset']:
         query_params.pop(pg_key, None)
 
+    query_params = await translate_null_query_params(query_params)
     is_valid, msg = model.validate_params(query_params)
     if not is_valid:
         raise HTTPException(
@@ -250,6 +251,7 @@ async def get_nodes_count(request: Request, kind: str = "node"):
 
     query_params = dict(request.query_params)
 
+    query_params = await translate_null_query_params(query_params)
     is_valid, msg = model.validate_params(query_params)
     if not is_valid:
         raise HTTPException(
--- a/api/main.py
+++ b/api/main.py
@@ -174,6 +174,15 @@ def get_password_hash(password: Password):
 # -----------------------------------------------------------------------------
 # Nodes
 
+async def translate_null_query_params(query_params: dict):
+    """Translate null query parameters to None"""
+    translated = query_params.copy()
+    for key, value in query_params.items():
+        if value == 'null':
+            translated[key] = None
+    return translated
+
+
 @app.get('/node/{node_id}', response_model=Union[Regression, Node])
 async def get_node(node_id: str, kind: str = "node"):
     """Get node information from the provided node id"""
--- a/doc/api-details.md
+++ b/doc/api-details.md
@@ -164,6 +164,17 @@ $ curl 'http://localhost:8001/nodes?name=checkout&created__gt=2022-12-06T04:59:0
 
 > **Note** In order to support comparison operators in URL request parameters, models can not contain `__` in the field name.
 
+Nodes with `null` fields can also be retrieved using the endpoint.
+For example, the below command will get all the nodes with `parent` field set to `null`:
+
+```
+$ curl 'http://localhost:8001/nodes?parent=null'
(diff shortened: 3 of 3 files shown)


JenySadadia:
In order to get documents matching provided null fields, translated attributes to `{"$type": 10}` using operator `type` of mongodb with BSON type number(i.e. 10 for null).
Another way is to translate `null` to `None`. That will return documents matching null fields but it will also retrieve documents that do not contain the provided field in case the field doesn't exist in the document.

Reference: https://www.mongodb.com/docs/manual/tutorial/query-for-null-fields/#std-label-faq-developers-query-for-nulls

gctucker (commented):
Have you tried with setting an empty value instead of a magic `null` string?

JenySadadia:
> Have you tried with setting an empty value instead of a magic `null` string?

Actually, I gave it a thought when I started working on this.
The reason I preferred to go with `null` string is, whenever a user queries the `/nodes` endpoint, he will decide on filters for the next query based on the response received from the endpoint.
Currently, the response dictionary has `null` value when fields are not set.
So, at the first glance, any user will think of querying null fields by providing the `null` string. And that is what Alex tried when she wanted to get nodes with `null` parent.
Furthermore, queries with `null` string seem more specific and descriptive to me.

gctucker:
OK, here are a couple of downsides of using `null` I can think of:
* it's a magic string so fields for which `null` is a real value will have a problem
* special translation needs to be done to convert `null` which adds code complexity (so more testing, scope for bugs etc.)

If there's already a way to pass empty values then it's worth thinking twice about adding this support and consider potential alternatives.

JenySadadia:
> OK, here are a couple of downsides of using `null` I can think of:
> 
> * it's a magic string so fields for which `null` is a real value will have a problem
> * special translation needs to be done to convert `null` which adds code complexity (so more testing, scope for bugs etc.)
> 
> If there's already a way to pass empty values then it's worth thinking twice about adding this support and consider potential alternatives.

Okay. Yes, the translation and validation for `null` params add some complexity to the code.
In case we use empty query parameters for this use case, how would we separate queries for null values and empty strings (assuming we have some fields which can have an empty string as a value)?

api/models.py
```diff
@@ -241,7 +241,7 @@ def validate_params(cls, params: dict):
             return False, f"Invalid result value '{result}'"
 
         parent = params.get('parent')
-        if parent:
+        if parent and parent != 'null':
```
gctucker, line 245: I don't get this, why not just translate `null` as `None` when handing the query?  That way we keep the `null` special value as something for the API and the rest of the code doesn't have to change.

api/models.py
```diff
@@ -259,7 +259,7 @@ def translate_fields(cls, params: dict):
         """
         translated = params.copy()
         parent = params.get('parent')
-        if parent:
+        if parent and parent != 'null':
```
gctucker, line 263: Same comment as before, I think `null` should have been translated to `None` before reaching this point.

api/models.py
```diff
@@ -338,8 +339,9 @@ def validate_params(cls, params: dict):
             return False, f"Invalid state value '{state}'"
 
         result = params.get('regression_data.result')
-        if result and result not in [result.value for result in ResultValues]:
-            return False, f"Invalid result value '{result}'"
+        if result and result != 'null':
```
gctucker, line 344: This seems way too intrusive, if we want to make `null` in the query become `None` then I believe we should be able to do this in a single place as an extension to what FastAPI is already doing when parsing the query parameters.

gctucker (commented):
OK I think `param=null`  is probably the best approach here, the drawbacks are pretty narrow as I doubt we'll ever have a test case or a git repository or a test platform called `null`.  However, it seems there are many places in the code where this handled on a case-by-case basis which is not what I was expecting as it should be possible to have a mechanical `null` -> `None` translation at the FastAPI query parsing level.  Doing this everywhere means we're likely to not implement it in some places which will cause some inconsistencies, i.e. are all optional arguments already supporting `null` in this PR? Also it's a lot of manual code churn just to have some basic translation in place.  These symptoms seem to show it's implemented in the wrong layer, this seems more like an API built-in thing than something specific to the models.  It's a bit like the comparison operators and object IDs.

api/models.py
JenySadadia: Yes, translating `null` to `None` would make further validation and translation easier.
I will address this soon.

JenySadadia:
> OK I think `param=null` is probably the best approach here, the drawbacks are pretty narrow as I doubt we'll ever have a test case or a git repository or a test platform called `null`. However, it seems there are many places in the code where this handled on a case-by-case basis which is not what I was expecting as it should be possible to have a mechanical `null` -> `None` translation at the FastAPI query parsing level. Doing this everywhere means we're likely to not implement it in some places which will cause some inconsistencies, i.e. are all optional arguments already supporting `null` in this PR? Also it's a lot of manual code churn just to have some basic translation in place. These symptoms seem to show it's implemented in the wrong layer, this seems more like an API built-in thing than something specific to the models. It's a bit like the comparison operators and object IDs.

Thanks for your feedback.
I have translated `null` request parameters to `None` in the `main` layer.
Also, we need to move methods for query parameter translation and validation from `models` to `main` as it is specific to endpoint request parameters and not models.
That can be a follow-up PR.

JenySadadia:
I have dropped the translation of `None` parameters to `{'$type': 10}` from the DB layer.
The known issue after dropping the translation is that the API will get all the node objects in case the query parameter provided doesn't exist in the model schema.
e.g. `curl http://localhost:8001/nodes?foo=null` will return all the nodes from the DB.

Mongo DB documentation for the same can be found at https://www.mongodb.com/docs/manual/tutorial/query-for-null-fields/#std-label-faq-developers-query-for-nulls

I will update the API documentation and file a GH issue to keep track of the issue.

gctucker:
From the link above, the Mongo DB query will match:

> documents that either contain the item field whose value is null or that do not contain the item field

Presumably, when doing a query with a field that's not in any document (i.e. not part of the schema) and the value is not `null` then it will return nothing?

For example:
```
db.find_by_attributes(node_model, {'banana': None})
```
will return all the nodes, but:
```
db.find_by_attributes(node_model, {'banana': 'green'})
```
will return no nodes.  Is that right?

JenySadadia:
> From the link above, the Mongo DB query will match:
> 
> > documents that either contain the item field whose value is null or that do not contain the item field
> 
> Presumably, when doing a query with a field that's not in any document (i.e. not part of the schema) and the value is not `null` then it will return nothing?
> 
> For example:
> 
> ```
> db.find_by_attributes(node_model, {'banana': None})
> ```
> 
> will return all the nodes, but:
> 
> ```
> db.find_by_attributes(node_model, {'banana': 'green'})
> ```
> 
> will return no nodes. Is that right?

Yes, that is correct.

gctucker:
OK thanks for confirming.  I think it should be fine to just document this behaviour in the API docs for the endpoints that do "fuzzy matching" with an arbitrary set of attributes such as `find` and `count`.

JenySadadia:
> OK thanks for confirming. I think it should be fine to just document this behaviour in the API docs for the endpoints that do "fuzzy matching" with an arbitrary set of attributes such as `find` and `count`.

Yes, I will update the API documentation to state that.

JenySadadia:
Updated the documentation for `nodes` and `count` endpoints.

gctucker (approved):
LGTM as long as the documentation changes looks good on the staging website.

mgalka (approved):
LGTM

Issue #199, opened 2022-11-25
Introduce End-to-End testing of kernelci-api


mgalka:
An enviroment for running End-to-End tests of kernelci-api needs to be created.

End-to-End tests are needed to keep kernelci-api operational and detect failures in a timely manner. Apart from unit tests the project should also have a set of End-to-End tests. 
These tests should setup and run api server as well as a database with pre-populated set of data for each run. After the test runtime environment should be disposed.

- [ ] Create scripts to run testing environment
- [ ] Populate testing environment database with predefined data
- [ ] Run simple test case
- [ ] Prepare scripts to dispose the testing environment

gctucker:
Side note: we might be able to run this in Azure pipeline from GitHub actions.  See for example the actions run by U-Boot PRs:
https://github.com/u-boot/u-boot/pulls

as defined in this file (moar YAML):
https://github.com/u-boot/u-boot/blob/master/.azure-pipelines.yml

JenySadadia:
Here is the proposed solution for introducing E2E tests:
- Add a step for building and running docker services (`docker-compose up --build`) in Github Actions
- Create a script(shell/python) to connect to Database and create a test database (maybe `test_db`)
- Insert test data to the database from the script
- Add a step to Github actions to run the script for DB connection after services are up and running
- Add step/job to run tests in Github Actions
          - We may have separate `yml` files for each test case
          - Python `requests` package can be used to call API endpoints
          - I am not considering using any test framework (Selenium or Jmeter) at the moment
 - Dispose of the test database and disconnect the DB service from the script

gctucker:
> Dispose of the test database and disconnect the DB service from the script

I guess this will be dealt with almost entirely automatically as the GitHub action environment will be garbage-collected just like the Python object holding a connection to the API or database.  But yes, it's a good idea to try and clean things up at the end of the check for completeness.

gctucker:
Based on recent discussions, I think this can be split into tests that only involve `kernelci-api` with some specific client-side service to exercise the API without involving `kernelci-pipeline`.  That way, we can test the API itself in a self-contained way and even only rely on a single `docker-compose` instance to simplify things e.g. networking.

[...]

PR #210, opened 2023-02-01
T40027 Introduce end-to-end testing for API

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/199

Added `test-docker-compose.yaml` to introduce `test` service.
The service will run tests from `e2e_tests` package.
Added below tests:
- Test API root handler
- Test admin user creation and get access token
- Test regular user creation using admin bearer token
- Negative test for regular user creation
- Test for subscribe handler
- Test for unsubscribe handler
- Add a test pipeline for node handler with pubsub event listener
- Test for count handler
- Test for regression handler

Also, added `test.yaml` in github workflows to run tests within a test environment created by `test-docker-compose.yaml`.

### DIFF
--- a/test-docker-compose.yaml
+++ b/test-docker-compose.yaml
@@ -1,0 +1,60 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2023 Collabora Limited
+# Author: Jeny Sadadia <jeny.sadadia@collabora.com>
+
+version: '3'
+services:
+
+  api:
+    container_name: 'test-kernelci-api'
+    build:
+      context: 'docker/api'
+      args:
+        - REQUIREMENTS=${REQUIREMENTS:-requirements.txt}
+    volumes:
+      - './api:/home/kernelci/api'
+    ports:
+      - '${API_HOST_PORT:-8001}:8000'
+    env_file:
+      - '.env'
+
+  db:
+    container_name: 'test-kernelci-api-db'
+    image: 'mongo:5.0'
+
+  redis:
+    container_name: 'test-kernelci-api-redis'
+    image: 'redis:6.2'
+    volumes:
+      - './docker/redis/data:/data'
+
+  storage:
+    container_name: 'test-kernelci-api-storage'
+    image: 'nginx:1.21.3'
+    volumes:
+      - './docker/storage/data:/usr/share/nginx/html'
+    ports:
+      - ${STORAGE_HOST_PORT:-8002}:80
+
+  ssh:
+    container_name: 'test-kernelci-api-ssh'
+    build:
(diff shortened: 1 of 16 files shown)


JenySadadia:
Updated `BASE_URL` for `TestClient` to include version prefix.
Also, updated the request path in all the tests.

test-docker-compose.yaml
```diff
@@ -0,0 +1,67 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2023 Collabora Limited
+# Author: Jeny Sadadia <jeny.sadadia@collabora.com>
+
+version: '3'
+services:
+
+  api:
+    container_name: 'kernelci-api'
```
mgalka, line 10: I think it'd be better to use different container names than the ones used for production to make it clear they're just for testing and can be safely destroyed after running the tests. E.g. you can use something like a `test-` prefix for a name.

mgalka (commented):
Please, have a look at the comment. I also think it may be a good idea to write a helper script that could be used to run e2e tests locally.

test-docker-compose.yaml
JenySadadia: Yes, we need to rename the containers.
Also, I will rename the mongodb volume used for tests to avoid accessing data from `db` service instead.

gctucker:
@mgalka Could you please take another look?

JenySadadia:
> Please, have a look at the comment. I also think it may be a good idea to write a helper script that could be used to run e2e tests locally.

Created script `run_e2e_tests.sh` to run tests locally.
I will update `API Testing` section of `README.md` as well.

JenySadadia:
Hello @mgalka ,
I have covered almost all the major end-to-end tests for API in the PR.
Let me know in case I missed anything.

mgalka:
I think that is good to be merged now.
There is couple of  things that we should think of as a follow-up actions:

- move `requirements-*.txt` files from the `docker` subdirectory to the root. By convention these belong to the main Python project directory and I don't see any particular reason to store them as a part of docker specific files. Apart from that e2e tests should be rather using the application and  testing dependencies rather than having their own requirements.txt file

- Group unit and e2e tests as submodules of the `tests` module instead of keeping them as separate directories in the root project directory. 

```
- tests -
         |_ __init__.py
         |_ unit_tests -
                        |_ __init__.py
                        |_ test_count_handler.py
                        |_...
        |_ e2e_tests -
                        |_ __init__.py
                        |_ test_regression_handler.py
                        |_ ...

mgalka (approved):
I think it's OK to land it, please have a look at my [comment](https://github.com/kernelci/kernelci-api/pull/210#issuecomment-1451798764) though

Issue #212, opened 2023-02-15
Enable GitHub linter annotations


JenySadadia:
Enable Github linter annotations for every push and pull request. Git diffs should get inline annotated with found errors.
Also, to speed up the Actions checks, run linters only on modified files.

- [ ]  Add [Github Action](https://github.com/marketplace/actions/python-lint-annotate-maintain-code-quality-with-annotated-prs-and-pushes) to workflow.
- [ ] Add step to get modified files and send them to the linter

PR #216, opened 2023-02-21
T40660 workflows: enable python-lint-annotate

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/212

Add a job 'lint' using python-lint-annotate to run pylint with GitHub inline annotations.

Tested with https://github.com/kernelci/kernelci-api/pull/217

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,3 +59,23 @@ jobs:
       - name: Run pytest
         run: |
           pytest -v test/
+
+  lint:
+    runs-on: ubuntu-22.04
+    name: Lint
+
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v3
+
+      - name: Run linter annotations
+        uses: marian-code/python-lint-annotate@v3
+        with:
+          python-root-list: "./test/* ./api/*"
+          use-black: false
+          use-flake8: false
+          use-isort: false
+          use-mypy: false
+          use-pycodestyle: true
+          use-pydocstyle: false
+          use-vulture: false
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -85,3 +85,4 @@ jobs:
           use-pycodestyle: true
           use-pydocstyle: false
           use-vulture: false
+          python-version: "3.10"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -47,7 +47,7 @@ jobs:
         run: |
           pylint api.auth
           pylint api.db
-          pylint -d R0903 api.models
+          pylint api.models
           pylint api.pubsub
           pylint test
           pylint api.main
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -75,6 +75,17 @@ jobs:
           echo "CHANGED_FILES=$(echo $(git diff --name-only ${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }} -- | grep \.py))" >> $GITHUB_ENV
 
       - if: env.CHANGED_FILES
+        name: Set up Python
+        uses: actions/setup-python@master
+        with:
+          python-version: "3.10"
+
+      - if: env.CHANGED_FILES
+        name: Install Python packages
+        run: |
+          pip install -r docker/api/requirements-dev.txt
+
+      - if: env.CHANGED_FILES
         uses: marian-code/python-lint-annotate@v3
         with:
           python-root-list: ${{ env.CHANGED_FILES }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -67,11 +67,17 @@ jobs:
     steps:
       - name: Check out source code
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 32  # This is necessary to get the commits
 
-      - name: Run linter annotations
+      - name: Get changed python files between base and head
+        run: >
+          echo "CHANGED_FILES=$(echo $(git diff --name-only ${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }} -- | grep \.py))" >> $GITHUB_ENV
+
+      - if: env.CHANGED_FILES
         uses: marian-code/python-lint-annotate@v3
         with:
-          python-root-list: "./test/* ./api/*"
+          python-root-list: ${{ env.CHANGED_FILES }}
           use-black: false
           use-flake8: false
           use-isort: false
--- a/api/models.py
+++ b/api/models.py
@@ -4,6 +4,10 @@
 # Author: Guillaume Tucker <guillaume.tucker@collabora.com>
 # Author: Jeny Sadadia <jeny.sadadia@collabora.com>
 
+# Disable below flag as some models are just for storing the data and do not
+# need methods
+# pylint: disable=too-few-public-methods
+
 """KernelCI API model definitions"""
 
 from datetime import datetime, timedelta


JenySadadia:
Hello @gctucker 
I am checking why the `lint` job isn't failing even if it finds issues with `models.py`.

.github/workflows/main.yml
```diff
@@ -39,23 +39,42 @@ jobs:
...
-
-      - name: Run pylint
-        run: |
-          pylint api.auth
-          pylint api.db
-          pylint -d R0903 api.models
-          pylint api.pubsub
-          pylint test
-          pylint api.main
-
```
gctucker, line 54: We need to keep this to report errors, the annotations only add annotations.

gctucker (changes requested):

.github/workflows/main.yml
```diff
@@ -39,23 +39,42 @@ jobs:
...
+        with:
+          python-root-list: ${{ env.CHANGED_FILES }}
+          use-black: false
+          use-flake8: false
+          use-isort: false
+          use-mypy: false
+          use-pycodestyle: true
+          use-pydocstyle: false
+          use-vulture: false
+          python-version: "3.10"
```
gctucker, line 80: I believe this should be dropped as the setup-python step already sets it up with a single version.  IIRC this option is only useful when several Python versions are installed.

.github/workflows/main.yml
JenySadadia: That's what I mentioned in the earlier comment. I thought there should be a way to fail `lint` job.

JenySadadia: Anyway, I'll revert this.

gctucker: Yes I had a look at this when enabling it on kernelci-core and I didn't see an obvious way to make the annotation action fail the check if some linters reported any errors.  So for now we need to keep both actions.

.github/workflows/main.yml
JenySadadia: setup-python step explicitly declares python version 3.10 to be used.
But somehow the lint job uses python version 3.8 from actions within `marian-code/python-lint-annotate@v3`.
I added `python-version` to correct it.

gctucker: Ah I guess `ubuntu-22.04` comes with Python 3.8 already installed then.  In this case I guess we need to keep the explicit `python-version` option and I should do the same with `kernelci-core`.

.github/workflows/main.yml
JenySadadia: Oh, okay.

.github/workflows/main.yml
JenySadadia: Yes.

gctucker: Actually, I double checked and `ubuntu-22.04` comes with Python 3.10.  The `marian-code` action runs its own `setup-python` step with 3.8 by default:
```
Run actions/setup-python@v2
  with:
    python-version: 3.8
    architecture: x64
    cache: pip
    token: ***
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.10.10/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.10.10/x64/lib
    CHANGED_FILES: kci kernelci/config/api.py
```
So we can just drop the first `setup-python` step in `main.yml` and rely only on the `python-version` in the annotation action.

JenySadadia: Okay. Thanks for confirming.

JenySadadia: Also, why the value 32 is used in `fetch-depth: 32` in `kernelci-core`?

gctucker: Because we never have PRs with more than 32 commits, in practice.

gctucker: I've added a commit to this PR to use the `python-version` attribute in the annotation action:
https://github.com/kernelci/kernelci-core/pull/1748

gctucker:
I see the linter annotation action is reporting issues with importing `bson` and `pydantic` packages.  It seems like the `requirements.txt` were not installed.  Now I wonder how that's working for `kernelci-core`.

gctucker:
> I see the linter annotation action is reporting issues with importing `bson` and `pydantic` packages. It seems like the `requirements.txt` were not installed. Now I wonder how that's working for `kernelci-core`.

OK I've reworked this accordingly to install `requirements.txt` in the linter annotations action and kept the `setup-python` step too:
https://github.com/kernelci/kernelci-core/pull/1748

JenySadadia:
> I see the linter annotation action is reporting issues with importing `bson` and `pydantic` packages. It seems like the `requirements.txt` were not installed. Now I wonder how that's working for `kernelci-core`.

Okay. I need to fix this here as well.

.github/workflows/main.yml
```diff
@@ -59,3 +59,41 @@ jobs:
...
+
+  lint:
+    runs-on: ubuntu-22.04
+    name: Lint
+
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0  # This is necessary to get the commits
```
gctucker, line 71: The issue with `0` is you get the full history, which can take a lot of time and isn't useful past the commits in the PR.

.github/workflows/main.yml
```diff
@@ -59,3 +59,41 @@ jobs:
...
+      - if: env.CHANGED_FILES
+        name: Set up Python
+        uses: actions/setup-python@master
+        with:
+          python-version: "3.10"
+
+      - if: env.CHANGED_FILES
+        name: Install Python packages
+        run: |
+          pip install -r requirements-dev.txt
```
gctucker, line 86: I think it should be `requirements.txt` rather than the `-dev` one.  The linter tools etc. are installed later.

gctucker (commented):
I think it would be worth creating a temporary draft PR with some linter errors to check they're working as expected.  See what I did for kernelci-core:
https://github.com/kernelci/kernelci-core/pull/1749

If you have other means of checking it's working as expected then fine, but it's hard to tell at the moment.

.github/workflows/main.yml
JenySadadia: Files under `test` are using packages from `requirements-dev.txt`. Hence need to install them as well.

JenySadadia:
> I think it would be worth creating a temporary draft PR with some linter errors to check they're working as expected. See what I did for kernelci-core: [kernelci/kernelci-core#1749](https://github.com/kernelci/kernelci-core/pull/1749)
> 
> If you have other means of checking it's working as expected then fine, but it's hard to tell at the moment.

For the sake of testing, I added a few temporary changes to this PR itself and then discarded it after testing.

gctucker:
> 

OK great.  I can't use that to review the PR though so I'll just have to assume it's working as expected ;)

.github/workflows/main.yml
gctucker: Ah OK, that's a difference with `kernelci-core` where the main `requirements.txt` file has everything to run unit tests as well e.g. `pytest`.  I'm not sure what's the general convention for things that should go in `-dev.txt`.

JenySadadia:
> > 
> 
> OK great. I can't use that to review the PR though so I'll just have to assume it's working as expected ;)

Created https://github.com/kernelci/kernelci-api/pull/217/files for testing.

.github/workflows/main.yml
gctucker: So `requirements.txt` has `pylint==2.12.2` here.  Which version of pylint is used by the annotation action?

gctucker: Interestingly, the annotation action installs `pylint==2.12.1`.

gctucker: Not sure how to debug that.  For now it shouldn't matter as they're very close versions but if for any reason we start running different pylint versions or other linter tools then we could start having different errors being reported which can be inconvenient.

Maybe `requirements-dev.txt` could be split into one file with the unit tests requirements and another one with the linters?  Say `requirements-tests.txt` which would include `requirements.txt` and add things like `pytest` and `fakeredis`, then `requirements-dev.txt` which would include `requirements-tests.txt` and add `pylint` and `pycodestyle` on top.  Then the annotations would be using `requirements-tests.txt`.

JenySadadia: The general convention is requirements for testing and development go to `requirements-dev.txt`. Still, there are no hard rules for that.

.github/workflows/main.yml
JenySadadia: I need to fix this.

.github/workflows/main.yml
JenySadadia: Yes, we need to resolve the issues with version conflicts. 
It is a good idea to split requirements into multiple files.
How about adding `pylint` and `pycodestyle` to `requirements-linter.txt` file? Then the linter requirements file will include `requirements-tests.txt`.

api/models.py
```diff
@@ -4,6 +4,8 @@
 # Author: Guillaume Tucker <guillaume.tucker@collabora.com>
 # Author: Jeny Sadadia <jeny.sadadia@collabora.com>
 
+# pylint: disable=too-few-public-methods
```
gctucker, line 7: It might be good to add a comment about why it's OK to disable this check (basically, because models don't have methods...).

.github/workflows/main.yml
```diff
@@ -67,11 +67,17 @@ jobs:
     steps:
       - name: Check out source code
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 32  # This is necessary to get the commits
```
gctucker, line 71: This is not related to looking for changed files.  Maybe it should just be part of the first commit that adds the linter job?

.github/workflows/main.yml
```diff
@@ -74,6 +74,17 @@ jobs:
         run: >
           echo "CHANGED_FILES=$(echo $(git diff --name-only ${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }} -- | grep \.py))" >> $GITHUB_ENV
 
+      - if: env.CHANGED_FILES
```
gctucker, line 77: I found the commit subject a bit misleading, it looked like some fixes in the Python code to fix import errors reported by pylint.  Maybe something like "install requirements" would be clearer.

gctucker (commented):
I think we can ignore the fact that pylint and pycodestyle are being installed twice for now as it's not causing any issue in practice.  So we should be able to merge this PR with the regular `requirements-dev.txt` and then rework this somehow.

JenySadadia:
> I think we can ignore the fact that pylint and pycodestyle are being installed twice for now as it's not causing any issue in practice. So we should be able to merge this PR with the regular `requirements-dev.txt` and then rework this somehow.

Please see https://github.com/kernelci/kernelci-api/pull/218

gctucker:
> > I think we can ignore the fact that pylint and pycodestyle are being installed twice for now as it's not causing any issue in practice. So we should be able to merge this PR with the regular `requirements-dev.txt` and then rework this somehow.
> 
> Please see #218

Yes, I mean we don't need to depend on #218 to be merged before merging this one.

.github/workflows/main.yml
gctucker: Are you still experimenting with this?

.github/workflows/main.yml
JenySadadia: I think `CHANGED_FILES` will be dependent on how many commits have been fetched.
I am experimenting with it by providing the `fetch-depth: 2`. But it seems like it doesn't change anything.

.github/workflows/main.yml
JenySadadia: Updated the commit message.

.github/workflows/main.yml
gctucker: It won't change anything if the PR has no more than 2 commits on top of the destination branch.  But most PRs have more than that, and rarely more than 20 so that's why I picked 32 as a "round" number.  If some commits are not checked out, then they won't have annotations.

JenySadadia: So, as an experiment, I tried `fetch-depth: 2` and `fetch-depth: 3` in https://github.com/kernelci/kernelci-api/pull/217. Instead of checking out commits based on provided value, it gets all the commits in the checkout and shows annotations for all modified files.

gctucker: OK yes so it is _also_ related to `CHANGED_FILES` as if the depth is less than the number of commits in the PR then the diff revision is not valid.

As we found out today, `--depth` is the number of extra commits to fetch in the history.  So with `--depth=1` you get the current revision and one history commit so 2 commits.

JenySadadia: > As we found out today, `--depth` is the number of extra commits to fetch in the history. So with `--depth=1` you get the current revision and one history commit so 2 commits.

Actually, that is true for `--depth=2`.
Also, I added another linter error to https://github.com/kernelci/kernelci-api/pull/217 to confirm if the annotations are shown based on the provided depth. It is working fine.

JenySadadia:
Could you please take another look at this @gctucker?
I think it is ready to be merged.

.github/workflows/main.yml
gctucker: I'm pretty sure there were 3 commits with `--depth=2` but maybe that's because one was a merge commit.  Either way, I think that's resolved now.

gctucker (approved):

PR #219, opened 2023-03-02
Consolidate e2e test container

gctucker:
Reuse the Docker image from `docker/api` in end-to-end tests rather than keeping a copy in `docker/test` with minor tweaks.

Also use the same `requirements-dev.txt` file for both the `api` and `test` containers to rely on the Docker build cache and avoid doing 2 builds.  I haven't timed it precisely but the GitHub action duration seems to have reduced from 3min to 2min.


docker/api/requirements-dev.txt
```diff
@@ -1,3 +1,8 @@
 -r requirements-tests.txt
 pycodestyle==2.8.0
 pylint==2.12.2
+pytest==6.2.5
+pytest-asyncio==0.16.0
+pytest-dependency==0.5.1
+pytest-order==1.0.1
+httpx==0.23.3
```
JenySadadia, line 8: These should be added to `requirements-tests.txt` as these packages are specific for testing.

test-docker-compose.yaml
```diff
@@ -11,7 +11,7 @@ services:
     build:
       context: 'docker/api'
       args:
-        - REQUIREMENTS=${REQUIREMENTS:-requirements.txt}
+        - REQUIREMENTS=${REQUIREMENTS:-requirements-dev.txt}
```
JenySadadia, line 14: What is the need of changing the requirements file here?

gctucker: It's explained in the commit message.

docker/api/requirements-dev.txt
gctucker: OK

JenySadadia (approved):
LGTM

PR #223, opened 2023-03-08
Update generated API documentation URLs

gctucker:
Since the introduction of versioning, the generated documentation URLs need a prefix e.g. `latest`.  Update the README and static documentation accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Please refer to the [architecture](https://kernelci.org/docs/api/overview/#api-a
 
 ## Getting started with docker-compose
 
-Please refer to [start Docker containers](https://kernelci.org/docs/api/getting-started/#start-docker-compose) of all the services. 
+Please refer to [start Docker containers](https://kernelci.org/docs/api/getting-started/#start-docker-compose) of all the services.
 
 Note that the FastAPI server is running on port 8000 inside the container, but
 it's exposed to the host via port 8001 to avoid conflicts with other services.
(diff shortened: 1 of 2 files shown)


JenySadadia (commented):
As a follow-up PR `doc/api-details` and `doc/getting-started` should also be tweaked to use `latest` prefix.

gctucker:
> As a follow-up PR `doc/api-details` and `doc/getting-started` should also be tweaked to use `latest` prefix.

Why is that?  These pages are not in the generated documentation.  We could have archives of the static documentation for past API releases at some point, but that's not the same thing.

In fact I think it would be better if the generated docs had a way to redirect to `latest` by default.

JenySadadia:
> 

I am not suggesting modifying the page URLs. It's about the content of the pages.
e.g. https://kernelci.org/docs/api/api-details/#create-an-api-token-with-security-scopes this will need to be tweaked to use `/latest/token` endpoint in the sample command.

gctucker:
Ah sorry, I see. Yes that makes sense.

gctucker:
@JenySadadia I've created #228 for this, feel free to pick it up.  Otherwise I can take a look when I'm done with what I'm currently working on.

JenySadadia:
> @JenySadadia I've created #228 for this, feel free to pick it up. Otherwise I can take a look when I'm done with what I'm currently working on.

Thanks. I guess I'll pick it up.

gctucker:
@JenySadadia Is this OK to merge now?

JenySadadia:
> @JenySadadia Is this OK to merge now?

Yes, let me merge it.

JenySadadia (approved):
LGTM

PR #224, opened 2023-03-08
api.main: rename "me" as "whoami"

gctucker:
Rename the /me entry point as /whoami which refers to the standard "whoami" tool to retrieve information about the current user.  This is a better description of what the purpose of the endpoint and would avoid ambiguity with other endpoints related to the current user such as profile information etc.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -177,9 +177,9 @@ async def login_for_access_token(
     return {"access_token": access_token, "token_type": "bearer"}
 
 
-@app.get('/me', response_model=User)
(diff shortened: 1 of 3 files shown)


gctucker (commented):
There's a pylint issue but it's not related to this PR:
`Unable to import 'e2e_tests.conftest'` in `test_user_creation.py`

JenySadadia:
> 

Thanks for pointing that out.
We haven't enabled `pylint` for e2e tests yet. I think it is worth creating an issue to keep track of it.

PR #225, opened 2023-03-09
api.main: drop unused get_root_node endpoint

gctucker:
The get_root_node endpoint is not used anywhere in the current pipeline and it's not clear whether it will be useful or not.  So drop this now to simplify the API and avoid having to maintain unused code.

Depends on https://github.com/kernelci/kernelci-core/pull/1809


JenySadadia:
In the former implementation of the pipeline, the test report service was fetching `checkout` node using `get_root_node` based on the completed test node.
Since the implementation has been changed I guess it is OK to drop it.

Bdw this needs to be dropped first https://github.com/kernelci/kernelci-core/blob/main/kernelci/db/kernelci_api.py#L170

gctucker:
> Bdw this needs to be dropped first https://github.com/kernelci/kernelci-core/blob/main/kernelci/db/kernelci_api.py#L170

Actually the whole file is going to get dropped as there's now `kernelci.api`, that's how I realised this endpoint wasn't being used.  See #1789

JenySadadia (approved):
LGTM. Let's merge it after we drop file `kernelci.db.kernelci_api`.

PR #229, opened 2023-03-20
T41382 Fix `create_admin_user` script with API version

JenySadadia:
As a part of API versioning implementation, append API version `latest` suffix to default
URL used for requesting API endpoints to create an admin user. Adjust URL path for `/hash` endpoint accordingly for correct URL parsing.

### DIFF
--- a/create_admin_user
+++ b/create_admin_user
@@ -18,7 +18,7 @@ import sys
 import urllib
 
 
-DEFAULT_URL = 'http://localhost:8001'
+DEFAULT_URL = 'http://localhost:8001/latest/'
 
 
 def get_hashed_password(password, url):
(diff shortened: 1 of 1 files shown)


gctucker:
How about an admin user creating other admins with `kci user` though?

JenySadadia:
> How about an admin user creating other admins with `kci user` though?

What is the requirement for other admins?
I don't think the provision for creating admins with `kci user` is required atm.

gctucker:
I think it would be good at some point to be able to do this:
```
kci user add <name> --scope admin
```
which should be allowed if the current user is already admin or to be precise, if the user token has the right permission scope.  So the tools that come with `kernelci-api` would only really be needed to bootstrap this by creating the first admin user.

gctucker (approved):

Issue #228, opened 2023-03-14
Update documentation with `/latest` prefix in URLs


gctucker:
The API documentation provides some sample `curl` commands which use raw URLs to access the API.  Now that the API is versioned, the version needs to be added to the URL e.g. `http://localhost:8001/latest/token` with `latest` to always use the current development version.

PR #231, opened 2023-03-22
T41423 Update documentation with `/latest` prefix in URLs

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/228

Since the implementation of versioned APIs, all the API endpoints have API version prefix i.e. `latest` for the current development version.
 Fix all the URLs of sample curl commands with API version in `doc/getting-started` and `doc/api-details` pages.

### DIFF
--- a/doc/api-details.md
+++ b/doc/api-details.md
@@ -59,7 +59,7 @@ provided with white space separation. For example:
 
 ```
 $ curl -X 'POST' \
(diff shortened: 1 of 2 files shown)


gctucker (approved):
Verified on https://static.staging.kernelci.org/docs/api/api-details/

Issue #230, opened 2023-03-21
Enable autoindex in nginx


gctucker:
At the moment, the default image is being used which doesn't enable autoindex.  As such, visiting http://localhost:8002/ locally returns HTTP 403 Forbidden.  Having a listing of the available files and directories like on https://storage.kernelci.org/ would be useful to easily go through the uploaded artifacts.

This can be done by providing a custom configuration file.  See the reference documentation for the Nginx Docker image here:

https://hub.docker.com/_/nginx

and some sample way of adjusting the configuration file here:

https://github.com/dceoy/docker-nginx-autoindex/blob/master/Dockerfile

Rather than using a non-standard Docker image or adding a special Dockerfile, keeping an Nginx config file under source control and installing it in the container via `docker-compose.yaml` would seem like the best way forward to be able to easily update Nginx versions using the official images.

PR #232, opened 2023-03-23
T41424 Enable autoindex in nginx

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/230 

Created directory `config`  under `docker.storage` to store configuration file for `storage` service.
Added `nginx.conf` file to modify `server` block copied from `/etc/nginx/conf.d/default.conf` to enable autoindex.
Mounted the config file in `docker-compose`.

### DIFF
--- a/docker/storage/config/nginx.conf
+++ b/docker/storage/config/nginx.conf
@@ -1,0 +1,45 @@
+server {
+    listen       80;
+    listen  [::]:80;
+    server_name  localhost;
+
+    #access_log  /var/log/nginx/host.access.log  main;
+
+    location / {
+        autoindex on;
+        root   /usr/share/nginx/html;
+        index  index.html index.htm;
+    }
+
+    #error_page  404              /404.html;
+
+    # redirect server error pages to the static page /50x.html
+    #
+    error_page   500 502 503 504  /50x.html;
+    location = /50x.html {
+        root   /usr/share/nginx/html;
+    }
+
+    # proxy the PHP scripts to Apache listening on 127.0.0.1:80
+    #
+    #location ~ \.php$ {
+    #    proxy_pass   http://127.0.0.1;
+    #}
+
+    # pass the PHP scripts to FastCGI server listening on 127.0.0.1:9000
+    #
+    #location ~ \.php$ {
+    #    root           html;
+    #    fastcgi_pass   127.0.0.1:9000;
+    #    fastcgi_index  index.php;
+    #    fastcgi_param  SCRIPT_FILENAME  /scripts$fastcgi_script_name;
+    #    include        fastcgi_params;
+    #}
+
+    # deny access to .htaccess files, if Apache's document root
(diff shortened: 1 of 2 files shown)


gctucker:
Tested OK on staging: http://staging.kernelci.org:9080/

gctucker:
Note: Browsers might try to automatically redirect to https://staging.kernelci.org:9080/ which won't work as there's no SSL certificate set up for this.  You can work around this by opening it in an incognito window or by adding an exception in the browser settings.

gctucker:
(also `curl http://staging.kernelci.org:9080/` works...)

gctucker:
I've moved the API storage directory to `/data/storage.staging.kernelci.org` on the server and bind-mounted it so it's now available both via port 9080 and also here (which works with HTTPS):
https://storage.staging.kernelci.org/api/

gctucker:
@mgalka @VinceHillier 

I've added this line in `fstab` so it will mount it again upon reboot:
```
/data/storage.staging.kernelci.org/api		/home/kernelci/kernelci-deploy/checkout/kernelci-api/docker/storage/data	none	bind
```
Please note that `README.txt` is under source control, so if that gets changed in the `storage.kernelci.org` directory it will appear in the checkout.  So it shouldn't get garbage-collected.

gctucker (approved):

docker/storage/config/nginx.conf
```diff
@@ -0,0 +1,45 @@
...
+    #
+    #location ~ \.php$ {
+    #    root           html;
+    #    fastcgi_pass   127.0.0.1:9000;
+    #    fastcgi_index  index.php;
+    #    fastcgi_param  SCRIPT_FILENAME  /scripts$fastcgi_script_name;
+    #    include        fastcgi_params;
+    #}
+
+    # deny access to .htaccess files, if Apache's document root
```
gctucker, line 39: So actually, we could maybe just have added `.htaccess` with the `autoindex` option if Nginx supports it like Apache.  But a config file is probably cleaner anyway.

JenySadadia:
Thanks for the testing @gctucker 
Adding `server` block within `http` context to the default `/etc/nginx/nginx.conf` file was another way of enabling `autoindex`.
However, the config file for nginx is subject to the change with different nginx version.
Hence, I decided to add custom configuration to `/etc/nginx/conf.d/default.conf` file.

docker/storage/config/nginx.conf
gctucker, line 1: Ah I missed the fact this was being copied as `default.conf`.  Maybe this file should be renamed to `default.conf` to remove ambiguity as it's not actually `nginx.conf`?

gctucker: Maybe just make another commit / PR to fix it with a rename?

JenySadadia: Created https://github.com/kernelci/kernelci-api/pull/233

PR #233, opened 2023-03-23
T41424  docker/storage/config: rename nginx config file

JenySadadia:
Rename `nginx.conf` to `default.conf` to avoid the ambiguity with the default nginx configuration file name which resides at `/etc/nginx/nginx.conf` in the nginx container. The mounted custom configuration file typically resides at `/etc/nginx/conf.d/default.conf` in the container. Update `docker-compose` file to change mount
for `storage` service accordingly.


gctucker (approved):
Tested OK on staging.

PR #236, opened 2023-03-31
Fix URLs for `openapi.json`

gctucker:
Add the missing `latest/` API version prefix to URLs for the `openapi.json` file.  The URLs without the prefix return a top-level JSON file which is not a description of the actual API.


JenySadadia:
Tested OK on staging.

PR #238, opened 2023-04-06
api.db: use ObjectId directly in .find_one()

gctucker:
The .find_one() collection method can use an ObjectId directly, which is simpler than providing a dictionary filter using '_id'.  Update all calls to .find_one() accordingly.


JenySadadia (approved):
Tested OK on staging.

PR #239, opened 2023-04-06
Rework parameters translation and add support for integer values

gctucker:
Simplify the implementation for translating parameter values using generators and add steps to convert decimal strings to integers so they can be matched and compared with integers in the database.


gctucker:
OK this works, but there are a couple of things to consider.  We could rely on the type of the model field to cast values automatically, but still I think in some cases it makes sense to have an explicit cast in particular for arbitrary data in Node.data.  Then if we keep the `__int` operator we also probably should add `__str` and maybe other primitive types.  We should also add unit tests for the operators.  I'll create some GitHub issues to keep track of that.

gctucker:
One last thing, we can't currently combine comparison operators with `__int`.  So we can't compare strings with a value that has only digits as that's converted to a number automatically now.  I guess in most cases we'll need numeric comparisons with numeric values, so it's kind of a special corner-case.  Also only decimal numbers are currently cast automatically - it should be possible to use hexadecimal etc. with something like `foobar=0x1234abcd__int` in principle but I haven't tried it.  Probably something to add to the unit tests once we have them.

gctucker:
See #250 and #251 about `__str` and unit tests.

PR #241, opened 2023-04-06
Add `Node.data` and basic `KernelBuild` model

gctucker:
Add `Node.data` field with a basic `KernelBuild` model for nodes of the `kbuild` kind.


gctucker:
After discussing this with others, it would seem better to keep the data validation on the client side and just have a generic dictionary in the Node model. The approach of relying on Node.kind to do an extra validation step on the API side has some downsides: not included in the generated OpenAPI schema and documentation, tied to pipeline jobs so subject to change with no justification for an API update...  Maybe some generic schema will become more apparent and suitable for API later.

gctucker:
Replaced by #247 and #248

PR #242, opened 2023-04-11
api.models: drop Node.validate_{state,result} for Enum fields

gctucker:
Drop the custom validation methods for Enum fields Node.state and Node.result as Pydantic is already doing that.  If the value in the input data isn't valid for an Enum field then it will raise an error like for any other invalid field data.

Fixes: 513cb0a20fe9 ("api: Improve validation for request query params")

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -242,22 +242,8 @@ class Node(DatabaseModel):
         self.updated = datetime.utcnow()
 
     @classmethod
-    def validate_state(cls, state):
-        """Validate Node.state"""
-        if state and state not in [state.value for state in StateValues]:
-            raise ValueError(f"Invalid state value '{state}'")
-
-    @classmethod
-    def validate_result(cls, result):
-        """Validate Node.result"""
-        if result and result not in [result.value for result in ResultValues]:
-            raise ValueError(f"Invalid result value '{result}'")
-
-    @classmethod
     def validate_params(cls, params: dict):
         """Validate Node parameters"""
-        Node.validate_state(params.get('state'))
-        Node.validate_result(params.get('result'))
         parent = params.get('parent')
         if parent:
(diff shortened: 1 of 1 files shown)


api/models.py
```diff
@@ -240,23 +240,9 @@ class Node(DatabaseModel):
...
-        """Validate Node.state"""
-        if state and state not in [state.value for state in StateValues]:
-            raise ValueError(f"Invalid state value '{state}'")
-
-    @classmethod
-    def validate_result(cls, result):
-        """Validate Node.result"""
-        if result and result not in [result.value for result in ResultValues]:
-            raise ValueError(f"Invalid result value '{result}'")
-
```
JenySadadia, line 254: These methods are validating GET request query parameters for `/nodes`. It is not for the validation of POST node requests.

```
curl http://localhost:8001/latest/nodes?state=don
{"detail":"Invalid state value 'don'"}
```
We won't get any intimation of an invalid value without the validation:
```
$ curl http://localhost:8001/latest/nodes?state=don
{"items":[],"total":0,"limit":50,"offset":0}
```

JenySadadia: I guess it is fine to drop these validations as it will work as expected if the value does not match with provided `state` or `result`.

gctucker: Ah OK, so I might need to tweak the commit message.

gctucker: Also there's nothing in the model class that shows it's only used for GET queries.  Then I wonder if we also need the `PyObjectId` validation.  Well I made this change initially as a side-effect of a bug I found with the default value of `Node.state` which had the `Enum` object instance rather than a plain string.  I'll go through this validation code again to double check.

gctucker:
Replaced with #242 which drops all the custom parameter validation code as it's not necessary.

PR #243, opened 2023-04-11
api.models: use (str, Enum) as mixin for string enums

gctucker:
Use (str, Enum) as mixin parent classes for Enum fields that are only strings.  This helps ensure that only strings are accepted as values for these Enum types.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -43,7 +43,7 @@ class PyObjectId(ObjectId):
         return ObjectId(value)
 
 
-class StateValues(enum.Enum):
+class StateValues(str, enum.Enum):
     """Enumeration to declare values to be used for Node.state"""
 
     RUNNING = 'running'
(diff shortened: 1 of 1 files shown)


api/models.py
```diff
@@ -43,7 +43,7 @@ def validate(cls, value):
         return ObjectId(value)
 
 
-class StateValues(enum.Enum):
+class StateValues(str, enum.Enum):
```
JenySadadia, line 46: How about directly using `StrEnum`?
https://docs.python.org/3/library/enum.html#enum.StrEnum

gctucker: > New in version 3.11

We're still on 3.10, but yes I guess that's something we could do when we upgrade.

JenySadadia: OK

JenySadadia (approved):
LGTM

PR #245, opened 2023-04-11
Add extra fields in the Node event data

gctucker:
Consolidate the code used to send events when Node objects are being created or updated, and add some extra fields to avoid always fetching the whole Node object.  This should provide enough information for a pipeline to determine which jobs should be run next based on the event itself.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -196,6 +196,12 @@ def _get_node_event_data(operation, node):
     return {
         'op': operation,
         'id': str(node.id),
+        'name': node.name,
+        'path': node.path,
+        'group': node.group,
+        'state': node.state,
+        'result': node.result,
+        'revision': node.revision.dict(),
     }
 
 
--- a/api/main.py
+++ b/api/main.py
@@ -192,6 +192,13 @@ def get_password_hash(password: Password):
 # -----------------------------------------------------------------------------
 # Nodes
 
+def _get_node_event_data(operation, node):
+    return {
+        'op': operation,
+        'id': str(node.id),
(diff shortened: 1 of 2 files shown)


api/main.py
```diff
@@ -196,6 +196,11 @@ def _get_node_event_data(operation, node):
     return {
         'op': operation,
         'id': str(node.id),
+        'name': node.name,
+        'path': node.path,
+        'group': node.group,
```
JenySadadia, line 201: Atm no service is using `path` or `group` to filter events.
Maybe we can add them to the event data when they are actually required?

gctucker: I think it's very likely they'll get used as part of some pipeline job conditions and they're very generic.  Typically, some jobs would have the condition that they get run when a particular test result is passing (or failing).  I'm also trying to minimise the number of changes in this area, so my judgment was it was less likely to be removed later than added very soon.

JenySadadia: Okay

api/models.py
```diff
@@ -163,7 +163,24 @@ class Revision(BaseModel):
...
+        description="Name of the base defconfig file used for the build"
+    )
+    fragments: list = Field(
+        description="List of the config fragment names for the build"
+    )
+    compiler: str = Field(
+        description="Name of the compiler used for the build"
+    )
+
+
```
JenySadadia, line 181: I guess you are going to drop this. Right?

gctucker: Yes it just needs a rebase

PR #247, opened 2023-04-17
api.models: add Node.data field for arbitrary data

gctucker:
Add a Node.data optional dictionary field to contain arbitrary data associated with the node.  This makes it possible for client code to handle data that doesn't require a schema definition in the API, for example any meta-data associated with particular test jobs.

Update unit tests and e2e tests accordingly.


JenySadadia (approved):
LGTM

JenySadadia:
Tested OK on staging.

PR #248, opened 2023-04-17
api.models: allow Node.kind to take other values than 'node'

gctucker:
Remove const=True from Node.kind to allow it to have other values than the default 'node'.  This is to be able to deduce if Node.data should be validated against a particular model, or to know its type in general.

The original assumption was that we would have derivative classes from Node but this isn't too practical as the base class can't be used with Pydantic with derivated objects since it can't validate the extra fields.  So instead, Node can become a more general class rather than having specialised derivative ones.  The Node.data part can be used for specialisation instead.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -192,8 +192,7 @@ class Node(DatabaseModel):
     """KernelCI primitive node object model for generic test results"""
     kind: str = Field(
         default='node',
-        description='Type of the object',
-        const=True
+        description='Type of the object'
     )
     name: str = Field(
         description='Name of the node object'


JenySadadia (approved):

JenySadadia:
A follow-up can be to modify `Regression` model based on this change.

The below lines can be dropped from `Regression` model now:
```
    kind: str = Field(
        default='regression',
        description='Type of the object',
        const=True
    )
```
And default `kind` value can be added to `post_regression` handler.

JenySadadia:
Tested OK on staging with `kind=kbuild` nodes.

gctucker:
> Tested OK on staging with `kind=kbuild` nodes.

Actually we'll stop using nodes with the kind set to `kbuild` but that's kind of a detail.  I guess we might revert this commit later if the kind has to be fixed in the models rather than provided by the client code.

PR #253, opened 2023-04-20
api.models: drop Node.validate_params()

gctucker:
Drop the Node.validate_params() custom validation method used with GET queries as invalid values will not match any result and this is used with fuzzy-matching queries that don't validate the input using Pydantic.  Also, values of the wrong type are already reported as errors by the lower layers.  For example, with an invalid ObjectId:

    $ ./kci node find parent=xyz --limit=1
    400 Client Error: Bad Request for url: http://172.17.0.1:8001/latest/nodes?parent=xyz&limit=1
    'xyz' is not a valid ObjectId, it must be a 12-byte input or a 24-character hex string

Other queries such as the POST and PUT ones rely directly on Pydantic to validate the input data against the model so there's no need for custom validation there either (and .validate_params() wasn't used in these cases, only with fuzzy-matching GET queries to find or count nodes).

Fixes: 513cb0a20fe9 ("api: Improve validation for request query params")

Replaces #242

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -241,7 +241,6 @@ async def get_nodes(request: Request, kind: str = "node"):
 
     try:
         model = get_model_from_kind(kind)
-        model.validate_params(query_params)
         translated_params = model.translate_fields(query_params)
         return await db.find_by_attributes(model, translated_params)
     except KeyError as error:
@@ -263,7 +262,6 @@ async def get_nodes_count(request: Request, kind: str = "node"):
 
     try:
         model = get_model_from_kind(kind)
-        model.validate_params(query_params)
         translated_params = model.translate_fields(query_params)
         return await db.count(model, translated_params)
     except KeyError as error:
--- a/api/models.py
+++ b/api/models.py
@@ -242,27 +242,6 @@ class Node(DatabaseModel):
         self.updated = datetime.utcnow()
 
     @classmethod
-    def validate_state(cls, state):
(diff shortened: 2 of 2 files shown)


JenySadadia (commented):
The `validate_params` method was originally added to raise `ValueError` from API upon invalid request params and catch the exception from pipeline services to keep them in running state even after some error has occurred.
However, the pipeline is not catching `ValueError` anymore and exiting even with param validation:
```
kernelci-pipeline-tarball | requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: http://172.17.0.1:8001/latest/nodes?parent=available&limit=100&offset=0
kernelci-pipeline-tarball exited with code 1
```
I guess we can merge this change as long as we want services to exit on HTTP exceptions.

JenySadadia:
> Yes I think error handling in client side is a separate issue, that should also get addressed as a follow up. I'd like to ultimately remove the tarball service to have it implemented as a job. I think the runner service is already dealing with HTTP errors.

I took tarball as just an example. I guess all the services need to handle HTTP errors for `get_nodes` now.

gctucker:
Well all clients need to handle HTTP errors anyway, that's unrelated to the API I think.  From the API's point of view, it should just return the correct HTTP depending on the query and we should try to avoid HTTP 5xx.

JenySadadia (approved):

Issue #246, opened 2023-04-14
Field aliases are ignored in paginated data


gctucker:
When returning a Pydantic object from the API, for example a `Node`, the JSON data has an `_id` field which is the internal field name used by MongoDB.  The Pydantic model has an `id` field with an alias for `_id` so that the object has `Node.id`.  So in principle, users should only see `id` in the JSON data sent or received to/from the API and not `_id`.

For simple cases like getting a Node by id, this can be done:
```python
@app.get('/node/{node_id}', response_model=Union[Regression, Node],
         response_model_by_alias=False)
async def get_node(node_id: str, kind: str = "node"):
```
Then the returned data will have `id` and not `_id` which is the alias.  This option `response_model_by_alias=False` works fine and we can just use it.  However, when doing pagination such as in this case, the option doesn't apply to the paginated data since it's one layer further down the stack:
```python
@app.get('/nodes', response_model=PageModel, response_model_by_alias=False)
async def get_nodes(request: Request, kind: str = "node"):
```
FastAPI would just not use aliases in `PageModel`, but that's not what we want.  We want to not use aliases in the objects inside the page object.  I'm not sure if there is an option to do that in `fastapi_pagination`, so this needs some investigation.

It's also worth noting that the generated documentation and OpenAPI schema also appears to be using the alias:
https://staging.kernelci.org:9000/latest/docs#/default/get_node_node__node_id__get
with `_id` in Node objects.

What's strange is that both `Node.dict()` and `Node.json()` return data structures that don't use the alias but have `id`.

Essentially, it would be good if `_id` was only used at the MongoDB layer to remain consistent with all uses of the models.

gctucker:
@JenySadadia Maybe create a follow-up issue about investigating whether this could be fixed in the `fastapi-paginator` package?

PR #254, opened 2023-04-21
Serialize response without field alias

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/246

The base model `ModelId` has an alias `_id` for `id` field. Instead of returning response with `id` field upon various requests (e.g. POST node), various endpoint handlers return `_id` in the response dictionary. Change that default behavior by setting `response_model_by_alias=False` in the endpoint handlers for POST nodes, PUT nodes, and GET node by id.

We need to serialize models to generate responses without using alias for GET `/nodes` requests. In usual cases providing `response_model_by_alias` to the endpoint definition serves the purpose. However, that doesn't work in the case of paginated data. Hence, need to serialize it manually.


api/models.py
```diff
@@ -79,6 +80,23 @@ class Config:
...
+        This is required to get models with `id` field in the response
+        instead of `_id`.
+        In usual cases providing `response_model_by_alias` to endpoint
+        definition serves the purpose. However, that doesn't work in case
+        of paginated data. Hence, need to serialize it manually.
+        """
+        serialized_data = []
+        for obj in data:
+            serialized_data.append(
+                jsonable_encoder(cls(**obj), by_alias=False)
```
gctucker, line 96: How about `obj.dict()` or `obj.json()`?

gctucker: Also I think this should probably be outside of `api.models` so we don't add a dependency on `fastapi`.  Maybe it could be in `main` instead as that's where FastAPI is actually used.

gctucker:
I wonder if we could add an option in `fastapi_pagination` do not use aliases recursively in the paginated data, like the standard option provided by FastAPI when returning objects directly.

JenySadadia:
> I wonder if we could add an option in `fastapi_pagination` do not use aliases recursively in the paginated data, like the standard option provided by FastAPI when returning objects directly.

I investigated it thoroughly. Pydantic has not enabled support for adding an option to `Config` class for this case.
Also, `fastapi-pagination` does not provide it either.

gctucker:
> > I wonder if we could add an option in `fastapi_pagination` do not use aliases recursively in the paginated data, like the standard option provided by FastAPI when returning objects directly.
> 
> I investigated it thoroughly. Pydantic has not enabled support for adding an option to `Config` class for this case. Also, `fastapi-pagination` does not provide it either.

Yes that's what I mean, I think it might be a good idea to add support for it in `fastapi-pagination`.

api/models.py
gctucker: Why would you prefer to do what FastAPI is doing in its underlying implementation rather than the method provided by the object?

Actually I guess it should be something like this:
```python
for kwargs in data:
    obj = cls(**kwargs)
    serialised_data.append(obj.dict(by_alias=False))
```
The pagination logic isn't really class-specific so this could be a standalone function with a model argument to know the class to call to make the objects, assuming all the objects are of the same type.

JenySadadia:
> > > I wonder if we could add an option in `fastapi_pagination` do not use aliases recursively in the paginated data, like the standard option provided by FastAPI when returning objects directly.
> > 
> > 
> > I investigated it thoroughly. Pydantic has not enabled support for adding an option to `Config` class for this case. Also, `fastapi-pagination` does not provide it either.
> 
> Yes that's what I mean, I think it might be a good idea to add support for it in `fastapi-pagination`.

Yes, maybe I can contribute to the package.

gctucker:
I guess it's useful to have a solution in our code first to fix the issue, and then if it becomes an option in the pagination package then we could simplify things on our side and just use the option.  We could also use a fork of the package to test the changes on staging temporarily, but we can't really keep using downstream changes so it will need to be accepted upstream eventually or we'll just have to stick to our local solution.

JenySadadia:
> I guess it's useful to have a solution in our code first to fix the issue, and then if it becomes an option in the pagination package then we could simplify things on our side and just use the option. We could also use a fork of the package to test the changes on staging temporarily, but we can't really keep using downstream changes so it will need to be accepted upstream eventually or we'll just have to stick to our local solution.

Yes, let's implement it locally first.

JenySadadia:
I'll fix test failures first and then we need to update client services in pipeline and API bindings in `kernelci-core` as well.

api/models.py
JenySadadia: > Why would you prefer to do what FastAPI is doing in its underlying implementation rather than the method provided by the object?
> 
> Actually I guess it should be something like this:
> 
> ```python
> for kwargs in data:
>     obj = cls(**kwargs)
>     serialised_data.append(obj.dict(by_alias=False))
> ```
> 
> The pagination logic isn't really class-specific so this could be a standalone function with a model argument to know the class to call to make the objects, assuming all the objects are of the same type.

Yes, I agree that we should reduce dependency.
I dropped the encoder from fastapi.
It turns out that `cls(**obj).dict()` works fine as by default `by_alias` flag is set to `False`.

JenySadadia: > 

I think the serialization should be specific to models.
It's possible that the various models will require serialization differently based on their fields.

gctucker: It's one thing to add the possibility to override how serialisation is done in each model, and another one to add a standard option for not using aliases which is what we're doing here.

JenySadadia: OK. Updated the code.

gctucker (approved):

gctucker:
This seems to be working fine on staging although I don't have particular results with these changes.  The other changes in kernelci-core need to be merged too then we can double check everything is working as expected with the main branches locally.

JenySadadia:
> This seems to be working fine on staging although I don't have particular results with these changes. The other changes in kernelci-core need to be merged too then we can double check everything is working as expected with the main branches locally.

We need to merge this one too https://github.com/kernelci/kernelci-pipeline/pull/255

PR #265, opened 2023-06-12
Add support for user groups

JenySadadia:
Add support for user groups.
- Added model `UserGroup`
- Added a python script to create default groups i.e. admin and users
- Added POST handler for it
- Added GET handler for it
- Added e2e test for user group creation

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 #
-# Copyright (C) 2021, 2022 Collabora Limited
+# Copyright (C) 2021-2023 Collabora Limited
 # Author: Guillaume Tucker <guillaume.tucker@collabora.com>
 # Author: Jeny Sadadia <jeny.sadadia@collabora.com>
 
@@ -33,6 +33,7 @@ from .models import (
     Hierarchy,
     Regression,
     User,
+    UserGroup,
     Password,
     get_model_from_kind
 )
@@ -135,6 +136,24 @@ async def post_user(
         ) from error
     await pubsub.publish_cloudevent('user', {'op': operation,
                                              'id': str(obj.id)})
+    return obj
+
+
+@app.post('/group', response_model=UserGroup, response_model_by_alias=False)
+async def post_user_group(
+        group: UserGroup,
+        current_user: User = Security(get_user, scopes=["admin"])):
+    """Create new user group"""
+    try:
+        obj = await db.create(group)
+    except DuplicateKeyError as error:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=f"Group '{group.name}' already exists. \
+Use a different group name."
+        ) from error
+    await pubsub.publish_cloudevent('user_group', {'op': 'created',
+                                                   'id': str(obj.id)})
     return obj
 
 
--- a/api/main.py
+++ b/api/main.py
@@ -168,6 +168,13 @@ async def get_user_groups(request: Request):
     return paginated_resp
 
 
+@app.get('/group/{group_id}', response_model=UserGroup,
+         response_model_by_alias=False)
+async def get_group(group_id: str):
+    """Get user group information from the provided group id"""
+    return await db.find_by_id(UserGroup, group_id)
+
+
 @app.get('/')
 async def root():
     """Root endpoint handler"""
--- a/api/main.py
+++ b/api/main.py
@@ -157,6 +157,17 @@ Use a different group name."
     return obj
 
 
+@app.get('/groups', response_model=PageModel)
+async def get_user_groups(request: Request):
+    """Get all the user groups if no request parameters have passed.
+       Get all the matching user groups otherwise."""
+    query_params = dict(request.query_params)
+    paginated_resp = await db.find_by_attributes(UserGroup, query_params)
+    paginated_resp.items = serialize_paginated_data(
+        UserGroup, paginated_resp.items)
+    return paginated_resp
+
+
 @app.get('/')
 async def root():
     """Root endpoint handler"""
--- a/init_user_groups
+++ b/init_user_groups
@@ -1,0 +1,31 @@
+#!/usr/bin/env python3
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2023 Collabora Limited
+# Author: Jeny Sadadia <jeny.sadadia@collabora.com>
+
+"""Command line utility for creating default groups"""
+
+import subprocess
+
+
+DEFAULT_GROUPS = ['admin']
+
+def create_group(group_name):
+    """
+    Create user group
+    """
+    cmd = f"docker-compose exec db /bin/mongo kernelci --eval \
+\"db.usergroup.insert({{name: '{group_name}'}})\""
+    ret_code = subprocess.call(cmd, shell=True)
+    return ret_code == 0
+
+
+def main():
+    for group in DEFAULT_GROUPS:
+        if not create_group(group):
+            print(f"Failed to execute command for adding group: {group}")
+
+if __name__ == "__main__":
+    main()
--- a/api/db.py
+++ b/api/db.py
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 #
-# Copyright (C) 2021, 2022 Collabora Limited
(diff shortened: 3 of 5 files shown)


JenySadadia:
Unit tests for `/user_group` handler can be a follow-up PR.

api/main.py
```diff
@@ -138,6 +139,32 @@ async def post_user(
...
+        group_name: str,
+        current_user: User = Security(get_user, scopes=["admin"])):
+    """Create new user group"""
+    try:
+        obj = await db.create(UserGroup(user_group=group_name))
+        operation = 'created'
+    except DuplicateKeyError as error:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=f"{group_name} already exists. Use different group name."
```
gctucker, line 153:
```suggestion
            detail=f"{group_name} already exists. Use a different group name."
```

api/main.py
```diff
@@ -138,6 +139,32 @@ async def post_user(
...
 
 
+@app.post('/user_group/{group_name}', response_model=UserGroup)
+async def post_user_group(
+        group_name: str,
+        current_user: User = Security(get_user, scopes=["admin"])):
+    """Create new user group"""
+    try:
+        obj = await db.create(UserGroup(user_group=group_name))
+        operation = 'created'
```
gctucker, line 149: I don't think we need this variable, it's only used once when calling `publish_cloudevent()`.

init_user_groups
```diff
@@ -0,0 +1,32 @@
...
+#
+# Copyright (C) 2023 Collabora Limited
+# Author: Jeny Sadadia <jeny.sadadia@collabora.com>
+
+"""Command line utility for creating default user groups"""
+
+import subprocess
+
+
+DEFAULT_GROUPS = ['admin', 'users']
```
gctucker, line 13: If users can belong to several groups, then I think we don't need the `users` group.

api/main.py
```diff
@@ -138,6 +139,32 @@ async def post_user(
...
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=f"{group_name} already exists. Use different group name."
+        ) from error
+    await pubsub.publish_cloudevent('user_group', {'op': operation,
+                                                   'id': str(obj.id)})
+    return obj
+
+
+@app.get('/user_group', response_model=PageModel)
```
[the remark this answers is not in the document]
JenySadadia: Renamed it to `/group`.
Also added `/group/{group_id}` to get group by ID.

init_user_groups
JenySadadia: As per our discussion yesterday, it makes sense to just have "admin" as a default group and discard "users" one. That way we can prevent all the regular users from accessing each other's data.

api/main.py
```diff
@@ -138,6 +139,41 @@ async def post_user(
     return obj
 
 
+@app.post('/group/{group_name}', response_model=UserGroup)
```
gctucker, line 142: One thing, maybe we don't need `group_name` in the URL?  This could be sent as part of the POST data.  I think having the group name in POST and a group ID in GET is not ideal as it creates some asymmetry between the query methods.

api/main.py
```diff
@@ -138,6 +139,41 @@ async def post_user(
...
+async def post_user_group(
+        group_name: str,
+        current_user: User = Security(get_user, scopes=["admin"])):
+    """Create new user group"""
+    try:
+        obj = await db.create(UserGroup(user_group=group_name))
```
gctucker, line 148: Shouldn't it be `name=`?
```suggestion
        obj = await db.create(UserGroup(name=group_name))
```

api/main.py
JenySadadia: Since the `group_name` is a string and not a Pydantic model, by default, it will be treated as a path or query parameter.
See https://fastapi.tiangolo.com/tutorial/body/#request-body-path-query-parameters
We need to create a model if we want to receive it in the request body. Or we can use `Body(...)` as suggested here https://github.com/tiangolo/fastapi/discussions/9010. But this way we won't have any validation for the input.
Also, I don't see much of an issue here as we have two separate methods i.e. GET and POST.

api/main.py
JenySadadia: Oh, yes. I missed it somehow. Let me fix it.

api/main.py
gctucker: I mean this, just like when creating a new node:
```python
@app.post('/group', response_model=UserGroup)
async def post_group(group: UserGroup, current_user: User = Security(get_user, scopes=["admin"])):
```

gctucker: About the URI paths, the only inconsistency I can think of is that we don't know the id before creating a group (same with users and nodes...).  Apart from this, we can have:
```
http://baseurl.com/group/<group-id>
```
and then use POST, PUT and GET with the same URI path to manage this resource.

Note: In an ideal world we could let the client create the IDs (like with KCIDB) but I think this would be suboptimal from a MongoDB point of view as we can't create ObjectID() with any arbitrary ID.  So we would need to create another index and have both the MongoDB `_id` and the user-supplied ID.

JenySadadia: > About the URI paths, the only inconsistency I can think of is that we don't know the id before creating a group (same with users and nodes...). Apart from this, we can have:
> 
> ```
> http://baseurl.com/group/<group-id>
> ```

No, we don't need an id to create a user group. Please look at the sample command below:
```
$ curl -X 'POST' 'http://localhost:8001/latest/group' -H 'accept: application/json'   -H 'Content-Type: application/json'  -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJhZG1pbiIsInNjb3BlcyI6WyJhZG1pbiJdfQ.4fGx_hBFdD-1JG7z38A-E0v8vkjnCa3Zj6a1p1fqWyk' -d '{"name": "kernelci"}'
{"_id":"648ff894bd39930355ed16ad","name":"kernelci"}
```
And yes, we already have `groups` and `group/{group_id}` endpoints implemented.

gctucker: Yes I know we don't need an ID, and that's precisely the only inconsistency I can see with how the URIs are designed.  Ideally, the same URI should be used for creating, updating, getting, deleting (all the HTTP verbs) a same resource.  I think we can have an exception here for creating objects as it's a pretty common use-case.  If we need the API to be closer to KCIDB then we might have to revisit this and rely on the client to create an ID.

JenySadadia: In what scenarios we would need to let the client create an ID?

gctucker: KCIDB is based on the design principles that the same data sent to the API should be retrievable, so that means all the fields have to be provided by the user.  Also it's designed for gathering data from multiple CI systems that already have their own database with their own IDs, like KernelCI is sending data with the IDs generated by MongoDB to KCIDB.  So if we wanted to align the new KernelCI API with KCIDB's schema we would also require users to provide IDs, and probably also timestamps etc.  But it's not a requirement for now, I think we can consider this as a long-term potential goal.  We should just try to avoid designing the new API in ways that will make it harder to align with KCIDB later.

api/main.py
```diff
@@ -138,6 +139,42 @@ async def post_user(
...
+        ) from error
+    await pubsub.publish_cloudevent('user_group', {'op': 'created',
+                                                   'id': str(obj.id)})
+    return obj
+
+
+@app.get('/groups', response_model=PageModel)
+async def get_user_groups(request: Request):
+    """Get all the user groups if no request parameters have passed.
+       Get all the matching user groups otherwise."""
```
gctucker, line 163: By the way, I thought docstrings were meant to have only a one-line summary first and then a blank line and paragraphs like git commits:
https://peps.python.org/pep-0257/#multi-line-docstrings

Following this convention, it could be reworded with something like "Find user groups" as a one-line summary and then the full description as a paragraph.

gctucker: Ah but I see this convention wasn't followed for the other endpoints either, so that's not really an issue with this PR.  I guess we can just have a follow-up to enable `pydocstyle` and fix all the warnings.

gctucker (approved):
OK I think it's all ready to be merged now.  We probably should update `kci user` when all the changes are ready on the API side to cover groups as well.

api/main.py
JenySadadia: So in that case we would need to introduce another `id` field that the user can provide as you suggested.
I think we can do that easily without messing with the existing schema.

Issue #256, opened 2023-05-04
Add user ownership to nodes


gctucker:
The new API allows more flexibility and any user with a token should be able to post results.  However, to be able to distinguish the results coming from the main KernelCI pipeline and for users to be able to find their own data, it would seem necessary to introduce the concept of nodes ownership.  Each node already has a creation date, we could add a field with the ID of the user who created it.  Then we could have simple rules such as only this user could update the node document, but others could refer to it e.g. to add child nodes.  We might need to add groups as a follow-up, for example all the runtimes orchestrated by the main pipeline could have users in some main KernelCI group and that could be used to sort results.  Also it could be a way for multiple users in a single organisation to submit data and update each others' nodes.

Before writing custom code for this, it would seem worth investigating existing frameworks.  We're already using JWT tokens which contain scopes and the OAuth2 standard for user authentication, and FastAPI has a security module with tools for this.  It would seem very simple to just add a user ID and add checks whenever a node is being updated whether the request was sent by the same user if we wanted to get started quickly.  But if we can directly adopt an existing framework with potentially some standard features to make the new API easier to use with other services.  There's a [StackOverflow discussion](https://stackoverflow.com/questions/58864648/oauth-2-0-flow-for-user-groups-organizations) about this among other things, it mentions LDAP so that could potentially be a way forward.

- [ ] `UserGroup` model implemented
- [ ] Logic in place to have users belong to groups
  - [x] #283

JenySadadia:
As per my investigation, `FastAPI` doesn't have any plugin/tool or package to serve this purpose.
Some packages like https://pypi.org/project/fastapi_permissions/ and https://github.com/fastapi-users/fastapi-users provide utilities for user creation but we already have it in place with `OAuth2`. So I don't think any of them will be useful in this case.
`FastAPI` doesn't have any integration for LDAP as well. Maybe this can be used if we want to integrate LDAP operations manually https://www.python-ldap.org/en/python-ldap-3.3.0/.

Bdw this can be done using `scopes` from `OAuth2` scheme:
We can create a model `group` with `name` and `id` fields.
We will have two default groups `users` and `admin` group.
By default, user accounts with `admin` scope will be added to `admin` group, and `users` scope will be added to `users` group. `Node` will have two additional fields. One is the user ID of the user that created the node and another is the group ID that the user belongs to. When we receive `PUT` node request, we can check if it's the same user that created the object and if so we can grant permission for the requested action.
[...]

PR #268, opened 2023-06-14
User ownership to nodes

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/256

Add user ownership to nodes. 
- Add `Node.user_id` field to store user ID that created the node
- Added method to authorize user for PUT node requests
- Exclude `user_id` field from response data of node related requests i.e. GET, PUT, and POST

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -390,7 +390,8 @@ async def post_node(node: Node, current_user: str = Depends(get_user)):
 
 
 @app.put('/node/{node_id}', response_model=Node, response_model_by_alias=False)
-async def put_node(node_id: str, node: Node, token: str = Depends(get_user)):
+async def put_node(node_id: str, node: Node,
+                   user: str = Depends(authorize_user)):
     """Update an already added node"""
     node.id = ObjectId(node_id)
     node_from_id = await db.find_by_id(Node, node_id)
@@ -406,6 +407,11 @@ async def put_node(node_id: str, node: Node, token: str = Depends(get_user)):
             status_code=status.HTTP_400_BAD_REQUEST,
             detail=message
         )
+
+    # Do not update node ownership fields
+    update_data = node.dict(exclude={'user', 'user_groups'})
+    node = node_from_id.copy(update=update_data)
+
     obj = await db.update(node)
     data = _get_node_event_data('updated', obj)
     await pubsub.publish_cloudevent('node', data)
--- a/api/main.py
+++ b/api/main.py
@@ -342,8 +342,17 @@ async def get_nodes_count(request: Request, kind: str = "node"):
         ) from error
 
 
+async def _verify_user_group_existence(user_groups: List[str]):
+    """Check if user group exists"""
+    for group_name in user_groups:
+        if not await db.find_one(UserGroup, name=group_name):
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail=f"User group does not exist with name: {group_name}")
+
+
 @app.post('/node', response_model=Node, response_model_by_alias=False)
-async def post_node(node: Node, token: str = Depends(get_user)):
+async def post_node(node: Node, current_user: str = Depends(get_user)):
     """Create a new node"""
     if node.parent:
         parent = await db.find_by_id(Node, node.parent)
@@ -352,6 +361,9 @@ async def post_node(node: Node, token: str = Depends(get_user)):
                 status_code=status.HTTP_404_NOT_FOUND,
                 detail=f"Parent not found with id: {node.parent}"
             )
+
+    await _verify_user_group_existence(node.user_groups)
+    node.owner = current_user.username
     obj = await db.create(node)
     data = _get_node_event_data('created', obj)
     await pubsub.publish_cloudevent('node', data)
--- a/api/main.py
+++ b/api/main.py
@@ -118,6 +118,25 @@ async def get_user(user: User = Depends(get_current_user)):
     return user
 
 
+async def authorize_user(node_id: str, user: User = Depends(get_current_user)):
+    """Return the user if active, authenticated, and authorized"""
+    if not user.active:
+        raise HTTPException(status_code=400, detail="Inactive user")
+
+    # Only the user that created the node or any other user from the permitted
+    # user groups will be allowed to update the node
+    node_from_id = await db.find_by_id(Node, node_id)
+    if node_from_id.owner:
+        if not user.username == node_from_id.owner:
+            if not any(group.name in node_from_id.user_groups
+                       for group in user.groups):
+                raise HTTPException(
+                    status_code=status.HTTP_401_UNAUTHORIZED,
+                    detail="Unauthorized to complete the operation"
+                )
+    return user
+
+
 @app.post('/user/{username}', response_model=User,
           response_model_by_alias=False)
 async def post_user(
--- a/api/models.py
+++ b/api/models.py
@@ -248,6 +248,13 @@ class Node(DatabaseModel):
     holdoff: Optional[datetime] = Field(
         description="Node expiry timestamp while in Available state"
     )
+    owner: Optional[str] = Field(
+        description="Username of node owner"
+    )
+    user_groups: List[str] = Field(
+        default=[],
+        description="User groups that are permitted to update node"
+    )
 
     _OBJECT_ID_FIELDS = ['parent']
     _TIMESTAMP_FIELDS = ['created', 'updated', 'timeout', 'holdoff']
--- a/tests/e2e_tests/test_node_handler.py
+++ b/tests/e2e_tests/test_node_handler.py
@@ -35,6 +35,7 @@ async def create_node(test_async_client, node):
             'holdoff',
             'kind',
             'name',
+            'owner',
             'path',
             'parent',
             'result',
@@ -42,6 +43,7 @@ async def create_node(test_async_client, node):
             'state',
             'timeout',
             'updated',
+            'user_groups',
         }
     return response
 
@@ -70,6 +72,7 @@ async def get_node_by_id(test_async_client, node_id):
             'holdoff',
             'kind',
             'name',
+            'owner',
             'path',
             'parent',
             'result',
@@ -77,6 +80,7 @@ async def get_node_by_id(test_async_client, node_id):
             'state',
             'timeout',
             'updated',
+            'user_groups',
         }
     return response
 
@@ -133,6 +137,7 @@ async def update_node(test_async_client, node):
             'holdoff',
             'kind',
             'name',
+            'owner',
             'path',
             'parent',
             'result',
@@ -140,4 +145,5 @@ async def update_node(test_async_client, node):
             'state',
             'timeout',
             'updated',
+            'user_groups',
         }
--- a/tests/e2e_tests/test_regression_handler.py
+++ b/tests/e2e_tests/test_regression_handler.py
@@ -37,6 +37,7 @@ async def create_regression(test_async_client, regression_node):
             'holdoff',
             'kind',
             'name',
+            'owner',
             'path',
             'parent',
             'result',
@@ -45,6 +46,7 @@ async def create_regression(test_async_client, regression_node):
             'state',
             'timeout',
             'updated',
+            'user_groups',
         }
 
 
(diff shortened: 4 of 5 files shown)


gctucker:
Looks like this can be rebased and the skip label removed now that #265 and #266 have been merged.

JenySadadia:
> Looks like this can be rebased and the skip label removed now that #265 and #266 have been merged.

I am working on adding `NodeOwnership` model instead of adding `Node.username` and `Node.user_groups` fields.
That way we can easily manage update operations and permissions related to ownership. That can be separate from actual node objects.

gctucker:
What's the issue with having `user` and `group` fields directly in the Node model?

JenySadadia:
> What's the issue with having `user` and `group` fields directly in the Node model?

Fields for ownership should be editable by authorized users only.
Any user that updates the node should not be able to update `username` field. Also, `groups` should only be updated by node owner (and maybe by admin). It is easy to have a separate model than to have these constraints in POST node.
Also, we can have different endpoints defined for this model just like the user profile model.

gctucker:
I don't see how this makes any difference for the node model though, the API will need to check if the user sending PUT is the owner of the node or belongs to the node's group in the application layer, this can't be done at the API validation layer as the node needs to be retrieved from the database first.

JenySadadia:
> I don't see how this makes any difference for the node model though, the API will need to check if the user sending PUT is the owner of the node or belongs to the node's group in the application layer, this can't be done at the API validation layer as the node needs to be retrieved from the database first.

As per our last discussion, I have implemented `NodeOwnership` model. The ownership object will be created in POST request. It will be used to authorize user in PUT requests.

JenySadadia:
Also, need to add `unique_items` constraint to `NodeOwnership.user_groups`.

api/main.py
```diff
@@ -341,8 +362,18 @@ async def get_nodes_count(request: Request, kind: str = "node"):
...
+    for group_name in user_groups:
+        if not await db.find_one(UserGroup, name=group_name):
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail=f"User group does not exist with name: {group_name}")
+
+
 @app.post('/node', response_model=Node, response_model_by_alias=False)
-async def post_node(node: Node, token: str = Depends(get_user)):
+async def post_node(node: Node, user_groups: List[str] = [],
```
gctucker, line 375: BTW The linter warning here is about using `None` instead, as otherwise the same object reference keeps being used for default values

gctucker:
The issue with this is that it's creating separate documents with references to a User and a Node, which effectively creates a typical many-to-many relationship as used in SQL databases.  This means that every time a node is being handled, we need to look up a NodeOwnership document and then maybe also a User document.  However, NoSQL databases like MongoDB aren't good for that.  Instead, the typical way of doing it is by embedding the models in the main document itself.  So we could have a NodeOwnership model but that would need to be stored as Node.ownership.

I understand why you wanted to do that, to be able to deal with POST request permissions separately for the Node fields and for NodeOwnership, but in this case we can't realistically do it this way mainly for performance reasons.

I believe the best way to handle this right now is to have a Node model with just plain fields `user` and `groups` containing the names of the owner and groups the node belongs to, but with some logic at the application level to not allow changing them when handling PUT requests (except maybe for admin users).  I'm not sure if we can make them constant fields but if we can then that would work too initially.  It would keep the implementation very simple, reduce the burden of testing and be less bug-prone, and also it will appear directly in the API generated documentation.  The limitation would be that nodes can't be assigned to another owner or groups once created, which might be entirely acceptable even in the long-term anyway as I don't think we have a known use-case where changing ownership is useful as long as all nodes are publicly readable.

JenySadadia:
> The issue with this is that it's creating separate documents with references to a User and a Node, which effectively creates a typical many-to-many relationship as used in SQL databases. This means that every time a node is being handled, we need to look up a NodeOwnership document and then maybe also a User document. However, NoSQL databases like MongoDB aren't good for that. Instead, the typical way of doing it is by embedding the models in the main document itself. So we could have a NodeOwnership model but that would need to be stored as Node.ownership.
> 
> I understand why you wanted to do that, to be able to deal with POST request permissions separately for the Node fields and for NodeOwnership, but in this case we can't realistically do it this way mainly for performance reasons.
> 
> I believe the best way to handle this right now is to have a Node model with just plain fields `user` and `groups` containing the names of the owner and groups the node belongs to, but with some logic at the application level to not allow changing them when handling PUT requests (except maybe for admin users). I'm not sure if we can make them constant fields but if we can then that would work too initially. It would keep the implementation very simple, reduce the burden of testing and be less bug-prone, and also it will appear directly in the API generated documentation. The limitation would be that nodes can't be assigned to another owner or groups once created, which might be entirely acceptable even in the long-term anyway as I don't think we have a known use-case where changing ownership is useful as long as all nodes are publicly readable.

Hello @gctucker Thanks for the feedback.
I read some documentation from mongodb about this and it seems like the better approach here is to embed fields in the same document rather than query database at each PUT request. The ownership related information is not hard to manage in a single document and also we would need it frequently.
I will update the PR accordingly.

gctucker:
Yes, then having a model just for that or fields directly in Node is more like an implementation detail as it'll appear pretty much in the same way in MongoDB documents.  Maybe a sub-model will create a embedded dictionary but that should be a minor thing from a performance and database design point-of-view.  Whatever makes the models and application logic better designed would work in this case.

JenySadadia:
Updated the PR.
Re-assigned ownership details in `put_node` as of now. It will need to be tweaked once we discuss about various permissions for updating node ownership.
The basic case I can think of is, if the user is an owner then should be allowed to update `user_groups`.
Also, admin users should have permission to update that info.

JenySadadia:
We can use `update_one` instead of `replace_one` for updating documents partially.

gctucker:
> The basic case I can think of is, if the user is an owner then should be allowed to update `user_groups`.

I mean, from an actual use-case.  Why change the group assigned to an existing node?

JenySadadia:
> > The basic case I can think of is, if the user is an owner then should be allowed to update `user_groups`.
> 
> I mean, from an actual use-case. Why change the group assigned to an existing node?

If the owner wants to share the node with some other group that was not added at the time of node creation.

gctucker:
We don't have this problem as long as all nodes are publicly readable.  Changing the group would only be to allow some other group of users to update the node later on, and I'm pretty sure we can live without the possibility to do this for a long time.  In fact, if a node is in Done state then I think we've said it shouldn't be possible to update it anyway?

JenySadadia:
> We don't have this problem as long as all nodes are publicly readable. Changing the group would only be to allow some other group of users to update the node later on, and I'm pretty sure we can live without the possibility to do this for a long time. In fact, if a node is in Done state then I think we've said it shouldn't be possible to update it anyway?

Yes, update is only possible if the node is not in "done" state. Okay. So re-assigning values would do for now.

JenySadadia:
After I get approval on initial commits, I also need to add changes for `put_nodes`.

JenySadadia:
> > We don't have this problem as long as all nodes are publicly readable. Changing the group would only be to allow some other group of users to update the node later on, and I'm pretty sure we can live without the possibility to do this for a long time. In fact, if a node is in Done state then I think we've said it shouldn't be possible to update it anyway?
> 
> Yes, update is only possible if the node is not in "done" state. Okay. So re-assigning values would do for now.

Found a better solution to deal with this issue. 
Pydantic provides `exclude` parameter to be used while converting the model to a dictionary. Used it to exclude node ownership fields in PUT requests.

api/models.py
```diff
@@ -247,6 +247,13 @@ class Node(DatabaseModel):
     holdoff: Optional[datetime] = Field(
         description="Node expiry timestamp while in Available state"
     )
+    user: Optional[str] = Field(
+        description="Username of node owner"
+    )
+    user_groups: List[str] = Field(
+        default=[],
+        description="User groups that are permitted to update node"
+    )
```
gctucker, line 256: Oh I thought we wanted to just have one group, not a list of groups.  I guess this is fine but a bit more complicated.

api/models.py
```diff
@@ -247,6 +247,13 @@ class Node(DatabaseModel):
     holdoff: Optional[datetime] = Field(
         description="Node expiry timestamp while in Available state"
     )
+    user: Optional[str] = Field(
```
gctucker, line 250: Maybe `owner` would be a bit better, but that's a detail.  We can always tweak things like this a bit later.

api/main.py
```diff
@@ -31,6 +31,7 @@
 from .db import Database
 from .models import (
     Node,
+    NodeOwnership,
```
gctucker: This model doesn't exist yet, maybe it's created in a later commit.

api/main.py
```diff
@@ -351,6 +361,9 @@ async def post_node(node: Node, token: str = Depends(get_user)):
                 status_code=status.HTTP_404_NOT_FOUND,
                 detail=f"Parent not found with id: {node.parent}"
             )
+
+    await _verify_user_group_existence(node.user_groups)
```
gctucker, line 383: One thought, we could also skip this and instead accept that the owner and group may not always exist in the database.  That way, when renaming or deleting users and groups we don't have to update retroactively all the older nodes.  I think it's easy to handle as the user and group lookups are done directly in the application logic.

api/main.py
```diff
@@ -31,7 +31,6 @@
 from .db import Database
 from .models import (
     Node,
-    NodeOwnership,
```
gctucker: Right, that's some diff noise with the previous commit.

api/main.py
```diff
@@ -118,6 +117,25 @@ async def get_user(user: User = Depends(get_current_user)):
...
 
+async def authorize_user(node_id: str, user: User = Depends(get_current_user)):
+    """Return the user if active, authenticated, and authorized"""
+    if not user.active:
+        raise HTTPException(status_code=400, detail="Inactive user")
+
+    # Only the user that created the node or any other user from the permitted
+    # user groups will be allowed to update the node
+    node_from_id = await db.find_by_id(Node, node_id)
+    if node_from_id.user:
```
gctucker, line 128: Shouldn't all nodes have an owner?

gctucker (commented):
Just a few more comments, generally speaking I think this is good to go now.

api/models.py
JenySadadia: I agree. I will rename it.

api/main.py
JenySadadia: This check is due to already existing nodes do not have ownership details.
Maybe we can create a user and assign it to all the existing nodes?

api/main.py
JenySadadia: I don't think inserting invalid data into the database for the sake of simplicity is a good idea.

api/models.py
JenySadadia: Done.

api/models.py
gctucker: How is this related?  I like to think of Unix file systems, users can belong to a number of groups but files have only one group.

The current implementation seems to be working so that's fine, we can always simplify things by reducing it to a single group later on.

api/main.py
gctucker: It depends how we define invalid, having an non-existing object ID would be more problematic I think.  But even that we're not checking afaik, say when setting a parent node id.  Ultimately, if the user provides bad data then it's a problem on the user's side.

What's important though is that we define exactly who is in charge of the data consistency.  If we want to check that groups exist and other similar checks around node ownership then we need to do it entirely, and update all nodes when a group or user gets renamed etc.  Otherwise, we can just say it's something the user is in charge of and be fault-tolerant to return HTTP 4xx errors when doing operations on data that doesn't exist.  Having fixed user names and groups is also a way to solve this by not having the use-case where things need to be renamed and updated.

api/main.py
gctucker: Yes I think all existing nodes should be updated with `admin` user or something like that.  We can also delete old nodes as this is the staging database.

gctucker (approved):

gctucker:
Tested OK on staging (no side-effects on the pipeline).

api/main.py
JenySadadia: Yes, I think deleting nodes would be easier.

PR #277, opened 2023-06-21
Remove  `User.is_admin` field

JenySadadia:
Created on top of https://github.com/kernelci/kernelci-api/pull/266

Fixes https://github.com/kernelci/kernelci-api/issues/278

`User.is_admin` field was introduced to verify if the user is an admin user. Now we have a specific user group `admin` and all the admin users will be assigned to it. Instead of `is_admin` flag, we can use `User.groups` to check if the user belongs to the `admin` group and hence we can check if it is an admin user.
Thus, we can drop `User.is_admin` field now.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -214,7 +214,8 @@ async def login_for_access_token(
             )
 
     if 'admin' in form_data.scopes:
-        if user.is_admin is not True:
+        if not user.groups or not any(
+                group.name == 'admin' for group in user.groups):
             raise HTTPException(
                 status_code=status.HTTP_400_BAD_REQUEST,
                 detail="Not allowed to use admin scope"
--- a/api/main.py
+++ b/api/main.py
@@ -120,7 +120,7 @@ async def get_user(user: User = Depends(get_current_user)):
 @app.post('/user/{username}', response_model=User,
           response_model_by_alias=False)
 async def post_user(
-        username: str, password: Password, is_admin: bool = False,
+        username: str, password: Password,
         groups: List[str] = Query([]),
         current_user: User = Security(get_user, scopes=["admin"])):
     """Create new user"""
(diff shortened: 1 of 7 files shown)


gctucker:
Please rebase now that #266 is merged.

JenySadadia:
> Please rebase now that #266 is merged.

Done.

gctucker:
Ideally there should be a migration to update all User documents with the `admin` group for admin users and remove the `is_admin` field.

gctucker:
This could also just be done by hand in MongoDB shell on staging as the schema is not quite set in stone yet, but having a migration is probably a good way to also check that migrations are working as expected.

gctucker:
Also the users collection in the staging database might need some manual clean-up anyway :p

```
> db.user.find({}, {is_admin: 1, groups: 1})
{ "_id" : ObjectId("62058f7a91aebf7ab63f1492") }
{ "_id" : ObjectId("62554673a68f9bac63a0ead2") }
{ "_id" : ObjectId("62f3ec222d0a223569d55be7"), "is_admin" : 1 }
{ "_id" : ObjectId("62f3ed1b8238a289a951d1be"), "is_admin" : false }
{ "_id" : ObjectId("6359369fa2736d6b586b7fec"), "is_admin" : 1 }
{ "_id" : ObjectId("63593e30d3f352d83f163878"), "is_admin" : false }
{ "_id" : ObjectId("64705c1ae76325f50371875e"), "is_admin" : 0 }

```

First all the users should have a `is_admin` field and it should be set to `true` or `false` as this is a boolean field.  Then we can apply a migration to remove the field and add `groups` with `[admin]` when `is_admin` is `true`.

gctucker (approved):
OK I think we can merge this without the migration as things are still moving too much.

PR #279, opened 2023-06-21
api.main: GET users and user profiles

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/300

Implement an endpoint to get users.
It also accepts request query parameters to get matching user.
e.g. `http://API_URL/users?username=test`

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -173,6 +173,26 @@ async def post_user(
     return obj
 
 
+@app.get('/users/profile', response_model=PageModel,
+         response_model_include={"items": {"__all__": {"profile": {
+                                    "username", "groups"}}},
+                                 "total": {"__all__"},
+                                 "limit": {"__all__"},
+                                 "offset": {"__all__"},
+                                 })
+async def get_users_profile(request: Request):
+    """Get profile of all the users if no request parameters have passed.
+       Get the matching user profile otherwise."""
+    query_params = dict(request.query_params)
+    # Drop pagination parameters from query as they're already in arguments
+    for pg_key in ['limit', 'offset']:
+        query_params.pop(pg_key, None)
+    paginated_resp = await db.find_by_attributes(User, query_params)
+    paginated_resp.items = serialize_paginated_data(
+        User, paginated_resp.items)
+    return paginated_resp
+
+
 @app.post('/group', response_model=UserGroup, response_model_by_alias=False)
 async def post_user_group(
         group: UserGroup,
--- a/api/main.py
+++ b/api/main.py
@@ -211,6 +211,16 @@ async def get_users(
     return paginated_resp
 
 
+@app.get('/user/{user_id}', response_model=User,
+         response_model_by_alias=False,
+         response_model_exclude={"profile": {"hashed_password"}})
+async def get_user_by_id(
+        user_id: str,
+        current_user: User = Security(get_user, scopes=["admin"])):
+    """Get user matching provided user id"""
+    return await db.find_by_id(User, user_id)
+
+
 @app.post('/group', response_model=UserGroup, response_model_by_alias=False)
 async def post_user_group(
         group: UserGroup,
--- a/api/main.py
+++ b/api/main.py
@@ -193,6 +193,24 @@ async def get_users_profile(request: Request):
     return paginated_resp
 
 
+@app.get('/users', response_model=PageModel,
+         response_model_exclude={"items": {"__all__": {"profile": {
+                                    "hashed_password"}}}})
+async def get_users(
+        request: Request,
+        current_user: User = Security(get_user, scopes=["admin"])):
+    """Get all the users if no request parameters have passed.
+       Get the matching users otherwise."""
+    query_params = dict(request.query_params)
+    # Drop pagination parameters from query as they're already in arguments
+    for pg_key in ['limit', 'offset']:
+        query_params.pop(pg_key, None)
+    paginated_resp = await db.find_by_attributes(User, query_params)
+    paginated_resp.items = serialize_paginated_data(
+        User, paginated_resp.items)
+    return paginated_resp
+
+
 @app.post('/group', response_model=UserGroup, response_model_by_alias=False)
 async def post_user_group(
         group: UserGroup,


gctucker:
This is returning the full User documents including the hashed passwords, so adding `skip` label for now.  I think we might need to define a PublicUser model just as a return type from the API for this kind of use-case.  It could contain things like the username, real name of the user, list of groups, created time etc. like say on a GitHub user profile.

JenySadadia:
> This is returning the full User documents including the hashed passwords, so adding `skip` label for now. I think we might need to define a PublicUser model just as a return type from the API for this kind of use-case. It could contain things like the username, real name of the user, list of groups, created time etc. like say on a GitHub user profile.

Yes, this needs to be tweaked once we add user profile model.

gctucker:
> > This is returning the full User documents including the hashed passwords, so adding `skip` label for now. I think we might need to define a PublicUser model just as a return type from the API for this kind of use-case. It could contain things like the username, real name of the user, list of groups, created time etc. like say on a GitHub user profile.
> 
> Yes, this needs to be tweaked once we add user profile model.

Even then, I'm not sure all the user profile should be public.  So we could have this in the database:
```
User:
  groups: []
  created:
  other_admin_field:
  profile: UserProfile:
    username:
    full_name
    email:
    preferences:
    other_private_attributes:
```
as a way to allow POST queries to let the user update the `profile` but only admins can update the main User model.

Then we would also have this model, not stored in the database but only part of the user API:
```
UserPublicProfile:
  username:
  full_name:
  groups: []
  other_public_fields:
```

gctucker:
Another way to look at this, along the lines of the discussion around updating models for node ownership, is to just have a User model for the database and a UserProfile model shared with the public.  Then the application would encode rules about what fields can be read or written depending on the user initiating the query, rather than at the API validation level.  I think that's actually almost inevitable as things like `username` and `groups` would need to be accessible publicly or at least by the user but also need to be in the main User model and not a profile sub-model.

What is going to be more important is that we use the same UserProfile model for PUT and GET requests, so we get the same data in and out even if that's not directly an object stored in the database.  Then admins should be able to POST, PUT and GET the full User objects which would be directly the same model as in the database.  The only potential exception to this that I can see is for the password field, maybe it should never be returned at all and therefore not included in the data models used by the API but handled with a dedicated endpoint instead.

JenySadadia:
> Another way to look at this, along the lines of the discussion around updating models for node ownership, is to just have a User model for the database and a UserProfile model shared with the public. Then the application would encode rules about what fields can be read or written depending on the user initiating the query, rather than at the API validation level. I think that's actually almost inevitable as things like `username` and `groups` would need to be accessible publicly or at least by the user but also need to be in the main User model and not a profile sub-model.
> 
> What is going to be more important is that we use the same UserProfile model for PUT and GET requests, so we get the same data in and out even if that's not directly an object stored in the database. Then admins should be able to POST, PUT and GET the full User objects which would be directly the same model as in the database. The only potential exception to this that I can see is for the password field, maybe it should never be returned at all and therefore not included in the data models used by the API but handled with a dedicated endpoint instead.

Please see https://github.com/kernelci/kernelci-api/pull/292

JenySadadia:
Updated the PR to get user public profiles.

gctucker:
Well, #292 is not what I was suggesting with a different model for the API and for the database but I'll take a look next week.  I think we basically need to get the simplest implementation merged to get the feature enabled and then we can extend it as second priority.

JenySadadia:
> Well, #292 is not what I was suggesting with a different model for the API and for the database but I'll take a look next week. I think we basically need to get the simplest implementation merged to get the feature enabled and then we can extend it as second priority.

I have implemented `UserProfile` model. That will be accessible by all users with GET `/users` request. Maybe we should have `/users/public` for getting user profiles and `/users` for getting user models by admin. POST `/user` will create `UserProfile` and `User` models.

JenySadadia:
Commands will look like:
```
$ curl http://localhost:8001/latest/users/profile
{"items":[{"id":"64a7d92d6f9a9950b9507485","username":"admin","groups":[{"id":"64a2bcff4ba7b7c80ee3c570","name":"admin"}]},{"id":"64a84222c061af66e4d0071a","username":"jeny","groups":[]}],"total":2,"limit":50,"offset":0}
```
```
$ curl http://localhost:8001/latest/users/profile?username=jeny
{"items":[{"id":"64a84222c061af66e4d0071a","username":"jeny","groups":[]}],"total":1,"limit":50,"offset":0}
```
```
$ curl -X 'GET' 'http://localhost:8001/latest/users' -H 'accept: application/json'   -H 'Content-Type: application/json'  -H 'Authorization: Bearer TOKEN'
{"items":[{"id":"64a7d92d6f9a9950b9507486","active":true,"profile":{"id":"64a7d92d6f9a9950b9507485","username":"admin","hashed_password":"$2b$12$N.GfaUJy/a775R9PQcH6IeBM0TCCwYWA2XLKKYZG6.MiVldlWuh6O","groups":[{"id":"64a2bcff4ba7b7c80ee3c570","name":"admin"}]}},{"id":"64a84222c061af66e4d0071b","active":true,"profile":{"id":"64a84222c061af66e4d0071a","username":"jeny","hashed_password":"$2b$12$P90NaB1LzKI/dfUPhJ4jBOKFlXFbsVmQI/4p8CuL4hXUFAo1JVzh2","groups":[]}}],"total":2,"limit":50,"offset":0}
```
```
$ curl -X 'GET' 'http://localhost:8001/latest/users?profile.username=jeny' -H 'accept: application/json'   -H 'Content-Type: application/json'  -H 'Authorization: Bearer TOKEN'
{"items":[{"id":"64a84222c061af66e4d0071b","active":true,"profile":{"id":"64a84222c061af66e4d0071a","username":"jeny","hashed_password":"$2b$12$P90NaB1LzKI/dfUPhJ4jBOKFlXFbsVmQI/4p8CuL4hXUFAo1JVzh2","groups":[]}}],"total":1,"limit":50,"offset":0}
```

JenySadadia:
Added handler to get user profile by ID:
```
$ curl http://localhost:8001/latest/users/profile/64a7d92d6f9a9950b9507485
{"id":"64a7d92d6f9a9950b9507485","username":"admin","groups":[{"id":"64a2bcff4ba7b7c80ee3c570","name":"admin"}]}
```

JenySadadia:
Updated the PR to accommodate user profile changes.

JenySadadia:
Tested OK on staging:
```
$ curl https://staging.kernelci.org:9000/latest/users/profile
{"items":[{"profile":{"username":"jeny","hashed_password":"$2b$12$UjgKqbbJbc3ZM03Hb5aaUO0OJ2KIaqXD.SosWh2U0EEY2tFDQRMi6","groups":[]}},{"profile":{"username":"testAdmin","hashed_password":"$2b$12$7nysdbTVI7dW6AeFDsMxs.bcc9q/k.1FXycqLmmRRV2mArBj74P3W","groups":[{"id":"649190b01432ccfdb575a250","name":"admin"}]}},{"profile":{"username":"staging.kernelci.org","hashed_password":"$2b$12$Dir/xrihCf50kozxKdTX6u5IhnbhldDBjOL446OOChCncwu4K4HIG","groups":[]}},{"profile":{"username":"gtucker","hashed_password":"$2b$12$Z2cjFiUlUiJeD8Si93xIV.4EIheWGwtnWNACpiLMJxONnj2npjsSS","groups":[]}},{"profile":{"username":"alex","hashed_password":"$2b$12$CCvyfkT1M3mVwCg/FEn.5ezINiQe5PbLcE.FdlXwHIXuyy9maNC5G","groups":[]}},{"profile":{"username":"mg-admin","hashed_password":"$2b$12$7sFqfTxfqSqPgz/4yn9tuuUju2G0dHMOpVlbJceik5hil5CwzdHeu","groups":[]}},{"profile":{"username":"rcn","hashed_password":"$2b$12$jce8d99FKxULmfBsMSACxuIALgu2NAYG8ufcre5ECARGURv.HYxY6","groups":[]}},{"profile":{"username":"admin","hashed_password":"$2b$12$WxssucnPAABBsihbWsU2ceDBQDmpMf8c7gbMuVP861DjTzay5wBf.","groups":[{"id":"649190b01432ccfdb575a250","name":"admin"}]}}],"total":8,"limit":50,"offset":0}
```

gctucker:
Adding the skip label as this is returning the hashed passwords again.

JenySadadia:
> Adding the skip label as this is returning the hashed passwords again.

Yes, I need to exclude it.

gctucker:
But what about not including passwords in the profile model?  I think there are a number of things to clarify before we can resolve this and complete the implementation.

JenySadadia:
> But what about not including passwords in the profile model? I think there are a number of things to clarify before we can resolve this and complete the implementation.

I have moved all the editable fields by the user to `UserProfile` and the password is one of them.

gctucker:
Please see https://github.com/kernelci/kernelci-api/pull/292#issuecomment-1638237249

api/main.py
```diff
@@ -173,6 +173,54 @@ async def post_user(
...
+        query_params.pop(pg_key, None)
+    paginated_resp = await db.find_by_attributes(User, query_params)
+    paginated_resp.items = serialize_paginated_data(
+        User, paginated_resp.items)
+    return paginated_resp
+
+
+@app.get('/users', response_model=PageModel,
+         response_model_exclude={"items": {"__all__": {"profile": {
+                                    "hashed_password"}}}})
```
gctucker, line 198: OK I think that should be fine for now.

gctucker (approved):

PR #280, opened 2023-06-21
api.main: fix response model to not have alias for `/whoami`

JenySadadia:
Use `response_model_by_alias` flag for `/whoami` endpoint to get response without alias. This is to fix `_id` field in the response.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -228,7 +228,7 @@ async def login_for_access_token(
     return {"access_token": access_token, "token_type": "bearer"}
 
 
-@app.get('/whoami', response_model=User)
+@app.get('/whoami', response_model=User, response_model_by_alias=False)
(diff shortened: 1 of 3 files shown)


gctucker:
@JenySadadia I think this is good to go but please rebase it and I'll try it by hand when it's deployed on staging.

JenySadadia:
> @JenySadadia I think this is good to go but please rebase it and I'll try it by hand when it's deployed on staging.

Done.

gctucker (approved):
Tested OK:
```
$ ./kci user whoami --api-config=staging
{
    "id": "62554673a68f9bac63a0ead2",
    "username": "gtucker",
    "hashed_password": "...",
    "active": true,
    "groups": []
}
```

PR #281, opened 2023-06-22
Fix `/groups` for pagination parameters

JenySadadia:
Fix `get_user_groups` to drop pagination parameters (i.e. limit and offset) from request query parameters.

Fixes: 7800343 ("api.main: add GET handler for user groups")

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -162,6 +162,11 @@ async def get_user_groups(request: Request):
     """Get all the user groups if no request parameters have passed.
        Get all the matching user groups otherwise."""
     query_params = dict(request.query_params)
+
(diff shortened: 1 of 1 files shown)


gctucker:
We might need to create some sample dataset to test the API, in this case with a bunch of groups to exercise the pagination logic in a real end-to-end test.  That sounds worth a separate GitHub issue, about generating a sample dataset.

JenySadadia:
> We might need to create some sample dataset to test the API, in this case with a bunch of groups to exercise the pagination logic in a real end-to-end test. That sounds worth a separate GitHub issue, about generating a sample dataset.

Yes, I'll create one.

gctucker (approved):

PR #284, opened 2023-06-27
Use a MongoDB service string rather than explicit host and port

gctucker:
Use a `mongo://` kind of service string URL rather than a host and port to connect to a MongoDB instance.  Also look for a `MONGO_SERVICE` environment variable to load it if provided, which can contain a username and password.  This makes it possible to connect exactly the same way to the local instance in the `db` container or a remote Atlas instance.

### DIFF
--- a/api/db.py
+++ b/api/db.py
@@ -34,8 +34,8 @@ class Database:
         'gte': '$gte',
     }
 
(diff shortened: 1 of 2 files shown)


gctucker:
Tested OK on staging as well as with a test Atlas account.

PR #285, opened 2023-06-27
Rework admin user and group setup using `api.admin` module

gctucker:
Rework the tool to bootstrap a new database by creating an admin user and group directly into any MongoDB instance while reusing the Authentication, Database and model classes from the API.  This doesn't require any service to be running apart from MongoDB.


api/auth.py
```diff
@@ -39,8 +39,9 @@ class Authentication:
     should be a db.Database object.
     """
 
+    CRYPTCTX = CryptContext(schemes=["bcrypt"], deprecated="auto")
```
JenySadadia, line 42: I'd suggest `CRYPT_CTX` for better readability.
```suggestion
    CRYPT_CTX = CryptContext(schemes=["bcrypt"], deprecated="auto")
```

api/admin.py
```diff
@@ -0,0 +1,64 @@
...
+        return None
+    password = getpass.getpass(f"Password for user '{args.username}': ")
+    hashed_password = Authentication.get_password_hash(password)
+    print(f"Creating {username} user...")
+    return await db.create(User(
+        username=username,
+        hashed_password=hashed_password,
+        is_admin=True,
+        groups=[admin_group]
+    ))
```
JenySadadia, line 45: We have removed `is_admin` field.
```suggestion
    return await db.create(User(
        username=username,
        hashed_password=hashed_password,
        groups=[admin_group]
    ))
```

scripts/setup_admin_user
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -e
+
```
JenySadadia, line 4: How about starting `db` service from the script itself?

api/admin.py
gctucker: Yes, that was before the changes were merged in the User schema.  I'll update it.

scripts/setup_admin_user
gctucker: Because it's not necessarily for the `db` service, in fact I did this to set up a new database in Atlas.

JenySadadia: Oh, yes. Mongo URL can be anything and not just the local `db` service.

gctucker:
I'll also make a separate PR to update the documentation about creating the initial admin user accordingly.

api/admin.py
```diff
@@ -0,0 +1,63 @@
...
+        username=username,
+        hashed_password=hashed_password,
+        groups=[admin_group]
+    ))
+
+
+async def main(args):
+    db = Database(args.mongo, args.database)
+    group = await setup_admin_group(db)
+    user = await setup_admin_user(db, args.username, group)
```
JenySadadia, line 50: `pylint` is complaining about the unused `user` variable.

JenySadadia:
Tested OK on staging.

JenySadadia:
Also, need to drop https://github.com/kernelci/kernelci-api/blob/main/init_user_groups script.

gctucker:
Right, I think that can be done as a follow-up.  The documentation also needs to be updated when removing the previous script.

api/admin.py
JenySadadia: This can be fixed in a follow-up PR as well. Merging it for now.

JenySadadia:
> Right, I think that can be done as a follow-up. The documentation also needs to be updated when removing the previous script.

Sounds good.

PR #287, opened 2023-07-05
adding kubernetes manifest to test kernelci-api minikube deployment

### LINKED ISSUE #274: Standalone API deployment tested locally with MiniKube
This is the API counterpart of #273, to replicate the standard `docker-compose` deployment in a local MiniKube Kubernetes cluster.  If #273 is done first, then the pipeline can be running in MiniKube too and connect to the API services within the same cluster.  In principle this is likely to be the preferred way of deploying the API and pipeline services from a performance point of view.  Having them in separate clusters might help reduce downtime, for example if we only need to migrate the pipeline.  These are things we can already start considering and experiment with using MiniKube before recreating it in the Cloud with Azure as a follow-up task.

- [x] K8s manifest files for API: https://github.com/kernelci/kernelci-api/pull/287
- [x] Tested with `kci` command line on the host, connecting to the API service inside MiniKube e.g. `kci api hello`
- [x] Tested with full pipeline services running also in MiniKube if possible, and/or in `docker-compose` on the host
- [x] Configuration file created for automatically deploying the API services added to `kernelci-api` repository
- [x] Summary wrt deploying both pipeline and API services in Kubernetes

sbdtu5498:
Fixes #274
@gctucker @JenySadadia 
Some nits are remaining. Will try to add the README.md by tonight. And we can have a demonstration in tomorrow's meeting.


kube/api-service.yaml
```diff
@@ -0,0 +1,17 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: kernelci-api
+spec:
+  selector:
+    app: kernelci-api
+  ports:
+    - protocol: TCP
+      port: 8000
```
JenySadadia, line 15: We are exposing `api` service on port 8001.
```suggestion
      port: 8001
```

sbdtu5498: Oh, service port and Local port are not the same thing. So service port is just being used to accumulate the traffic from all the target pods' ports and provide stable endpoints to them because pods are too fragile. Then this traffic can be forwarded to the local host via Kubectl port-forward. I hope this explanation makes it a bit clear. Let me know anything else you would like to discuss.

sbdtu5498: Yes that's possible. It can be done via ingresses as well or this is a type of service called ClusterIP that is used for internal networking. But if we have cloud provider based load balancers, we can also use LoadBalancer based service. It can be done in multiple different way. We can also forward the traffic to local instance (i.e. outside kubernetes cluster) via kubectl port-forward <svc-name> desired-port:8000 something like this.

gctucker: Right, I think initially we would need to have something that's the most generic possible and not tailored to any particular Cloud provider or the definitive setup we'll have for kernelci.org.  So I would suggest to stick to this current approach as far as this PR is concerned with nothing in front of the port directly exposed by Kubernetes, which is only fine for a local setup but at least we'll get all this part tested and sorted out first.  I guess this should provide some reference for any Kubernetes-based deployment.

Then we can add a variant to set things up with SSL in Azure and also use the Atlas account for MongoDB rather than a local `db` service, but keep plain MongoDB in the reference config to not require Atlas as an external dependency - only as an option for those who which to use it.  This could then become the production-like deployment we would be using for beta-testing or "early access".

How does that sound?

sbdtu5498: Yeah, that's what I have thought. This deployment is already production ready. A few changes needs to be done. Not in API one in pipeline part there were a few security concerns such as DinD-based implementation, I have two more alternative. And for this one I have use mongo-data PVC as host path that is only for single node testing. When we will use Atlas we would have to change it a bit and use storage class that corresponds to Atlas. And then networking needs to be done, it would not require any heavylifting though. A bit of changes such as whether we would be using proxy such as nginx or envoy or can we use loadbalancers provided by Cloud providers themselves. But local testing is almost reaching it's final stage. Then we can move on to the next step soon.

gctucker: > storage class that corresponds to Atlas

What do you mean by this?  If we use Atlas, we just don't run the `db` service at all but connect directly to Atlas.

sbdtu5498: Yeah, something like that. Basically, every storage providers provide ways to generically consume storage via Kubernetes API. So the volume controllers in Kubernetes take care of underlying intricacies. We just have to create persistent volume claims and when we will write persistent volume manifest that has storage details after we buy it and configure it on the cloud provider then we can provide those details in persistent volme manifest and yes, we are good to go. These two concepts persistent volume claims and persistent volume exist in Kubernetes to separate area of concerns of dev people and ops people. Administrator can buy storages handle secrets and provide them to a Kubernetes cluster via persistent volume and Developers can consume those storages using persistent volume claim. Hope this makes sense.

sbdtu5498: > Ah, I don't think we want to run Docker-in-Docker in Kubernetes. That's just a way to test the ability of the system to run jobs directly in Docker. So we can just remove the runner-docker service in the Kubernetes setup. The "native" runner also doesn't really make much sense in Kubernetes, we only need the `runner-k8s` and `runner-lava` ones in practice. Sorry if this wasn't clearly explained in the issue description.

Sure I will remove them.

kube/api-service.yaml
```diff
@@ -0,0 +1,17 @@
...
+      targetPort: 8000
+  type: ClusterIP
```
JenySadadia, line 17: I guess this can be eliminated as the default service type is `ClusterIP`.

kube/api-service.yaml
JenySadadia: > Oh, service port and Local port are not the same thing. So service port is just being used to accumulate the traffic from all the target pods' ports and provide stable endpoints to them because pods are too fragile. Then this traffic can be forwarded to the local host via Kubectl port-forward. I hope this explanation makes it a bit clear. Let me know anything else you would like to discuss.

Thanks for the clarification. Do you think Ingress or Gateway API can be useful here?

kube/api-service.yaml
sbdtu5498: Yes, this can be. I have included it to increase clarity.

kube/db-deployment.yaml
```diff
@@ -0,0 +1,31 @@
...
+          image: mongo:5.0
+          volumeMounts:
+            - name: mongo-data
+              mountPath: /data/db
+          # Resource limits needs to be discussed.
+          resources:
+  volumes:
+    - name: mongo-data
+      persistentVolumeClaim:
+        claimName: mongo-data-pvc
```
JenySadadia, line 31: Please fix the indentation here. `volumes` should be a sub-field of `spec`.

kube/redis-deployment.yaml
```diff
@@ -0,0 +1,31 @@
...
+          image: redis:6.2
+          volumeMounts:
+            - name: redis-data
+              mountPath: /data
+          # Resource limits needs to be discussed.
+          resources:
+  volumes:
+    - name: redis-data
+      hostPath:
+        path: home/docker/kernelci-api/redis/data
```
JenySadadia, line 31: Same comment about indentation.

gctucker:
@sbdtu5498 Hello, we now have solutions in place for the other parts of the API for the Early Access deployment with MongoDB in Atlas and storage in Azure Files.  Being able to deploy the main API service and Redis in k8s (and also MongoDB, SSH and Nginx as part of a reference deployment) are the next things we need to complete and reach this milestone.  As we're aiming to have this Early Access deployment ready by the end of August, we need to resolve this k8s reference deployment this week in order to not be late.  How does that sound?  If needed we can discuss it on Thursday in the weekly meeting and Open Hours.

kube/redis-deployment.yaml
sbdtu5498: They are part of the spec already. I think you confused it with the template spec which is a different thing.

kube/db-deployment.yaml
sbdtu5498: Same for this one as well.

kube/redis-deployment.yaml
JenySadadia: I am also talking about `spec.template.volumes` field.
Please check the indentation of `volumes` in `api-deployment.yaml` file. That is the correct one.
It should be the same for other deployment files.

kube/db-deployment.yaml
sbdtu5498: Ah I see, thanks for noticing it. There was no YAML indenter so I missed it. Extremely sorry. I will fix it.

kube/api-service.yaml
sbdtu5498: Oh sorry, I just saw your comment. Gateway API sure is very useful, but for basic traffic management, ingress would be sufficient enough. And yes we would be setting up an nginx ingress. Also, Gateway API is the unification of all the networking constructs that is possible in Infra management so if our use case is something like using service meshes and a bit more complicated, then we can also use Gateway API or service mesh. GatewayAPI has probably matured enough to be used in production so I guess going with any of the following i.e. service mesh or gateway API wouldn't be a bad choice.

kube/redis-deployment.yaml
sbdtu5498: Will fix it. It was unnoticeable as I was fixing the indent manually in VS code using ctrl + ] so probably missed this. Thanks for pointing it out.

JenySadadia:
Hello @sbdtu5498 
I followed steps of `README.md` to test it locally.

Here is what I received when I tried to `port-forward`:
```
$ kubectl port-forward svc/kernelci-api 8001:8000
Forwarding from 127.0.0.1:8001 -> 8000
Forwarding from [::1]:8001 -> 8000
Handling connection for 8001
Handling connection for 8001
E0801 15:32:33.409232   43541 portforward.go:409] an error occurred forwarding 8001 -> 8000: error forwarding port 8000 to pod 93a0f50c07bbca18636ae94f8b34f772f5ce7026e97de4a305e48a10aceed278, uid : exit status 1: 2023/08/01 10:02:33 socat[21728] E connect(5, AF=2 127.0.0.1:8000, 16): Connection refused
E0801 15:32:33.409463   43541 portforward.go:409] an error occurred forwarding 8001 -> 8000: error forwarding port 8000 to pod 93a0f50c07bbca18636ae94f8b34f772f5ce7026e97de4a305e48a10aceed278, uid : exit status 1: 2023/08/01 10:02:33 socat[21729] E connect(5, AF=2 127.0.0.1:8000, 16): Connection refused
error: lost connection to pod
```

```
$ kubectl get pods
NAME                                    READY   STATUS             RESTARTS          AGE
api-66f5dbcd4c-vmnq2                    0/1     ImagePullBackOff   0                 27d
db-687fcbdf9f-hfdp7                     1/1     Running            2 (12m ago)       27d
git-clone-pod                           0/1     CrashLoopBackOff   6 (2m19s ago)     8m20s
github-cloning-job-d8wrn                0/1     Error              0                 8m28s
github-cloning-job-rns97                0/1     Error              0                 8m36s
github-cloning-pod                      0/1     CrashLoopBackOff   125 (5m17s ago)   4d22h
kernelci-api-66d74576dc-nmzxx           1/1     Running            0                 112s
kernelci-api-db-565d86c46b-62wqh        1/1     Running            0                 112s
kernelci-api-redis-5dc9b64bfd-9fl7t     1/1     Running            0                 111s
kernelci-api-ssh-78d74d59f7-4zbbq       1/1     Running            0                 111s
kernelci-api-storage-69b9bd9568-b5qtd   1/1     Running            0                 111s
```
```
$ kubectl get deployments
NAME                   READY   UP-TO-DATE   AVAILABLE   AGE
kernelci-api           1/1     1            1           4d21h
kernelci-api-db        1/1     1            1           4m21s
kernelci-api-redis     1/1     1            1           4m21s
kernelci-api-ssh       1/1     1            1           4d22h
kernelci-api-storage   1/1     1            1           4m20s
```
```
$ kubectl get svc
NAME                   TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)    AGE
kernelci-api           ClusterIP   10.98.188.192    <none>        8000/TCP   4d22h
kernelci-api-ssh       ClusterIP   10.109.194.222   <none>        22/TCP     4d22h
kernelci-api-storage   ClusterIP   10.98.233.67     <none>        80/TCP     4d22h
kubernetes             ClusterIP   10.96.0.1        <none>        443/TCP    27d
```

gctucker:
@JenySadadia Have you looked into the `redis` hostname issue?

JenySadadia:
> @JenySadadia Have you looked into the `redis` hostname issue?

I created `redis` service and used the service name to connect from `api`. It didn't work. I'll look into it again later.

JenySadadia:
@sbdtu5498 
`init-pod` job is still failing.
```
$ kubectl logs git-clone-pod
fatal: destination path '/data' already exists and is not an empty directory.
```

sbdtu5498:
> @sbdtu5498 `init-pod` job is still failing.
> 
> ```
> $ kubectl logs git-clone-pod
> fatal: destination path '/data' already exists and is not an empty directory.
> ```

Here the pod is failing because your probably reinitialized it. Once it is initialized, the data would be there and thus reinitialization would require some sort of cleaning mechanism. You can get inside minikube via minikube ssh and then you would be in the home directory. Where you would be able to clean the existing data folder using sudo rm -rf /data, Then when you will run this command it will work fine and fetch the latest instance of github repo.

sbdtu5498:
> > @sbdtu5498 `init-pod` job is still failing.
> > ```
> > $ kubectl logs git-clone-pod
> > fatal: destination path '/data' already exists and is not an empty directory.
> > ```
> 
> Here the pod is failing because your probably reinitialized it. Once it is initialized, the data would be there and thus reinitialization would require some sort of cleaning mechanism. You can get inside minikube via minikube ssh and then you would be in the home directory. Where you would be able to clean the existing data folder using sudo rm -rf /data, Then when you will run this command it will work fine and fetch the latest instance of github repo.

Yes this is the persisting problem that I was talking about. It is persisting because the current networking mechanism is hardcoded for docker and it will require some work around. To get it fixed we probably need to expose the official redis port in the deployment file first, then create a svc for the corresponding deployment and use the DNS of that svc for networking. That would fix the issue. I will try to fix it before tomorrow's meeting. Meanwhile if you can follow this approach and take a look it'd be great.

sbdtu5498:
> Hello @sbdtu5498 I followed steps of `README.md` to test it locally.
> 
> Here is what I received when I tried to `port-forward`:
> 
> ```
> $ kubectl port-forward svc/kernelci-api 8001:8000
> Forwarding from 127.0.0.1:8001 -> 8000
> Forwarding from [::1]:8001 -> 8000
> Handling connection for 8001
> Handling connection for 8001
> E0801 15:32:33.409232   43541 portforward.go:409] an error occurred forwarding 8001 -> 8000: error forwarding port 8000 to pod 93a0f50c07bbca18636ae94f8b34f772f5ce7026e97de4a305e48a10aceed278, uid : exit status 1: 2023/08/01 10:02:33 socat[21728] E connect(5, AF=2 127.0.0.1:8000, 16): Connection refused
> E0801 15:32:33.409463   43541 portforward.go:409] an error occurred forwarding 8001 -> 8000: error forwarding port 8000 to pod 93a0f50c07bbca18636ae94f8b34f772f5ce7026e97de4a305e48a10aceed278, uid : exit status 1: 2023/08/01 10:02:33 socat[21729] E connect(5, AF=2 127.0.0.1:8000, 16): Connection refused
> error: lost connection to pod
> ```
> 
> ```
> $ kubectl get pods
> NAME                                    READY   STATUS             RESTARTS          AGE
> git-clone-pod                           0/1     CrashLoopBackOff   6 (2m19s ago)     8m20s
> github-cloning-job-d8wrn                0/1     Error              0                 8m28s
> github-cloning-job-rns97                0/1     Error              0                 8m36s
> github-cloning-pod                      0/1     CrashLoopBackOff   125 (5m17s ago)   4d22h
> kernelci-api-66d74576dc-nmzxx           1/1     Running            0                 112s
> kernelci-api-db-565d86c46b-62wqh        1/1     Running            0                 112s
> kernelci-api-redis-5dc9b64bfd-9fl7t     1/1     Running            0                 111s
> kernelci-api-ssh-78d74d59f7-4zbbq       1/1     Running            0                 111s
> kernelci-api-storage-69b9bd9568-b5qtd   1/1     Running            0                 111s
> ```
> 
> ```
> $ kubectl get deployments
> NAME                   READY   UP-TO-DATE   AVAILABLE   AGE
> kernelci-api           1/1     1            1           4d21h
> kernelci-api-db        1/1     1            1           4m21s
> kernelci-api-redis     1/1     1            1           4m21s
> kernelci-api-ssh       1/1     1            1           4d22h
> kernelci-api-storage   1/1     1            1           4m20s
> ```
> 
> ```
> $ kubectl get svc
> NAME                   TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)    AGE
> kernelci-api           ClusterIP   10.98.188.192    <none>        8000/TCP   4d22h
> kernelci-api-ssh       ClusterIP   10.109.194.222   <none>        22/TCP     4d22h
> kernelci-api-storage   ClusterIP   10.98.233.67     <none>        80/TCP     4d22h
> kubernetes             ClusterIP   10.96.0.1        <none>        443/TCP    27d
> ```


@sbdtu5498 init-pod job is still failing.

$ kubectl logs git-clone-pod
fatal: destination path '/data' already exists and is not an empty directory.
Here the pod is failing because your probably reinitialized it. Once it is initialized, the data would be there and thus reinitialization would require some sort of cleaning mechanism. You can get inside minikube via minikube ssh and then you would be in the home directory. Where you would be able to clean the existing data folder using sudo rm -rf /data, Then when you will run this command it will work fine and fetch the latest instance of github repo.

Yes this is the persisting problem that I was talking about. It is persisting because the current networking mechanism is hardcoded for docker and it will require some work around. To get it fixed we probably need to expose the official redis port in the deployment file first, then create a svc for the corresponding deployment and use the DNS of that svc for networking. That would fix the issue. I will try to fix it before tomorrow's meeting. Meanwhile if you can follow this approach and take a look it'd be great.

To get more insight in the issue use kubectl logs <pod name corresponding to API deployment> and you'd be able to see the issue is persisting precisely because of the given reason.

kube/minkube/deployments/redis-deployment.yaml
```diff
@@ -0,0 +1,31 @@
...
+          image: redis:6.2
+          volumeMounts:
+            - name: redis-data
+              mountPath: /data
+          # Resource limits needs to be discussed.
+          resources:
+      volumes:
+        - name: redis-data
+          hostPath:
+            path: /home/docker/kernelci-api/redis/data
```
JenySadadia, line 31: Mount volume `data` directory for `redis` is located inside `kernelci-api/docker/`. 
```suggestion
            path: /home/docker/kernelci-api/docker/redis/data
```

JenySadadia:
> > @sbdtu5498 `init-pod` job is still failing.
> > ```
> > $ kubectl logs git-clone-pod
> > fatal: destination path '/data' already exists and is not an empty directory.
> > ```
> 
> Here the pod is failing because your probably reinitialized it. Once it is initialized, the data would be there and thus reinitialization would require some sort of cleaning mechanism. You can get inside minikube via minikube ssh and then you would be in the home directory. Where you would be able to clean the existing data folder using sudo rm -rf /data, Then when you will run this command it will work fine and fetch the latest instance of github repo.

Okay. Thanks. Yes that worked.

JenySadadia:
So the `api` service worked after I created `redis` service and used `redis.default.svc.cluster.local` as a redis hostname from `api.pubsub.py`.
```
$ kubectl logs kernelci-api-66d74576dc-c9hr2
INFO:     Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)
INFO:     Started reloader process [1] using watchgod
INFO:     Started server process [8]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
```
After the `port-forward`, it serves the requests at 8001:
```
$ curl http://localhost:8001/latest/
{"message":"KernelCI API"}
```

JenySadadia:
Please add `redis-service.yaml`:
```
apiVersion: v1
kind: Service
metadata:
  name: redis
spec:
  selector:
    app: redis
  ports:
    - protocol: TCP
      port: 6379
      targetPort: 6379
  type: ClusterIP
```

JenySadadia:
We need to create `db-service.yaml` and update `api.db` to use `db.default.svc.cluster.local` as DB hostname.
```
apiVersion: v1
kind: Service
metadata:
  name: db
spec:
  selector:
    app: kernelci-api-db
  ports:
    - protocol: TCP
      port: 27017
      targetPort: 27017
  type: ClusterIP
```

gctucker:
> How about renaming all the services and deployments to drop `kernelci-api-` prefix?

I think we still need a way to specify the Redis host anyway, so other deployments that don't call it `redis` or use an external instance will be able to use it.

Having different names gives the advantage of distinguishing the k8s deployment from the docker-compose one.  I think it's probably fine to leave the names as @sbdtu5498 defined them with the prefix.

kube/minkube/services/api-service.yaml
```diff
@@ -0,0 +1,17 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: kernelci-api
+spec:
+  selector:
+    app: kernelci-api
+  ports:
+    - protocol: TCP
+      port: 8000
```
JenySadadia, line 15:
```suggestion
      port: 8001
```

kube/minkube/services/ssh-service.yaml
```diff
@@ -0,0 +1,17 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: kernelci-api-ssh
+spec:
+  selector:
+    app: kernelci-api-ssh
+  ports:
+    - protocol: TCP
+      port: 22
```
JenySadadia, line 15:
```suggestion
      port: 8022
```

kube/minkube/services/storage-service.yaml
```diff
@@ -0,0 +1,17 @@
...
+apiVersion: v1
+kind: Service
+metadata:
+  name: kernelci-api-storage
+spec:
+  selector:
+    app: kernelci-api-storage
+  ports:
+    - protocol: TCP
+      port: 80
```
JenySadadia, line 15:
```suggestion
      port: 8002
```

JenySadadia (changes requested):
So when I tried to access `api` container from `db` service, it fails with port `8001`:
```
root@kernelci-api-db-565d86c46b-dsm9d:/# curl 'http://kernelci-api:8001/latest/nodes'
curl: (28) Failed to connect to kernelci-api port 8001: Connection timed out
```
But it worked when I tried with port `8000`:
```
root@kernelci-api-db-565d86c46b-dsm9d:/# curl 'http://kernelci-api:8000/latest/nodes'
{"items":[],"total":0,"limit":50,"offset":0}
```
From outside the cluster with port forwarding:
```
curl 'http://localhost:8001/latest/nodes'
{"items":[],"total":0,"limit":50,"offset":0}
```

So port forwarding will work for outside the cluster but within the cluster service will be exposed on `ports.port`:
```
ports:
    - protocol: TCP
      port: 8000
      targetPort: 8000
```
So, I'd suggest updating `ports.port` for all the services.

kube/minkube/services/ssh-service.yaml
gctucker: I think using the standard ports makes sense i.e. 22 for SSH, then when exposing the port to the public it can be a different one depending on the deployment specifics

kube/minkube/services/storage-service.yaml
gctucker: Ditto - I think this should be 80

kube/minkube/services/ssh-service.yaml
JenySadadia: This is to expose the container port for the pipeline.

kube/minkube/deployments/redis-deployment.yaml
sbdtu5498: I just checked, there is no docker/redis folder. And the redis/data folder is present in the node. Please check.

kube/minkube/services/storage-service.yaml
sbdtu5498: Done, @gctucker I think it makes more sense to do what @JenySadadia is suggesting as they are just dummy placeholder ports, so the actual thing is still intact with the container, so I have made the changes accordingly. It has somewhat increased the visibility as well.

kube/minkube/deployments/redis-deployment.yaml
JenySadadia: @sbdtu5498 Yes, that is true that `docker/redis` is not present in the git source tree. But please check `docker-compose.yaml`.
```
volumes:
      - './docker/redis/data:/data'
```
So when `redis` service is up, it will create a directory `docker/redis/data` just like `ssh` and `storage` services.

kube/minkube/deployments/db-deployment.yaml
```diff
@@ -0,0 +1,33 @@
...
+            - name: mongo-data
+              mountPath: /data/db
+          # Resource limits need to be discussed.
+          resources:
+          ports:
+            - containerPort: 27017  # Expose MongoDB standard port
+      volumes:
+        - name: mongo-data
+          persistentVolumeClaim:
+            claimName: mongo-data-pvc
```
JenySadadia, line 33: Please add a new line at the end of the file.

JenySadadia:
I have added commits to fix a couple of minor things.
I think It is good to be merged now. @gctucker

kube/minkube/deployments/db-deployment.yaml
JenySadadia: I have added a commit to fix it.

kube/minkube/deployments/redis-deployment.yaml
JenySadadia: I have added a commit to fix it.

gctucker:
@JenySadadia Could you please review again and approve this PR? It's still marked with changes required which isn't ideal when merging a PR.

JenySadadia (approved):

PR #292, opened 2023-07-06
Add user profile model

JenySadadia:
Create a separate model to store user information that will be accessible publicly and can be updated by user. Store private information along with profile to `User` model and only admins will be allowed to access/update it.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -388,7 +388,7 @@ async def post_node(node: Node, current_user: str = Depends(get_user)):
             )
 
     await _verify_user_group_existence(node.user_groups)
-    node.owner = current_user.username
+    node.owner = current_user.profile.username
     obj = await db.create(node)
     data = _get_node_event_data('created', obj)
     await pubsub.publish_cloudevent('node', data)
--- a/api/main.py
+++ b/api/main.py
@@ -36,6 +36,7 @@ from .models import (
     Regression,
     User,
     UserGroup,
+    UserProfile,
     Password,
     get_model_from_kind
 )
@@ -157,11 +158,11 @@ async def post_user(
                         detail=f"User group does not exist with name: \
 {group_name}")
                 group_obj.append(group)
-        obj = await db.create(User(
-                                username=username,
-                                hashed_password=hashed_password,
-                                groups=group_obj
-                                ))
+        profile = UserProfile(
+                    username=username,
+                    hashed_password=hashed_password,
+                    groups=group_obj)
+        obj = await db.create(User(profile=profile))
     except DuplicateKeyError as error:
         raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST,
--- a/api/main.py
+++ b/api/main.py
@@ -128,9 +128,9 @@ async def authorize_user(node_id: str, user: User = Depends(get_current_user)):
     # user groups will be allowed to update the node
     node_from_id = await db.find_by_id(Node, node_id)
     if node_from_id.owner:
-        if not user.username == node_from_id.owner:
+        if not user.profile.username == node_from_id.owner:
             if not any(group.name in node_from_id.user_groups
-                       for group in user.groups):
+                       for group in user.profile.groups):
                 raise HTTPException(
                     status_code=status.HTTP_401_UNAUTHORIZED,
                     detail="Unauthorized to complete the operation"
--- a/api/models.py
+++ b/api/models.py
@@ -117,23 +117,31 @@ class UserGroup(DatabaseModel):
         collection.create_index("name", unique=True)
 
 
-class User(DatabaseModel):
-    """API user model"""
+class UserProfile(BaseModel):
+    """API user profile model"""
     username: str
     hashed_password: str = Field(description="Hash of the plaintext password")
-    active: bool = Field(
-        default=True,
-        description="To check if user is active or not"
-    )
     groups: conlist(UserGroup, unique_items=True) = Field(
         default=[],
         description="A list of groups that user belongs to"
     )
 
+
+class User(DatabaseModel):
+    """API user model
+    The model will be accessible by admin users only"""
+    active: bool = Field(
+        default=True,
+        description="To check if user is active or not"
+    )
+    profile: UserProfile = Field(
+        description="User profile details accessible by all users"
+    )
+
     @classmethod
     def create_indexes(cls, collection):
         """Create an index to bind unique constraint to username"""
-        collection.create_index("username", unique=True)
+        collection.create_index("profile.username", unique=True)
 
 
 class KernelVersion(BaseModel):
--- a/api/admin.py
+++ b/api/admin.py
@@ -17,7 +17,7 @@ import getpass
 
 from .auth import Authentication
 from .db import Database
-from .models import User, UserGroup
+from .models import User, UserGroup, UserProfile
 
 
 async def setup_admin_group(db):
@@ -29,7 +29,8 @@ async def setup_admin_group(db):
 
 
 async def setup_admin_user(db, username, admin_group):
-    user_obj = await db.find_one(User, username=username)
+    user_obj = await db.find_one_by_attributes(User,
+                                               {'profile.username': username})
     if user_obj:
         print(f"User {username} already exists, aborting.")
         print(user_obj.json())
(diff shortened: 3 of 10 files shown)


api/main.py
```diff
@@ -138,11 +139,12 @@ async def post_user(
                         detail=f"User group does not exist with name: \
 {group_name}")
                 group_obj.append(group)
-        obj = await db.create(User(
+        profile = await db.create(UserProfile(
                                 username=username,
                                 hashed_password=hashed_password,
                                 groups=group_obj
                                 ))
+        obj = await db.create(User(profile=profile))
```
gctucker, line 147: Looks like the profile is created twice, first as a separate document and then again as nested in User.

gctucker (commented):
Let's get the basic user functionality merged and available first, then we can see how to deal with this as a follow-up.

api/main.py
JenySadadia: A profile is created just once. The already created profile document will be assigned to `User.profile`.
Please see the below documents from my local setup:
UserProfile `jeny`:
```
{ "_id" : ObjectId("64abdcc7ed84334468b3acbe"), "username" : "jeny", "hashed_password" : "$2b$12$QhsvplXElm34.HiF2rF6yO3WlYMLm1ZyDyo2hdtp3vDv0fdTjouqK", "groups" : [ { "_id" : ObjectId("64abe12e40679ecb42473768"), "name" : "kernelci" } ] }
```
User:
```
{ "_id" : ObjectId("64abdcc7ed84334468b3acbf"), "active" : true, "profile" : { "_id" : ObjectId("64abdcc7ed84334468b3acbe"), "username" : "jeny", "hashed_password" : "$2b$12$QhsvplXElm34.HiF2rF6yO3WlYMLm1ZyDyo2hdtp3vDv0fdTjouqK", "groups" : [ { "_id" : ObjectId("64abe12e40679ecb42473768"), "name" : "kernelci" } ] } }

```

gctucker: Yes, the UserProfile data is therefore created twice as "username" etc. is embedded in User.

I think we can just create a User document, then you can search using the User.profile.username attribute.

JenySadadia: This is the same as we are embedding user groups models in `UserProfile` model. We are just assigning user group objects while creating user profiles.

If we directly create a `User` model then it won't have any `UserProfile` model we can map to in the database.
```
{ "_id" : ObjectId("64afabd100848e93a69920ad"), "active" : true, "profile" : { "_id" : null, "username" : "jeny", "hashed_password" : "$2b$12$.x886aNDVsV1kUPEP0kOJ.FO29QYlrK1KJVjb7klciI4UtX28BKvS", "groups" : [ ] } }
```

gctucker: Yes but for groups, it makes sense to have a separate one as they're used in several places.  Each profile will only be used by a single user.

JenySadadia:
> @JenySadadia I think we can have the functionality in place to create user accounts and have node owners without an implementation for a user profile, which can be added later. So I would suggest to draw the line there, to have all the basic things implemented and completely done with just User and a user name, groups, password field and the corresponding API and CLI and documentation bits all aligned first. Then once we have this stable base, users can already start using it, and we can add support for profiles as an incremental step. How does that sound?

Yes, we can do that. However, since we already have a few PRs for user profile support, it won't be that much time-consuming to tweak them and merge them. If we add the profile later after the early access, we would need to migrate all the existing users and also revisit the changes from these PRs.

JenySadadia:
> My fear is that there might still be a couple of weeks of tweaks before this gets merged, which would delay the early access rollout. Avoiding this seems to me like a smaller impact than a migration later on.

How about I tweak all the related PRs today itself or latest by tomorrow? I don't think it would take up much time.

gctucker:
Alright let's try that.

JenySadadia:
Updated all existing users on staging instance to have `profile` field.

Tested OK on staging:
```
$ curl -X 'POST' 'https://staging.kernelci.org:9000/latest/user/test' -H 'accept: application/json'   -H 'Content-Type: application/json'  -H 'Authorization: Bearer TOKEN' -d '{"password": "****"}'
{"id":"64b15c1ad089ca2290f442c1","active":true,"profile":{"username":"test","hashed_password":"$2b$12$JnnfhTn1gn8SUaKV0f2H8.Ov6ZlFWuvgs5yyfMIYCfF3TVAe97.YK","groups":[]}}
```

api/models.py
```diff
@@ -117,23 +117,31 @@ def create_indexes(cls, collection):
         collection.create_index("name", unique=True)
 
 
-class User(DatabaseModel):
-    """API user model"""
+class UserProfile(BaseModel):
+    """API user profile model"""
     username: str
     hashed_password: str = Field(description="Hash of the plaintext password")
```
gctucker, line 123: I wonder if these 2 fields should rather be in the User model rather than User Profile.  I think the profile part hasn't been well defined yet.  The password will most likely require a dedicated API endpoint to update it, and maybe the username too.  In fact, even `groups` might be something to keep in `User` as it's not something the users should be able to update themselves.

gctucker (commented):
I can see a number of issues with user profiles which haven't been solved or even properly identified yet, and we don't really need user profiles for early access.  So I still think we should put this aside for now and keep it as a follow-up development activity.  We will need a way for users to change their passwords for early access and having it blocked because the user profile implementation is not ready is going to cause some problems.

JenySadadia:
> I can see a number of issues with user profiles which haven't been solved or even properly identified yet, and we don't really need user profiles for early access. So I still think we should put this aside for now and keep it as a follow-up development activity. We will need a way for users to change their passwords for early access and having it blocked because the user profile implementation is not ready is going to cause some problems.

The profiles can be updated using PUT `/user/profile/{username}` requests. Please see the PR https://github.com/kernelci/kernelci-api/pull/293 for this.
This provides the facility to update password as well as user groups. `username` once set can not be updated by this request. This is due to the fact that we bind tokens with usernames. If the user changes the username, the already existing token won't work.
What other issues do you feel I need to fix for user profiles?
Also, I agree that we do not need user profiles for early access. However, I have already created all the related PRs for it and also tested all the changes on staging. I feel that we are just a few fixes away to get them merged.

api/models.py
JenySadadia: I thought we agreed to move `username`, `password`, and `groups` to user profile. Only passwords should be hidden from the response when we get all user profiles.

gctucker:
A few things from the top of my head:
* users should not be able to change their groups themselves
* updating the password should be possible without an API token
* we might want to make `username` fixed initially, so it's unclear whether it should be a user profile field
* we might not need a model for storing the user profile in the database, only at the API level

gctucker:
We would need to go through this in more detail, these are just a few things I thought about while reviewing the code.

JenySadadia:
> A few things from the top of my head:
> 
> * users should not be able to change their groups themselves

Okay. And what about `Node.user_groups`? Should we allow users to set it while creating nodes?
> * updating the password should be possible without an API token

Yes, that's true. The PUT `/user/profile` is used to update a whole user profile. Thus, it needs a token for authentication.

> * we might want to make `username` fixed initially, so it's unclear whether it should be a user profile field

We can add another `name` field as you suggested earlier and make it a variable. Or we can allow `username` to get updated and notify the user to regenerate tokens.

> * we might not need a model for storing the user profile in the database, only at the API level

We are not storing `UserProfile` separately in the DB. It is being used in `User` model.

gctucker:
I'll reply to this in a more structured way on the GitHub issue.  I think we now need to look at a minimal implementation with a public profile set of attributes and then split out things we can leave for later.  Also it seems worth investigating if there's a library we can use directly with FastAPI to manage all the user / group side of things.

JenySadadia:
> I'll reply to this in a more structured way on the GitHub issue. I think we now need to look at a minimal implementation with a public profile set of attributes and then split out things we can leave for later. Also it seems worth investigating if there's a library we can use directly with FastAPI to manage all the user / group side of things.

This package provides a base user model, user manager, and basic endpoints. https://github.com/fastapi-users/fastapi-users

api/models.py
gctucker: Right, I guess that's not very clearly defined then.  One thing is about information users can read publicly, then what users can read from their own profile, then what can users update themselves.  Users shouldn't be able to escalate their privileges, and passwords shouldn't be readable at all (even encrypted as that facilitates brute-force attacks).

gctucker (commented):
As discussed, I think we can have this merged as-is and then rework things slightly to make it safe even if it means some features will be missing.  Then we can see if there's anything else that needs to be done for "early access" and update the backlog with things to do for a full production implementation (probably using an existing package, see #304).

gctucker (approved):

PR #293, opened 2023-07-10
Update user endpoint

JenySadadia:
Added PUT `/user/profile/<user-id>` endpoint to update user profiles.


gctucker:
@JenySadadia Please rebase

gctucker (approved):
The commit title seems to imply that the user profile model is being updated, rather than it's implementing an endpoint for updating user profile documents.  But that's a detail, just maybe something to keep in mind in the future to always try and be accurate in git commits.

PR #294, opened 2023-07-11
docker-compose.yaml: set MongoDB cache to 0.5 GB

gctucker:
Limit the MongoDB cache to 0.5 GB using the standard --wiredTigerCacheSizeGB command line argument.  The default value is to use 50% of the available RAM which is disproportionate for a regular development setup.


gctucker:
Tested OK on staging for several days.  This is required to keep the memory usage down.

PR #297, opened 2023-07-13
Update "Getting started" guide after setting up local environment

pawiecz:
Endpoint "/me" was replaced with "/whoami" in:
ef045973cf72cc70bda8736c73e8a62609275ad4

### DIFF
--- a/doc/getting-started.md
+++ b/doc/getting-started.md
@@ -125,7 +125,7 @@ For example, to check it's working:
 
 ```
 $ curl -X 'GET' \
-  'http://localhost:8001/latest/me' \
+  'http://localhost:8001/latest/whoami' \
   -H 'accept: application/json' \
   -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJib2IifQ.KHkILtsJaCmueOfFCj79HGr6kHamuZFdB1Yz_5GqcC4'
 {"_id":"615f30020eb7c3c6616e5ac3","username":"bob","hashed_password":"$2b$12$VtfVij6zz20F/Qr0Ri18O.11.0LJMMXyJxAJAHQbKU0jC96eo2fr.","active":true}
--- a/doc/getting-started.md
+++ b/doc/getting-started.md
@@ -128,7 +128,7 @@ $ curl -X 'GET' \
   'http://localhost:8001/latest/whoami' \
   -H 'accept: application/json' \
   -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJib2IifQ.KHkILtsJaCmueOfFCj79HGr6kHamuZFdB1Yz_5GqcC4'
-{"_id":"615f30020eb7c3c6616e5ac3","username":"bob","hashed_password":"$2b$12$VtfVij6zz20F/Qr0Ri18O.11.0LJMMXyJxAJAHQbKU0jC96eo2fr.","active":true}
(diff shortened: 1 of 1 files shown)


pawiecz:
(This PR will be updated with further changes)

doc/getting-started.md
```diff
@@ -123,7 +123,7 @@ For example, to check it's working:
 
 ```
 $ curl -X 'GET' \
-  'http://localhost:8001/latest/me' \
+  'http://localhost:8001/latest/whoami' \
   -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJib2IifQ.KHkILtsJaCmueOfFCj79HGr6kHamuZFdB1Yz_5GqcC4'
```
JenySadadia, line 129: Could you please update the sample command result as well? I should have fixed it as a part of https://github.com/kernelci/kernelci-api/pull/280 but I missed it. Since you are updating documentation for the same endpoint, I'd appreciate if you could add another commit with suggested change.
```suggestion
{"id":"615f30020eb7c3c6616e5ac3","username":"bob","hashed_password":"$2b$12$VtfVij6zz20F/Qr0Ri18O.11.0LJMMXyJxAJAHQbKU0jC96eo2fr.","active":true}
```

pawiecz: Sure thing! Thank you for pointing it out for me, I missed this change while setting up my local instance.

JenySadadia: Thank you.

pawiecz: Resolved in https://github.com/kernelci/kernelci-api/pull/297/commits/66dc4252cacb22e570002455c55f5e844f3f61bc

pawiecz:
There are still two subtasks which could be useful:

- updating raw `curl` requests to `kci` tool commands
- updating test set available by default

but I should not block other changes to the docs with this PR anymore.

This branch also applies without conflicts when rebased on [gctucker:doc-api-admin](https://github.com/gctucker/kernelci-api/tree/doc-api-admin) (I will update this PR when #312 is merged)

gctucker:
> * updating raw `curl` requests to `kci` tool commands

I think it might be a bit early to start relying on the `kci` tool as it's not distributed as a PyPI package yet so the syntax is subject to evolve, that's not a blocking issue.  Of course it's easier to use than raw `curl` commands, let's just hope it stays in sync with the code.

gctucker (commented):
LGTM - let's wait until this appears on the staging website to double check it looks as expected:
https://static.staging.kernelci.org/docs/api/getting-started/

doc/getting-started.md
```diff
@@ -207,6 +213,11 @@ file which provides environment variables for the Docker containers:
 echo "KCI_API_TOKEN=<your token>" >> .env
 ```
 
+> **Note** Admin API token created in the previous section can be used for
+> initial Pipeline instance testing. In production environment the best
+> practice would be to create a separate (non-admin) user with required scopes
+> and generate a new API token for that user.
+
```
JenySadadia, line 220: I don't think it would be a good idea to use API token with admin scope in the pipeline. The `admin` token should be used only in the API to perform admin-related activities like user and group management.

gctucker: Sure, but for a local developer setup with a single user it doesn't really matter.  It's just a bit better to have a plain user token to exercise the permission checks I guess.  It's a bit like Docker images with root as the default user, not great but with no consequence outside the container anyway.

JenySadadia: Yes, it won't be an issue for the local setup.

JenySadadia:
Tested OK on staging.

PR #301, opened 2023-07-17
Assign ownership details to a hierarchy of nodes

JenySadadia:
Depends on https://github.com/kernelci/kernelci-api/pull/292 for user profile changes.

api.main: set node ownership to a hierarchy of nodes
    
When submitting a hierarchy of nodes using `put_nodes`, set node ownership information to all the nodes recursively.
Also, use the `authorize_user` method for user authorization upon receiving the request.

### DIFF
--- a/api/admin.py
+++ b/api/admin.py
@@ -17,7 +17,7 @@ import getpass
 
 from .auth import Authentication
 from .db import Database
-from .models import User, UserGroup
+from .models import User, UserGroup, UserProfile
 
 
 async def setup_admin_group(db):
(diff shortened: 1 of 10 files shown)


gctucker (commented):
I don't think this depends on user profile changes, just the Node.owner field needs to be populated.  Which model it's using doesn't make a difference, as far as I can tell.

JenySadadia:
> I don't think this depends on user profile changes, just the Node.owner field needs to be populated. Which model it's using doesn't make a difference, as far as I can tell.

Please see below line of change. It needs to have `user.profile` field.
```
hierarchy.node.owner = user.profile.username
```

gctucker:
Ah yes, thanks.  Well if we don't have the UserProfile changes then we can just do `node.owner = user.username`.  But I see what you mean.

gctucker (approved):

PR #302, opened 2023-07-17
Update `Node.owner` to be a mandatory field

JenySadadia:
Remove `Optional` and make `Node.owner` a mandatory field. `Optional` was added as node ownership fields were not
present previously. Not marking the field `Optional` could have led to an issue for existing nodes on staging instance.
Update `authorize_user` method accordingly.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -127,14 +127,13 @@ async def authorize_user(node_id: str, user: User = Depends(get_current_user)):
     # Only the user that created the node or any other user from the permitted
     # user groups will be allowed to update the node
     node_from_id = await db.find_by_id(Node, node_id)
-    if node_from_id.owner:
-        if not user.profile.username == node_from_id.owner:
-            if not any(group.name in node_from_id.user_groups
-                       for group in user.profile.groups):
-                raise HTTPException(
-                    status_code=status.HTTP_401_UNAUTHORIZED,
-                    detail="Unauthorized to complete the operation"
-                )
+    if not user.profile.username == node_from_id.owner:
+        if not any(group.name in node_from_id.user_groups
+                   for group in user.groups):
+            raise HTTPException(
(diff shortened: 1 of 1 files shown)


gctucker:
I guess this one can be rebased if needed and the checks can be fixed now?  Also I believe you said we needed to handle it as a particular case since we may not know the actual owner when creating the Node object since it happens during validation and before running the API handler code.

JenySadadia:
> 

Yes, I need to fix it before we can merge.

JenySadadia:
I removed `Optional` from `Node.owner` and made it a required field. Received below error when tried to send POST node request without `owner` field:
```
{"detail":[{"loc":["body","owner"],"msg":"field required","type":"value_error.missing"}]}
```
We do not receive owner information from the client (pipeline). Also, for the sake of security, it is better to set the ownership information from the API side rather than the client side.
- I tried to use `default_factory` to set it internally when the request validation takes place. But it needs to access the current user and there is no way to pass the argument externally to the default factory function.
- FastAPI does not have `request_model_exclude` kind of a parameter as we have for the response model. This feature request has been raised here: https://github.com/tiangolo/fastapi/discussions/9058

Hence, for the time being, I kept the `owner` field as an `Optional`.

gctucker:
OK, yes the client side will only be sending a token which can then be used to find the user.  So for now we just have to deal with this at the API implementation level rather than in Pydantic / FastAPI.

gctucker:
@JenySadadia Are you planning to make any further changes to this particular PR?

JenySadadia:
> @JenySadadia Are you planning to make any further changes to this particular PR?
 
If you don't have any further suggestion, I am done with this PR.

gctucker (approved):
OK I think it's ready to be merged then.  We'll take another look at the whole user profile implementation and tweak anything if needed as a follow-up to be ready for the Early Access phase.

PR #305, opened 2023-07-18
docker/api/requirements: fix `api` container build failure

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/303

Fix build failures of `api` container. Error logs can be found at ```https://github.com/kernelci/kernelci-api/issues/303```.

Root cause:
`fastapi[all] 0.68.1` package installs `pyyaml 5.4.1` package internally. That again depends on `CPython` package. Due to the recent upgrade to `CPython 3.0.0`, `pyyaml 5.4.1` stopped building with `python 3.10`. To fix this, upgrading `fastapi[all]` to `0.99.1` version as that uses `pyyaml 6.0.1`. The selected `fastapi` version is compatible with current `fastapi-pagination` and `fastapi-versioning` packages.

PR #306, opened 2023-07-19
docker/api: pin PyYAML version to 5.3.1

gctucker:
The version of PyYAML was not being specified and newer releases rely on more recent dependencies which cascade into basically updating FastAPI and some parts of the code accordingly.  While doing this would make sense to keep the API codebase up-to-date, pin down the PyYAML version to 5.3.1 for now in requirements.txt so we avoid the whole issue in the short term.

Fixes: https://github.com/kernelci/kernelci-api/issues/303

### DIFF
--- a/docker/api/requirements.txt
+++ b/docker/api/requirements.txt
@@ -8,4 +8,5 @@ python-jose[cryptography]==3.3.0
 uvicorn[standard]==0.13.4
 motor==2.5.1
 pymongo-migrate==0.11.0
+pyyaml==5.3.1
 fastapi-versioning==0.10.0


gctucker:
Essentially, the `pyyaml` package just depends loosely on `cython` without specifying a version:
https://github.com/yaml/pyyaml/blob/master/pyproject.toml
```toml
[build-system]
requires = ["setuptools", "wheel", "Cython"]
build-backend = "setuptools.build_meta"
```
In this instance, `cython` got upgraded implicitly and that required also a cascading upgrade of PyYAML, Pydantic and FastAPI with some code changes as per #305 (which pins down `pydantic` to simplify things half-way through).

JenySadadia (approved):

PR #311, opened 2023-07-24
api.admin: add --admin-group command line argument

gctucker:
Add an --admin-group command line argument to let the user override the admin group name which is "admin" by default.  The code base still implicitly depends on the magic "admin" group name for admin user permissions so in practice this shouldn't need to be used other than for development purposes right now.


JenySadadia (approved):
LGTM

JenySadadia:
Tested OK on staging.

PR #312, opened 2023-07-24
Doc: use `api.admin`

gctucker:
Update the Getting Started documentation page to use the new `api.admin` module and delete `init_user_groups` as it's not used any more.


doc/getting-started.md
```diff
@@ -74,27 +74,27 @@ above with the root `/` endpoint and most `GET` requests to retrieve data.
...
+Password for user 'admin':
+Creating admin user...
 ```
 
+> **Note** Strictly speaking, only the `db` service needs to be running in
+> order to use this tool.  In fact it can also be used with any other MongoDB
+> instance such as an Atlas account using the `--mongo` command line argument.
+
 > **Note** For more details about how to create users via the raw API, see the
 > [API documentation](/docs/api/api-details/#users)
```
JenySadadia, line 99: We also need to update https://kernelci.org/docs/api/api-details/#users section to reflect `User` schema changes in the sample `curl` command responses. I'll create a GH issue for that.

JenySadadia (approved):
LGTM.
Let's wait until it gets reflected on the staging doc.

doc/getting-started.md
gctucker: Ah yes, thanks.  That's for another PR though as it's not directly related to the tools for bootstrapping a database with an admin user.

gctucker:
OK it's there now: https://static.staging.kernelci.org/docs/api/getting-started/#create-an-admin-user-account

For some reason it got stuck, it should have updated itself last night automatically.  Investigating...

doc/getting-started.md
JenySadadia: Yes, that is not related to this PR. Please see https://github.com/kernelci/kernelci-api/issues/313

JenySadadia:
Looks good on staging: https://static.staging.kernelci.org/docs/api/getting-started/#create-an-admin-user-account

PR #314, opened 2023-07-25
api.main: fix user.profiles.groups in authorize_user()

gctucker:
Fix how the authorize_user() function is checking if the node's group matches the user's groups as they're now in user.profile.groups.

Fixes: 73c1d5797653 ("api.main: drop a check for node owner in `authorize_user`")

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -129,7 +129,7 @@ async def authorize_user(node_id: str, user: User = Depends(get_current_user)):
     node_from_id = await db.find_by_id(Node, node_id)
     if not user.profile.username == node_from_id.owner:
         if not any(group.name in node_from_id.user_groups
-                   for group in user.groups):
+                   for group in user.profile.groups):
             raise HTTPException(
                 status_code=status.HTTP_401_UNAUTHORIZED,
                 detail="Unauthorized to complete the operation"


gctucker:
@JenySadadia By the way I also see an issue here:

```python
    if not user.profile.username == node_from_id.owner:
        if not any(group.name in node_from_id.user_groups
                   for group in user.profile.groups):
            raise HTTPException(
                status_code=status.HTTP_401_UNAUTHORIZED,
                detail="Unauthorized to complete the operation"
            )
    return user
```

If the node owner doesn't match the user then no exception is raised and the user is authorized.

JenySadadia:
> Fix how the authorize_user() function is checking if the node's group matches the user's groups as they're now in user.profile.groups.
> 
> Fixes: [73c1d57](https://github.com/kernelci/kernelci-api/commit/73c1d5797653b6fc188539999cff895220a4cc6e) ("api.main: drop a check for node owner in `authorize_user`")

Thanks for the fix.

JenySadadia:
> @JenySadadia By the way I also see an issue here:
> 
> ```python
>     if not user.profile.username == node_from_id.owner:
>         if not any(group.name in node_from_id.user_groups
>                    for group in user.profile.groups):
>             raise HTTPException(
>                 status_code=status.HTTP_401_UNAUTHORIZED,
>                 detail="Unauthorized to complete the operation"
>             )
>     return user
> ```
> 
> If the node owner doesn't match the user then no exception is raised and the user is authorized.

If the node owner doesn't match the user, it will check if the user belongs to the authorized group. If not, it will raise exception 401 Unauthorized.

gctucker:
> If the node owner doesn't match the user, it will check if the user belongs to the authorized group. If not, it will raise exception 401 Unauthorized.

I'm pretty sure I saw it authorize a user when it shouldn't, I believe that was with the LAVA callback user.  But it might have been another issue, and I've changed the LAVA callback token to be the same as the pipeline services now while the user management implementation is still moving.

At least I think the first line is a bit confusing, this would be easier to read imho:
```python
if user.profile.username != node_from_id.owner:
```
Maybe I misread it as `user.profile.username == node_from_id.owner` and missed the `not`.  I'll keep an eye on this, we'll see how things go and probably review the whole user / profile features when we start adding users and groups in case there are still any glitches.

JenySadadia:
> > If the node owner doesn't match the user, it will check if the user belongs to the authorized group. If not, it will raise exception 401 Unauthorized.
> 
> I'm pretty sure I saw it authorize a user when it shouldn't, I believe that was with the LAVA callback user. But it might have been another issue, and I've changed the LAVA callback token to be the same as the pipeline services now while the user management implementation is still moving.
> 
> At least I think the first line is a bit confusing, this would be easier to read imho:
> 
> ```python
> if user.profile.username != node_from_id.owner:
> ```
> 
Well yes, that is easier to read.

> Maybe I misread it as `user.profile.username == node_from_id.owner` and missed the `not`. I'll keep an eye on this, we'll see how things go and probably review the whole user / profile features when we start adding users and groups in case there are still any glitches.

Okay, sure.

PR #325, opened 2023-08-14
k8s: script to setup admin user

JenySadadia:
Added a script `kube/minikube/scripts/setup_admin_user.sh` to setup an admin user for k8s API instance.
At first, the script will get the pod name of `kernelci-api` service running on Minikube. Then it will run `api.admin` module inside the pod container to create an admin user.


JenySadadia:
After creating the admin user using the script, it can be used to get `KCI_API_TOKEN` for pipeline instance.

JenySadadia:
Rebased the PR.

gctucker (approved):
OK let's assume this is fine for now, then I think it'll be specific to MiniKube as for AKS we might do something different since the Atlas database is publicly available.

PR #328, opened 2023-08-29
pyproject.toml: add initial file

gctucker:
Add an initial pyproject.toml file to be able to build and install the
    api Python package.
    
    Thanks to Nikolay Yurin for finding the bug about the license file:
    https://github.com/pypa/pip/issues/12251
    
    This project uses a standard license so it's better to just rely on
    the LGPL-2.1-or-later SPDX indentifier instead of providing the path
    to the whole license file.
    
    Development dependencies have been omitted and all the required
    dependencies have a fixed version in order to build reproducible
    Docker images for production deployment.  Some follow-up changes can
    be made to have a better way to handle dependencies for development.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,34 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2023 Collabora Limited
+# Author: Guillaume Tucker <guillaume.tucker@collabora.com>
+
+[project]
+name = "kernelci-api"
+version = "0"
+description = "KernelCI API"
+readme = "README.md"
+requires-python = ">=3.10"
+license = {text = "LGPL-2.1-or-later"}
+dependencies = [
+  "aioredis[hiredis] == 2.0.0",
+  "cloudevents == 1.9.0",
+  "fastapi[all] == 0.68.1",
+  "fastapi-pagination == 0.9.3",
+  "passlib == 1.7.4",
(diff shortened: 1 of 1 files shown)


gctucker:
Thanks @yurinnick for finding the bug about the license file.  I've now updated #329 accordingly to not install dependencies from `requirements.txt` but only rely on `pyproject.toml`.

gctucker:
Tested OK with the early access deployment.

yurinnick:
You can add optional dependencies as well:

```
[project.optional-dependencies]
dev = [
  "pycodestyle ~= 2.8.0",
  "pylint ~= 2.12.2"
]
test = [
  "fakeredis ~= 1.7.0",
  "pytest ~= 6.2.5",
  "pytest-asyncio ~= 0.16.0",
  "pytest-dependency ~= 0.5.1",
  "pytest-mock ~= 3.6.1",
  "pytest-order ~= 1.0.1",
  "httpx ~= 0.23.3",
]
```

And install it with `pip install -e .[dev,test]`

yurinnick:
Just to be on the same page, these dependencies won't go into production image, unless `dev` and/or `test` groups  are explicitly specified during installation.

gctucker:
Yes, I know that too.  But I don't think we should just add them like that and keep the required dependencies pinned down as it's not convenient for development.

yurinnick:
Got it! So you want to keep to unpinned and flexible for development, but pinned to production Docker image build.

PR #329, opened 2023-08-29
docker/api: add Dockerfile.production

gctucker:
Add Dockerfile.production to build the kernelci/api Docker image for production deployment.  It installs the "api" package using pip directly in the image so it's self-contained.  This relies on pyproject.toml to be set up correctly.

This is an alternative to the standard Dockerfile used by docker-compose for a local deployment based directly on the source tree.

Depends on https://github.com/kernelci/kernelci-api/pull/328


gctucker:
Tested OK with the early access deployment.

PR #331, opened 2023-08-29
Add AKS deployment

gctucker:
Add the YAML files and README to create a deployment of the KernelCI API in AKS.


gctucker:
Tested OK with the AKS early-access deployment.

PR #332, opened 2023-08-29
kube: rename minkube as minikube

gctucker:
This is about MiniKube so let's fix the typo in the directory name.

Fixes: 3425d17baa1c ("adding kubernetes manifest to test kernelci-api minikube deployment")


nuclearcat:
I suggest to hold it a bit, maybe i can add patches that it will make this solution as independent self-deploy option that can work on other kubernetes clusters, not only minikube, so it can be renamed to something selfdeploy-k8s or something

gctucker:
No I think we need to fix this and then look for maybe consolidating things.  That will be needed in the next couple of months for the production deployment milestone, for now we need to wrap up the early access stuff.

PR #333, opened 2023-08-29
kube/minikube: fix Sanskar Bhushan copyright statements

gctucker:
Fix the copyright statements for Sanskar Bhushan to follow the convention of listing the copyright owners directly rather than referring to the "authors".  In many cases, the authors don't own the copyright if they're writing the code as part of their work with an employer.

### DIFF
--- a/kube/minkube/init/init-pod.yaml
+++ b/kube/minkube/init/init-pod.yaml
@@ -1,7 +1,6 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 #
-# Copyright (C) 2023 KernelCI Authors.
-# Author: Sanskar Bhushan <sbdtu5498@gmail.com>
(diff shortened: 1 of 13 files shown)


nuclearcat (approved):
LGTM

kube/minkube/init/init-pod.yaml
```diff
@@ -16,9 +15,8 @@ spec:
     - name: data-volume
       mountPath: /data
     # resource limits needs to be discussed
-    resources: 
+    resources:
```
nuclearcat, line 18: I think it is not related to copyright fixes and can be separate commit?

gctucker: No that's just whitespace automatically removed.  It's normal for trailing whitespace to cause git noise.

PR #339, opened 2023-09-01
Update API overview page

gctucker:
Update the API overview page with more information and add an embedded Atlas graph with the number of nodes.

### DIFF
--- a/doc/_index.md
+++ b/doc/_index.md
@@ -1,6 +1,6 @@
 ---
 title: "API & Pipeline"
-date: 2023-08-13
+date: 2023-09-01
 description: "KernelCI API and Pipeline"
 weight: 2
 ---
@@ -28,6 +28,12 @@ results.  There's also a result value, to tell whether the related pipeline
 step that produced the node passed or failed.  Finally there's a list of
 artifacts with URLs to know where to find all the related files (binaries,
 logs, generated results etc.).
+
+Here's a publicly available
+[graph](https://charts.mongodb.com/charts-kernelci-api-staging-otxuy/public/dashboards/5b52c3c6-81bb-4658-a5f3-8ebd2e980436#)
+showing the number of nodes added to the database every day:
+
+<div style="padding: 0 0 20px 0"><iframe style="background: #FFFFFF;border: none;border-radius: 2px;box-shadow: 0 2px 10px 0 rgba(70, 76, 79, .2);" width="640" height="480" src="https://charts.mongodb.com/charts-kernelci-api-staging-otxuy/embed/charts?id=64c9fa16-ba81-40e9-8cd7-cd56fb88fbf3&maxDataAge=3600&theme=light&autoRefresh=true"></iframe></div>
 
 > **Note:** The API doesn't manage storage, the only requirement is to provide
 > publicly-available HTTP(S) URLs for each artifact.
--- a/doc/_index.md
+++ b/doc/_index.md
@@ -1,13 +1,80 @@
 ---
 title: "API & Pipeline"
-date: 2022-03-04
-draft: true
-description: "KernelCI API and Modular Pipeline"
-weight: 4
+date: 2023-08-13
+description: "KernelCI API and Pipeline"
+weight: 2
 ---
 
-This section covers the new KernelCI API as well as the initial pipeline using
-it.
+## API Overview
+
+The KernelCI API is a server-side service which provides two main features: a
+database abstraction and a publisher / subscriber interface.  Another important
+concept is the fact that users own the data they send to the API.  Let's have a
+quick look at how this all fits together.
+
+### Database Abstraction
+
+All the data managed by KernelCI is stored in a MongoDB database using node
+objects.  These can contain data about any part of the testing hierarchy such
+as a kernel revision, a build, static test results, runtime functional tests,
+regressions etc.  Each node has a parent so they form a tree
+([DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph)).  There is
+typically one root node for each kernel revision with lots of child nodes
+containing all the test data that relates to it.
+
+Each node object also has a state which can be used when orchestrating the
+pipeline.  For example, a node will be in the Running state while awaiting some
+results.  There's also a result value, to tell whether the related pipeline
+step that produced the node passed or failed.  Finally there's a list of
+artifacts with URLs to know where to find all the related files (binaries,
+logs, generated results etc.).
+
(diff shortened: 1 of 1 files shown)


gctucker:
Deployed on staging: https://static.staging.kernelci.org/docs/api/

doc/_index.md
```diff
@@ -1,13 +1,85 @@
...
+objects.  These can contain data about any part of the testing hierarchy such
+as a kernel revision, a build, static test results, runtime functional tests,
+regressions etc.  Each node has a parent so they form a tree
+([DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph)).  There is
+typically one root node for each kernel revision with lots of child nodes
+containing all the test data that relates to it.
+
+Each node object also has a state which can be used when orchestrating the
+pipeline.  For example, a node will be in the Running state while awaiting some
+results.  There's also a result value, to tell whether the job passed or
```
pawiecz, line 27: What is a "job" in this context? A generic task to be run, K8s job, LAVA job or something else?

doc/_index.md
```diff
@@ -1,13 +1,85 @@
...
+all the related files (binaries, logs, generated results etc.).
+
+Here's a publicly available
+[graph](https://charts.mongodb.com/charts-kernelci-api-staging-otxuy/public/dashboards/5b52c3c6-81bb-4658-a5f3-8ebd2e980436#)
+showing the number of nodes added to the database every day:
+
+<div style="padding: 0 0 20px 0"><iframe style="background: #FFFFFF;border: none;border-radius: 2px;box-shadow: 0 2px 10px 0 rgba(70, 76, 79, .2);" width="640" height="480" src="https://charts.mongodb.com/charts-kernelci-api-staging-otxuy/embed/charts?id=64c9fa16-ba81-40e9-8cd7-cd56fb88fbf3&maxDataAge=3600&theme=light&autoRefresh=true"></iframe></div>
+
+> **Note:** The API doesn't manage storage, the only requirement is to provide
+> publicly-available HTTP(s) URLs for each artifact.
```
pawiecz, line 38: minor: typo: s/(s)/(S)/ ?

doc/_index.md
```diff
@@ -1,13 +1,85 @@
...
+
+### Publisher / Subscriber Interface
+
+Every time some data changes in the database, basically every time a node is
```
pawiecz, line 42: minor: typo: s/ is$// ?

doc/_index.md
```diff
@@ -1,13 +1,85 @@
...
+features based on how to handle these events.  The API generates events
+automatically whenever nodes are changing but clients may also use the
+interface to publish their own events and coordinate other parts of the
+pipeline too.
+
+### User Ownership
+
+Interacting with the API requires a token which is associated with a particular
+user.  Whenever a user sends some data such as a new node, it is owned by that
+user.  While all nodes are publicly readable, only the owner of the node can
```
pawiecz, line 61: I suppose admin has superuser rights which overrides this rule - should it be stated explicitly or is it irrelevant in this context?

doc/_index.md
```diff
@@ -1,13 +1,85 @@
...
+
+While the main KernelCI jobs will be creating nodes with users from a
+particular `kernelci.org` group, all the other users can create their own data
+which will coexist in the database.  Then your own nodes can have parents
+created by users.  For example, you may submit test results that relate to a
+kernel build provided by KernelCI.
+
+## Pipeline
+
+The Pipeline is all the client-side services which are running the actual jobs.
```
pawiecz, line 72: I might be missing some context (same as in L27) on what a "job" is.

I remember there was an attempt to make [test terms glossary](https://elinux.org/Test_Glossary). There's a "Run (noun)" term for what I believe a "job" in KCI API context might be - is it correct? Also, I'm not sure how widely this glossary is used.

doc/_index.md
```diff
@@ -1,13 +1,85 @@
...
+It's orchestrated based on events from the Pub/Sub interface and all the data
+is managed via the API.  Pipeline services are also responsible for uploading
+any artifacts to some independent storage services and provide public URLs to
+access them.
+
+A standart set of pipeline services is run directly by KernelCI alongside the
```
pawiecz, line 78: minor: typo s/standart/standard/ ?

pawiecz:
I forgot I should have added code suggestions for small changes, let me update the comments.

doc/_index.md
pawiecz:
```suggestion
A standard set of pipeline services is run directly by KernelCI alongside the
```

doc/_index.md
gctucker: Yes I think I've seen it written as HTTP(s) before but as it's either HTTP or HTTPS I think a capital letter makes more sense.

doc/_index.md
gctucker: I don't think that's relevant here as this is an overview - and in fact there appears to be a bug so even a user in the "admin" group can't update nodes apparently, see https://github.com/kernelci/kernelci-pipeline/issues/315

doc/_index.md
gctucker: It's true that it's not very well defined yet.  I'll rewrite this with some paraphrasing to avoid just using the word "job" as a wildcard.

Basically it's what you get when you run `kci job submit` etc. but it's very implicit for an overview documentation page.

doc/_index.md
gctucker: See other comment below.  I'll resolve this one as it's just another occurrence of the same issue.

doc/_index.md
gctucker: I think a "run" is a close match but it's a bit unclear in my opinion.  Also "job" means something specific in Kubernetes so it's probably better to avoid using it here to avoid confusion.  And yes while this glossary is very nicely put together I don't think it's an authoritative standard across the whole ecosystem so we can't just assume that everyone will follow the same definitions.

gctucker: Alright I've rephrased these parts and now there's no mention of "job" on this page.  I've replaced it with things like "pipeline steps", basically it's about what the pipeline effectively does.  Please let me know if this is OK.

gctucker:
Thanks for the review, I think I've addressed all the comments.  I'll now add another page about how to get an account and start using the public API instance deployed in AKS etc.  I think I'll merge this on Monday anyway to have the documentation ready for communicating about the beginning Early Access phase but it's always possible to add comments later on and I'll make follow-up PRs to fix anything else if needed.

PR #342, opened 2023-09-04
doc: rename getting-started to local-instance

gctucker:
Rename the getting-started page to local-instance as it's more about setting up a local instance with docker-compose than starting to use the API from a user point of view.  Update the Hugo front-matter accordingly.

### DIFF
--- a/doc/local-instance.md
+++ b/doc/local-instance.md
@@ -1,8 +1,7 @@
 ---
-title: "Getting Started"
-date: 2023-07-24
-description: "Setting up a local instance"
-weight: 1
+title: "Local instance"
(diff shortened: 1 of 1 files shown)


doc/local-instance.md
```diff
@@ -1,8 +1,7 @@
 ---
-title: "Getting Started"
-date: 2023-07-24
-description: "Setting up a local instance"
-weight: 1
+title: "Local instance"
+date: 2023-09-01
+description: "Setting up a local KernelCI instance"
```
JenySadadia, line 4: The `weight` property is removed.

PR #344, opened 2023-09-04
doc: add early-access page

gctucker:
Add a page about the Early Access phase to make it easier for new users to get started.

### DIFF
--- a/doc/early-access.md
+++ b/doc/early-access.md
@@ -1,0 +1,175 @@
+---
+title: "Early Access"
+date: 2023-09-04
+description: "Beta-testing the new API"
+weight: 2
+---
+
+As per the [API Transition
+Timeline](https://kernelci.org/blog/posts/2023/api-timeline/), the
+Early Access phase is from 4th September to 4th December 2023.  It
+allows users to request an account on the new KernelCI API for
+beta-testing purposes.
+
+There are still lots of incomplete or missing features with the new
+API & Pipeline.  It now has a [production-like
+deployment](https://github.com/kernelci/kernelci-api/tree/main/kube/aks)
+and can already run a minimalist pipeline with KUnit, a kernel build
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
LGTM

gctucker:
Deployed on staging:
https://static.staging.kernelci.org/docs/api/early-access/

doc/early-access.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+> **Tip**: If you're wondering how it managed to find the API, take a look at
+> `/etc/kernelci/core/api-configs.yaml` from within the Docker container.
+
+* Once you've received your confirmation email with your randomly-generated
+  password, you should change it to use your own arbitrary one instead:
+
+```sh
+$ kernelci@d7e27fe0ba0b:~$ kci user change_password --user <your-user-name>
```
pawiecz, line 71: minor: should the host prompt (`^$ `) be kept with the docker one?

I don't have strong preference for or against it - I like both differentiation from the first listing as well as the below ones being clean and concise

pawiecz (approved):
:+1:

doc/early-access.md
gctucker: Good point.  I think it's important to make it clear what is run directly on the host and what is in the container.  Let's see if I can clarify that, maybe like I did in the AKS README with first a step to start the container than assume the rest of the commands are all run from it, with some explanation.

gctucker: OK I've added a small paragraph about it, please let me know if this makes sense.

gctucker: (also removed a long container prompt as a result)

gctucker:
I've now added a part about AzureFiles as that's also important, as a default way for users to store data alongside the API during the Early Access phase.

Issue #343, opened 2023-09-04
Add user parameter when subscribing


gctucker:
When subscribing to receive events whenever data changes, it should be possible to filter out nodes that aren't owned by the subscribing user or a particular group.  That way, the `kci` user running the main pipeline won't get events whenever another user creates a new node etc.

JenySadadia:
Linked a PR.

PR #346, opened 2023-09-05
api.main: add node owner information to pubsub event

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/343

In order to enable filtering nodes based on owner, add node owner information to pubsub event data for `node` channel.


gctucker:
I think that's necessary but not sufficient, it doesn't take into account user groups.  That's fine, I'll just re-open the issue after merging.

gctucker (approved):

PR #353, opened 2023-09-14
api.main: PUT '/user/profile' endpoint improvments

JenySadadia:
Update handler name for PUT `/user/profile/<username>` endpoint.
Add proper docstring for the handler.
Also, make `email` field optional in the request query parameter in case user does not want to update it.


gctucker (commented):
LGTM at first glance, will give it a try on staging to complete the review.

PR #354, opened 2023-09-14
Update user model

JenySadadia:
api.main: add PUT '/user/<username>' endpoint
    
Implement a handler to update user models. Only admin users will be allowed to access this endpoint and be able to
update other user models. All user fields except password can be updated using this handler.

Depends on #353

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -267,6 +267,45 @@ async def put_user_profile(
     return obj
 
 
+@app.put('/user/{username}', response_model=User,
+         response_model_by_alias=False)
+async def put_user(
+        username: str,
+        email: str = None,
+        groups: List[str] = Query([]),
+        current_user: User = Security(get_user, scopes=["admin"])):
+    """Update user model
+    Allow admin users to update all user fields except password"""
+    user = await db.find_one_by_attributes(
+            User, {'profile.username': username})
+    if not user:
+        raise HTTPException(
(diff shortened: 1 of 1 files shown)


nuclearcat:
Is there other way how admin can update user password? Sometimes they might forget their passwords.

gctucker:
I guess we should have a solution for users to reset their passwords via email.

JenySadadia:
> Is there other way how admin can update user password? Sometimes they might forget their passwords.

Admins shouldn't be able to reset other users' passwords for the sake of security.
I agree with Guillaume that we should implement a different mechanism for it.

gctucker:
Note: For the Early Access phase, it's already possible to replace a password via the MongoDB shell.  I just don't think we should be adding an API feature for this as a production solution.  Let's see if some fully-fledged user account packages for FastAPI can handle all this out of the box.

gctucker:
@JenySadadia Is this PR based on top of #353 or does it replace it?

JenySadadia:
> @JenySadadia Is this PR based on top of #353 or does it replace it?

Yes, created on top of #353 to avoid merge conflicts.

JenySadadia:
`kci` command from https://github.com/kernelci/kernelci-core/pull/2103 can be used to test this PR.

gctucker (approved):
OK I guess we're good to merge this now, and later this will probably reworked with fastapi-users.

PR #357, opened 2023-09-18
api.operators: Add ne operator

nuclearcat:
Not equal is one of necessary basic mongodb operators, which might come handy to exclude certain results.

### DIFF
--- a/api/db.py
+++ b/api/db.py
@@ -32,6 +32,7 @@ class Database:
         'lte': '$lte',
         'gt': '$gt',
         'gte': '$gte',
+        'ne': '$ne',
     }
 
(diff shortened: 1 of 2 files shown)


gctucker:
@nuclearcat Please add "Fixes: <url>" with the issue about improving operators in the description.

gctucker (commented):
LGTM at first glance - let's just verify it by hand with the staging deployment as I don't think there are unit tests for operators yet.

JenySadadia (approved):
Tested OK locally.

gctucker:
By the way, I just realised that `api.operators` is not a meaningful prefix for the Git commit subject as it doesn't relate to the code structure.

PR #360, opened 2023-09-21
doc/early-access: clarify staging_config settings

gctucker:
Clarify the documentation about setting up storage to also mention that it's convenient to add storage_config to the settings file to avoid having to keep passing --storage-config on the command line.

### DIFF
--- a/doc/early-access.md
+++ b/doc/early-access.md
@@ -124,7 +124,9 @@ also use SSH or soon any S3-compatible storage.  There is a quota, currently
 so please don't rely on it for any persistent storage.
 
 The token from your confirmation email should also be stored in `kernelci.toml`
-in order to make use of it, for example:
+as `storage_cred` in order to make use of it, and it's convenient to also add a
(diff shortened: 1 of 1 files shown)


JenySadadia (approved):
Tested OK on staging.

gctucker:
Ah I just noticed a typo in the commit subject, it should be storage_config. Never mind :)

JenySadadia:
Oh, yes. I can revert the merge and we can fix it.

gctucker:
No I think we'll leave it as-is, adding a revert commit would be a bigger issue than a typo in the commit subject.

JenySadadia:
> No I think we'll leave it as-is, adding a revert commit would be a bigger issue than a typo in the commit subject.

Okay. What kind of issue? Is it considered to be bad practice?

gctucker:
Well a typo in the commit message is a bit misleading when reading the git history, but 2 additional commits to have a revert of the commit with a typo and then the same commit again without the typo just adds way more noise to the git history than the original typo.

PR #361, opened 2023-09-21
doc: remove DAG reference as it's just a plain tree

gctucker:
Remove the link to the definition of a DAG as the nodes form a simple tree structure which isn't really a DAG.  Nodes don't have multiple parents.

Fixes: b3113449fe78 ("doc: rewrite overview with more information")


gctucker:
Verified on staging: https://static.staging.kernelci.org/docs/api/

PR #362, opened 2023-09-22
api/main.py: Implement API request to send recovery email

nuclearcat:
Admin should be able to request new password that will be sent directly to user email.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -9,6 +9,8 @@
 """KernelCI API main module"""
 
 import os
+import random
+import string
 from typing import List, Union
 from fastapi import (
     Depends,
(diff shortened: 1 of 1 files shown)


gctucker (changes requested):
Sorry I don't think we can merge this, at least not right now.  There's already a way to manually change a user's password in MongoDB and we're in the middle of investigating a proper implementation for user management.

JenySadadia:
> Sorry I don't think we can merge this, at least not right now. There's already a way to manually change a user's password in MongoDB and we're in the middle of investigating a proper implementation for user management.

That is true. We should not add any new feature related to user management atm as it would need a rework based on the package we decide to integrate.

PR #363, opened 2023-09-25
PoC: `fastapi-users`

JenySadadia:
This is a PoC of `fastapi-users` package for user management tools in API.
This is intended for investigation and review only. Improvements are needed for the actual integration.

### DIFF
--- a/api/auth.py
+++ b/api/auth.py
@@ -195,3 +195,8 @@ auth_backend = AuthenticationBackend(
     transport=bearer_transport,
     get_strategy=get_jwt_strategy,
 )
+
+fastapi_users_instance = FastAPIUsers[TestUser, PydanticObjectId](
+    get_user_manager,
+    [auth_backend],
+)
--- a/api/auth.py
+++ b/api/auth.py
@@ -13,11 +13,12 @@ from pydantic import BaseModel, BaseSettings, Field
 from .db import Database
 from .models import User
 from fastapi_users.db import BeanieUserDatabase, ObjectIDIDMixin
-from fastapi_users import BaseUserManager
+from fastapi_users import BaseUserManager, FastAPIUsers
 from typing import Optional, Any, Dict
 from .models import TestUser
 from beanie import PydanticObjectId
 from fastapi import Depends, Request, Response
+from fastapi_users.authentication import AuthenticationBackend, BearerTransport, JWTStrategy
 
 
 class Token(BaseModel):
@@ -183,3 +184,14 @@ async def get_user_db():
 
 async def get_user_manager(user_db: BeanieUserDatabase = Depends(get_user_db)):
     yield UserManager(user_db)
+
+bearer_transport = BearerTransport(tokenUrl="auth/jwt/login")
+
+def get_jwt_strategy() -> JWTStrategy:
+    return JWTStrategy(secret=SECRET, lifetime_seconds=3600)
+
+auth_backend = AuthenticationBackend(
+    name="jwt",
+    transport=bearer_transport,
+    get_strategy=get_jwt_strategy,
+)
--- a/api/auth.py
+++ b/api/auth.py
@@ -12,6 +12,12 @@ from passlib.context import CryptContext
 from pydantic import BaseModel, BaseSettings, Field
 from .db import Database
 from .models import User
+from fastapi_users.db import BeanieUserDatabase, ObjectIDIDMixin
+from fastapi_users import BaseUserManager
+from typing import Optional, Any, Dict
+from .models import TestUser
+from beanie import PydanticObjectId
+from fastapi import Depends, Request, Response
 
 
 class Token(BaseModel):
@@ -125,3 +131,55 @@ class Authentication:
             if scope not in self._user_scopes:
                 return False, scope
         return True, None
+
+
+SECRET = "SECRET"
+
+class UserManager(ObjectIDIDMixin, BaseUserManager[TestUser, PydanticObjectId]):
+    reset_password_token_secret = SECRET
+    verification_token_secret = SECRET
+
+    async def on_after_register(self, user: TestUser, request: Optional[Request] = None):
+        print(f"User {user.id} has registered.")
+
+    async def on_after_login(
+        self,
+        user: TestUser,
+        request: Optional[Request] = None,
+        response: Optional[Response] = None,
+    ):
+        print(f"User {user.id} logged in.")
+
+    async def on_after_forgot_password(
+        self, user: TestUser, token: str, request: Optional[Request] = None
+    ):
+        print(f"User {user.id} has forgot their password. Reset token: {token}")
+
+    async def on_after_request_verify(
+        self, user: TestUser, token: str, request: Optional[Request] = None
+    ):
+        print(f"Verification requested for user {user.id}. Verification token: {token}")
+        return {"token": token}
+
+    async def on_after_verify(
+        self, user: TestUser, request: Optional[Request] = None
+    ):
+        print(f"Verification successful for user {user.id}")
+
+    async def on_after_update(
+        self,
+        user: TestUser,
+        update_dict: Dict[str, Any],
+        request: Optional[Request] = None,
+    ):
+        print(f"User {user.id} has been updated with {update_dict}.")
+
+    async def on_before_delete(self, user: TestUser, request: Optional[Request] = None):
+        print(f"User {user.id} is going to be deleted")
+
+async def get_user_db():
+        """Database adapter for fastapi-users"""
+        yield BeanieUserDatabase(TestUser)
+
+async def get_user_manager(user_db: BeanieUserDatabase = Depends(get_user_db)):
+    yield UserManager(user_db)
--- a/api/main.py
+++ b/api/main.py
@@ -38,12 +38,17 @@ from .models import (
     UserGroup,
     UserProfile,
     TestUser,
+    UserCreate,
+    UserRead,
+    UserUpdate,
     Password,
     get_model_from_kind
 )
 from .paginator_models import PageModel
 from .pubsub import PubSub, Subscription
 from beanie import init_beanie
+from .auth import fastapi_users_instance, auth_backend
+
 
 app = FastAPI()
 db = Database(service=(os.getenv('MONGO_SERVICE') or 'mongodb://db:27017'))
@@ -640,6 +645,38 @@ async def put_regression(regression_id: str, regression: Regression,
                                                    'id': str(obj.id)})
     return obj
 
+
+app.include_router(
+    fastapi_users_instance.get_auth_router(auth_backend, requires_verification=True),
+    prefix="/auth/jwt",
+    tags=["auth"]
+)
+app.include_router(
+    fastapi_users_instance.get_register_router(UserRead, UserCreate),
+    prefix="/auth",
+    tags=["auth"],
+)
+app.include_router(
+    fastapi_users_instance.get_reset_password_router(),
+    prefix="/auth",
+    tags=["auth"],
+)
+app.include_router(
+    fastapi_users_instance.get_verify_router(UserRead),
+    prefix="/auth",
+    tags=["auth"],
+)
+app.include_router(
+    fastapi_users_instance.get_users_router(UserRead, UserUpdate, requires_verification=True),
+    prefix="/users",
+    tags=["users"],
+)
+
+current_active_user = fastapi_users_instance.current_user(active=True)
+
+@app.get("/authenticated-route")
+async def authenticated_route(user: TestUser = Depends(current_active_user)):
+    return {"message": f"Hello {user.username}!"}
 
 app = VersionedFastAPI(
         app,
--- a/api/main.py
+++ b/api/main.py
@@ -37,11 +37,13 @@ from .models import (
     User,
     UserGroup,
     UserProfile,
+    TestUser,
     Password,
     get_model_from_kind
 )
 from .paginator_models import PageModel
 from .pubsub import PubSub, Subscription
+from beanie import init_beanie
 
 app = FastAPI()
 db = Database(service=(os.getenv('MONGO_SERVICE') or 'mongodb://db:27017'))
@@ -63,6 +65,15 @@ async def create_indexes():
     """Startup event handler to create database indexes"""
     await db.create_indexes()
 
+@app.on_event('startup')
+async def fastapi_users_beanie_init():
+    """Startup event handler for beanie"""
+    await init_beanie(
+        database=db.db,  
+        document_models=[
+            TestUser,  
+        ],
+    )
 
 @app.exception_handler(ValueError)
 async def value_error_exception_handler(request: Request, exc: ValueError):
@@ -639,6 +650,7 @@ app = VersionedFastAPI(
         on_startup=[
             pubsub_startup,
             create_indexes,
+            fastapi_users_beanie_init,
         ]
     )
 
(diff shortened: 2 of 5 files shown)


JenySadadia:
Below are the example commands for requesting routes provided by the package:

1. Register a user
```
$ curl -X 'POST' \
  'http://localhost:8001/latest/auth/register' \
  -H 'Content-Type: application/json' \
  -d '{
  "email": "user@example.com",
  "password": "test",
  "username": "test"
}'
{"id":"65116ac07e1037b63f946c08", "email": "user@example.com", "password": "test", "is_active": true,  "is_superuser": false, "is_verified": false, "username": "test"}
```
2. Verify email
Request token for verification:
```
curl -X 'POST' \
  'http://localhost:8001/latest/auth/request-verify-token' \
  -H 'Content-Type: application/json' \
  -d '{
  "email": "user@example.com"
}'
```
Reponse: null

API logs: 
```
kernelci-api | Verification requested for user 650bead8a4de36b741e3fbcf. Verification token: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiI2NTBiZWFkOGE0ZGUzNmI3NDFlM2ZiY2YiLCJlbWFpbCI6Implbnkuc2FkYWRpYUBnbWFpbC5jb20iLCJhdWQiOiJmYXN0YXBpLXVzZXJzOnZlcmlmeSIsImV4cCI6MTY5NTM4MDEwMn0.0byhPu0WTvdpli1IuEcCYY_b40-VjqqI-7TtOgnbwHo
```
Provide the verification token to `/verify` endpoint.
```
curl -X 'POST' \
  'http://localhost:8001/latest/auth/verify' \
  -H 'Content-Type: application/json' \
  -d '{
  "token": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiI2NTBiZWFkOGE0ZGUzNmI3NDFlM2ZiY2YiLCJlbWFpbCI6Implbnkuc2FkYWRpYUBnbWFpbC5jb20iLCJhdWQiOiJmYXN0YXBpLXVzZXJzOnZlcmlmeSIsImV4cCI6MTY5NTM4MDEwMn0.0byhPu0WTvdpli1IuEcCYY_b40-VjqqI-7TtOgnbwHo"
}'
{"id": null, "email": "user@example.com", "is_active": true, "is_superuser": false, "is_verified": true, "username": "test"}
```
> Note: need to investigate why `id` is `null`  in the response.
We can add logic to get the token in email using `on_after_request_verify` handler from `UserManager`.

3. Login
```
curl -X 'POST'   'http://localhost:8001/latest/auth/jwt/login'   -H 'accept: application/json'   -H 'Content-Type: application/x-www-form-urlencoded'   -d 'grant_type=&username=user@example.com&password=test&scope=&client_id=&client_secret='
{"access_token":"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiI2NTBiZWFkOGE0ZGUzNmI3NDFlM2ZiY2YiLCJhdWQiOlsiZmFzdGFwaS11c2VyczphdXRoIl19.a1D8r_z3L0MyIp83n88cHMtIv3DqxCDcH7wvSSwDeSo",
"token_type":"bearer"}
```
> Note: It supports tokens with OAuth scopes.

4. Logout
```
$ curl -X 'POST' \
  'http://localhost:8001/latest/auth/jwt/logout' \
  -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiI2NTBiZWU0MzBlZjA5M2U2YmVlZDg3OGUiLCJhdWQiOlsiZmFzdGFwaS11c2VyczphdXRoIl0sImV4cCI6MTY5NTI5MjMwN30.iYYMxxHf5F_qrMUbNkfom5lMevE5lPgrNf2NO_U7wVw'
Reponse: null
```
> Note: JWT logout is not implemented as the authentication backend was configured with token expiry time. Please see https://github.com/fastapi-users/fastapi-users/issues/154#issuecomment-615308037

5. Reset password
```
Get token for resetting password:
$ curl -X 'POST' \
  'http://localhost:8001/latest/auth/forgot-password' \
  -H 'Content-Type: application/json' \
  -d '{
  "email": "user@example.com"
}'
Reponse: null
```

API logs:
```
kernelci-api | User 650bead8a4de36b741e3fbcf has forgot their password. Reset token: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiI2NTBiZWFkOGE0ZGUzNmI3NDFlM2ZiY2YiLCJwYXNzd29yZF9mZ3B0IjoiJDJiJDEyJHQxUHI4R1hFTGRSWEFBS0g4NFVpLk8zRXdidTZtN1g5TWN6RU4zT2l5LlNiLzJqd2dob2VtIiwiYXVkIjoiZmFzdGFwaS11c2VyczpyZXNldCIsImV4cCI6MTY5NTMwNDYwMn0.R-Iew4EBTZqA_qtpYgUINcnpnacBVHB7Gdatb4-ZU9Q
```
 Reset the password using token:
```
$ curl -X 'POST' \
  'http://localhost:8001/latest/auth/reset-password' \
  -H 'Content-Type: application/json' \
  -d '{
  "token": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiI2NTBiZWFkOGE0ZGUzNmI3NDFlM2ZiY2YiLCJwYXNzd29yZF9mZ3B0IjoiJDJiJDEyJEZYeS9OVWZ4TGRIdlMxMVUvNmMzNGVDOG9nNTQxZWgwU1VYZGRRYmR3YWMxZGY5MGtKSjIuIiwiYXVkIjoiZmFzdGFwaS11c2VyczpyZXNldCIsImV4cCI6MTY5NTMwNTEyNn0.HKwIR-rN4qgJz-Fw1tLE4fbz7EA5SANFlBQPg3egN3I",
  "password": "jeny"
}'
```
> Note:  Sending tokens via emails can be done by adding logic to `UserManager.on_after_forgot_password` handler.

6. Get user
```
$ curl -X 'GET'   'http://localhost:8001/latest/users/me'   -H 'accept: application/json'   -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiI2NTBiZWFkOGE0ZGUzNmI3NDFlM2ZiY2YiLCJhdWQiOlsiZmFzdGFwaS11c2VyczphdXRoIl19.a1D8r_z3L0MyIp83n88cHMtIv3DqxCDcH7wvSSwDeSo'
{"id":"650bead8a4de36b741e3fbcf","email":"user@example.com","is_active":true,"is_superuser":false,"is_verified":true,"username":"test"}
```

7. Update user
```
$ curl -X 'PATCH'   'http://localhost:8001/latest/users/me'   -H 'accept: application/json'   -H 'Content-Type: application/json'   -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiI2NTBiZWFkOGE0ZGUzNmI3NDFlM2ZiY2YiLCJhdWQiOlsiZmFzdGFwaS11c2VyczphdXRoIl19.a1D8r_z3L0MyIp83n88cHMtIv3DqxCDcH7wvSSwDeSo' -d '{
  "password": "test",
  "email": "user1@example.com",
  "username": "test"
}'
{"id":"650bead8a4de36b741e3fbcf","email":"user1@example.com","is_active":true,"is_superuser":false,"is_verified":true,"username":"test"}
```

8. Get user by id  (accessible by superuser only)
```
$ curl -X 'GET'   'http://localhost:8001/latest/users/650bead8a4de36b741e3fbcf'   -H 'accept: application/json'   -H 'Content-Type: application/json'   -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiI2NTBkODZkZmQ2NDU2YTNiYTM2NDNiMDYiLCJhdWQiOlsiZmFzdGFwaS11c2VyczphdXRoIl19.Nb-u2UMAY6cXScO2MziGjZzYGN-H81Gj_HnV_32nThE'
{"id":"650bead8a4de36b741e3fbcf","email":"user@example.com","is_active":true,"is_superuser":false,"is_verified":false,"username":"test"}
```

9. Update user by id (accessible by superuser only)
```
$ curl -X 'PATCH'   'http://localhost:8001/latest/users/650be1de418d00f4cc61cd2e'   -H 'accept: application/json'   -H 'Content-Type: application/json'   -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiI2NTBkODZkZmQ2NDU2YTNiYTM2NDNiMDYiLCJhdWQiOlsiZmFzdGFwaS11c2VyczphdXRoIl19.Nb-u2UMAY6cXScO2MziGjZzYGN-H81Gj_HnV_32nThE' -d '{"username": "test"}'
{"id":"650be1de418d00f4cc61cd2e","email":"user@example.com","is_active":true,"is_superuser":false,"is_verified":false,"username":"test"}
```

10. Delete user by id (accessible by superuser only)
```
$ curl -X 'DELETE'   'http://localhost:8001/latest/users/650d85d6d6456a3ba3643b05'   -H 'accept: application/json'   -H 'Content-Type: application/json'   -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiI2NTBkODZkZmQ2NDU2YTNiYTM2NDNiMDYiLCJhdWQiOlsiZmFzdGFwaS11c2VyczphdXRoIl19.Nb-u2UMAY6cXScO2MziGjZzYGN-H81Gj_HnV_32nThE'
Returns HTTP 204 No Content
```

11. Authenticate user
```
$ curl -X 'GET' 'http://localhost:8001/latest/authenticated-route' -H 'accept: application/json'  -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiI2NTBiZWU0MzBlZjA5M2U2YmVlZDg3OGUiLCJhdWQiOlsiZmFzdGFwaS11c2VyczphdXRoIl0sImV4cCI6MTY5NTI5MjMwN30.iYYMxxHf5F_qrMUbNkfom5lMevE5lPgrNf2NO_U7wVw'
{"message":"Hello test!"}
```

gctucker:
Looks like a great initial investigation.  I shall give it a try to get a clearer idea of how this would work and try to provide some feedback.

On a side note, I see this uses `PATCH` queries for updating users.  I wonder if it's better than `PUT` which is what we're using with other endpoints currently.  Maybe `PATCH` is better for updating some data whereas `PUT` is normally expected to be used when replacing files?  I would need to read the standard definitions again to be sure.

JenySadadia:
> On a side note, I see this uses `PATCH` queries for updating users. I wonder if it's better than `PUT` which is what we're using with other endpoints currently. Maybe `PATCH` is better for updating some data whereas `PUT` is normally expected to be used when replacing files? I would need to read the standard definitions again to be sure.

Yes, `PATCH` is for partial updates, and `PUT` is for replacing objects.
I agree that `PATCH` is better here as with it we don't need to provide fields that are not changing.

api/models.py
```diff
@@ -24,6 +24,11 @@
     FileUrl,
     SecretStr,
 )
+from beanie import Indexed, Document
+from fastapi_users.db import BeanieBaseUser
+from fastapi_users import schemas
+from beanie import PydanticObjectId, Indexed
+from typing import Optional
```
gctucker, line 31: I wonder if these should rather be in a separate module to avoid "polluting" the whole `api.models` namespace with `fastapi_users` imports.

api/models.py
```diff
@@ -148,6 +153,24 @@ def create_indexes(cls, collection):
...
+class UserRead(schemas.BaseUser[PydanticObjectId]):
+    username: Indexed(str, unique=True)
+
+
+class UserCreate(schemas.BaseUserCreate):
+    username: Indexed(str, unique=True)
+
+
+class UserUpdate(schemas.BaseUserUpdate):
+    username: Optional[Indexed(str, unique=True)]
```
gctucker, line 170: Out of interest, why is this one optional whereas it's not in the other models?

api/main.py
```diff
@@ -63,6 +70,15 @@ async def create_indexes():
...
 
+@app.on_event('startup')
+async def fastapi_users_beanie_init():
+    """Startup event handler for beanie"""
+    await init_beanie(
+        database=db.db,  
+        document_models=[
+            TestUser,  
+        ],
+    )
```
gctucker, line 81: Maybe this call could be moved to `api.auth` with a method in the Database class to avoid exposing the underlying `.db` object?  That way only the Database class would follow the object-oriented principle of accessing its own data.  We already know we need to import the `beanie` modules in `api.auth` so it wouldn't be adding extra dependencies there.

gctucker (commented):
Looks great to me :)  Just a few comments but overall it seems like we're ready to upgrade this PR to a full implementation.

We might need to defer some things such as password reset via email and user moderation into follow-up PRs to discuss a few security issues with the sysadmins first.

@nuclearcat @VinceHillier ^ What do you guys think?

gctucker:
Ah also I think this doesn't enable new users to sing up with just an email address, so that's one less security thing to consider here.  We'll need to work this out properly for a full production deployment though.

api/models.py
```diff
@@ -148,6 +153,24 @@ def create_indexes(cls, collection):
         collection.create_index("profile.username", unique=True)
 
 
+class TestUser(BeanieBaseUser, Document):
```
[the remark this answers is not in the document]
JenySadadia: Yes, exactly.

api/models.py
JenySadadia: Yes, we should have a separate model file just like we have for pagination related models.

api/models.py
JenySadadia: That's to allow users to only send fields that are updating. If I remove the `Optional`, the endpoint would expect `username` field in the request even if the user wants it to be unchanged.
If you look at the definition of the parent class `BaseUserUpdate`, you'll see all the fields are optional.
```
class BaseUserUpdate(CreateUpdateDictModel):
    password: Optional[str]
    email: Optional[EmailStr]
    is_active: Optional[bool]
    is_superuser: Optional[bool]
    is_verified: Optional[bool]
```

api/main.py
JenySadadia: The reason I kept this call in `main.py` was that `init_beanie` needs an actual instantiated database object and not just the class reference. Still, I'll take another look if I can move things around.

api/auth.py
```diff
@@ -125,3 +132,71 @@ async def validate_scopes(self, requested_scopes):
...
+        user: TestUser,
+        request: Optional[Request] = None,
+        response: Optional[Response] = None,
+    ):
+        print(f"User {user.id} logged in.")
+
+    async def on_after_forgot_password(
+        self, user: TestUser, token: str, request: Optional[Request] = None
+    ):
+        print(f"User {user.id} has forgot their password. Reset token: {token}")
```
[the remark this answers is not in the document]
JenySadadia: > So I guess this is where an email would be sent with the reset token in a URL, and then the user would then come back via the on_after_request_verify method?

Yes, that is true.

JenySadadia:
Completed PoC. Moving forward with the actual integration.

api/models.py
gctucker: Right, I think this brings us back to the idea I mentioned a while ago about having a different model for the API and for the database.

PR #366, opened 2023-09-29
doc: fix absolute link for creating admin users

gctucker:
Fix the link to the part about creating admin users as this should be an absolute URI with a leading / rather than a relative one.

Fixes: c752060d7683 ("doc/api-details: update `user` section")


JenySadadia (approved):
Tested OK on staging.

PR #368, opened 2023-10-03
api.db: add `ne` operator

JenySadadia:
Support `ne`(not equal to) operator while querying objects from API.
e.g. `http://API_URL:8001/latest/nodes?name__ne=checkout` will fetch all the nodes except ones with name `checkout`.

### DIFF
--- a/doc/api-details.md
+++ b/doc/api-details.md
@@ -44,10 +44,11 @@ users are allowed to create user accounts.
 
 ### Create an admin user
 
-The very first admin user needs to be created with [`api.admin`](https://github.com/kernelci/kernelci-api/blob/main/api/admin.py)
+The very first admin user needs to be created with
+[`api.admin`](https://github.com/kernelci/kernelci-api/blob/main/api/admin.py)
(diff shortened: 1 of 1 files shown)


nuclearcat:
Isn't it already merged?
https://github.com/kernelci/kernelci-api/pull/357

gctucker:
Yes it's quite puzzling how GitHub can have a PR to change a line that's already changed in the same way :thinking:

gctucker:
@JenySadadia Your branch was based on an older revision that didn't have @nuclearcat's changes.  If you rebase than the changes in this PR disappear.

I would be interested to see how GitHub would manage that, like if we click on Merge now would it just mark it as "merged" without adding any actual commit to the branch?  I don't really need to find out...

JenySadadia:
This is weird.
Bdw looking at https://github.com/kernelci/kernelci-api/pull/357, I was the one who tested the PR and yet I created another one :sweat_smile:

JenySadadia:
It discarded my PR after rebase. The PR got closed automatically upon forced push.

PR #377, opened 2023-10-06
Integrate `fastapi-users` for user management

JenySadadia:
Use `fastapi-users` for user management logic in the API.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,13 +54,9 @@ jobs:
           pylint api.user_models
           pylint tests/unit_tests
 
-      - name: Export environment variables
-        run: |
-          echo "SECRET_KEY=$(openssl rand -hex 32)" >> $GITHUB_ENV
-
       - name: Run pytest
         run: |
-          pytest -v tests/unit_tests
+          SECRET_KEY=$(openssl rand -hex 32) SMTP_HOST=smtp.gmail.com SMTP_PORT=465 EMAIL_SENDER=test@kernelci.org EMAIL_PASSWORD=random pytest -vs tests/unit_tests/
 
   lint:
     runs-on: ubuntu-22.04
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -50,6 +50,8 @@ jobs:
           pylint api.main
           pylint api.models
           pylint api.pubsub
+          pylint api.user_manager
+          pylint api.user_models
           pylint tests/unit_tests
 
       - name: Export environment variables
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,6 +30,10 @@ jobs:
       - name: Export environment variables
         run: |
           echo "SECRET_KEY=$(openssl rand -hex 32)" > .env
+          echo "SMTP_HOST=smtp.gmail.com" >> .env
+          echo "SMTP_PORT=465" >> .env
+          echo "EMAIL_SENDER=test@kernelci.org" >> .env
+          echo "EMAIL_PASSWORD=random" >> .env
 
       - name: Build docker images
         run: docker-compose -f test-docker-compose.yaml build
--- a/api/admin.py
+++ b/api/admin.py
@@ -17,7 +17,8 @@ import getpass
 
 from .auth import Authentication
 from .db import Database
-from .models import User, UserGroup, UserProfile
+from .models import UserGroup
+from .user_models import User
 
 
 async def setup_admin_group(db, admin_group):
@@ -42,19 +43,18 @@ async def setup_admin_user(db, username, email, admin_group):
         return None
     hashed_password = Authentication.get_password_hash(password)
     print(f"Creating {username} user...")
-    profile = UserProfile(
+    return await db.create(User(
         username=username,
         hashed_password=hashed_password,
         email=email,
-        groups=[admin_group]
-    )
-    return await db.create(User(
-        profile=profile
+        groups=[admin_group],
+        is_superuser=1
     ))
 
 
 async def main(args):
     db = Database(args.mongo, args.database)
+    await db.initialize_beanie()
     group = await setup_admin_group(db, args.admin_group)
     user = await setup_admin_user(db, args.username, args.email, group)
     return True
--- a/api/db.py
+++ b/api/db.py
@@ -7,9 +7,11 @@
 """Database abstraction"""
 
 from bson import ObjectId
+from beanie import init_beanie
 from fastapi_pagination.ext.motor import paginate
 from motor import motor_asyncio
-from .models import Hierarchy, Node, User, Regression, UserGroup
+from .models import Hierarchy, Node, Regression, UserGroup
+from .user_models import User
 
 
 class Database:
@@ -38,6 +40,15 @@ class Database:
     def __init__(self, service='mongodb://db:27017', db_name='kernelci'):
         self._motor = motor_asyncio.AsyncIOMotorClient(service)
         self._db = self._motor[db_name]
+
+    async def initialize_beanie(self):
+        """Initialize Beanie ODM to use `fastapi-users` tools for MongoDB"""
+        await init_beanie(
+            database=self._db,
+            document_models=[
+                User,
+            ],
+        )
 
     def _get_collection(self, model):
         col = self.COLLECTIONS[model]
--- a/api/email_sender.py
+++ b/api/email_sender.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2023 Jeny Sadadia
+# Author: Jeny Sadadia <jeny.sadadia@collabora.com>
+
+"""SMTP Email Sender module"""
+
+from email.mime.multipart import MIMEMultipart
+import email
+import email.mime.text
+import smtplib
+from pydantic import BaseSettings, EmailStr
+
+
+class Settings(BaseSettings):
+    """Email settings"""
+    smtp_host: str
+    smtp_port: int
+    email_sender: EmailStr
+    email_password: str
+
+
+class EmailSender:
+    """Class to send email report using SMTP"""
+    def __init__(self):
+        self._settings = Settings()
+
+    def _smtp_connect(self):
+        """Method to create a connection with SMTP server"""
+        if self._settings.smtp_port == 465:
+            smtp = smtplib.SMTP_SSL(self._settings.smtp_host,
+                                    self._settings.smtp_port)
+        else:
+            smtp = smtplib.SMTP(self._settings.smtp_host,
+                                self._settings.smtp_port)
+            smtp.starttls()
+        smtp.login(self._settings.email_sender,
+                   self._settings.email_password)
+        return smtp
+
+    def _create_email(self, email_subject, email_content, email_recipient):
+        """Method to create an email message from email subject, contect,
+        sender, and receiver"""
+        email_msg = MIMEMultipart()
+        email_text = email.mime.text.MIMEText(email_content, "plain", "utf-8")
+        email_text.replace_header('Content-Transfer-Encoding', 'quopri')
+        email_text.set_payload(email_content, 'utf-8')
+        email_msg.attach(email_text)
+        email_msg['To'] = email_recipient
+        email_msg['From'] = self._settings.email_sender
+        email_msg['Subject'] = email_subject
+        return email_msg
+
+    def _send_email(self, email_msg):
+        """Method to send an email message using SMTP"""
+        smtp = self._smtp_connect()
+        if smtp:
+            smtp.send_message(email_msg)
+            smtp.quit()
+
+    def create_and_send_email(self, email_subject, email_content,
+                              email_recipient):
+        """Method to create and send email"""
+        email_msg = self._create_email(
+            email_subject, email_content, email_recipient
+        )
+        self._send_email(email_msg)
--- a/api/email_sender.py
+++ b/api/email_sender.py
@@ -11,14 +11,16 @@ from email.mime.multipart import MIMEMultipart
 import email
 import email.mime.text
 import smtplib
-from pydantic import BaseSettings, EmailStr
+# from pydantic import BaseSettings, EmailStr
+from pydantic import BaseSettings
 
 
 class Settings(BaseSettings):
     """Email settings"""
     smtp_host: str
     smtp_port: int
-    email_sender: EmailStr
+    # email_sender: EmailStr
+    email_sender: str
     email_password: str
 
 
--- a/api/main.py
+++ b/api/main.py
@@ -220,16 +220,6 @@ async def get_users(request: Request):
     return paginated_resp
 
 
-@app.get('/user/{user_id}', response_model=User,
-         response_model_by_alias=False,
-         response_model_exclude={"profile": {"hashed_password"}})
-async def get_user_by_id(
-        user_id: str,
-        current_user: User = Security(get_user, scopes=["admin"])):
-    """Get user matching provided user id"""
-    return await db.find_by_id(User, user_id)
-
-
 @app.put('/user/profile/{username}', response_model=User,
          response_model_include={"profile"},
          response_model_by_alias=False)
--- a/api/main.py
+++ b/api/main.py
@@ -47,6 +47,9 @@ from .pubsub import PubSub, Subscription
 from .user_manager import get_user_manager
 from .user_models import (
     User,
+    UserRead,
+    UserCreate,
+    UserUpdate,
 )
 
 # List of all the supported API versions.  This is a placeholder until the API
@@ -697,6 +700,38 @@ async def put_regression(regression_id: str, regression: Regression,
     await pubsub.publish_cloudevent('regression', {'op': operation,
                                                    'id': str(obj.id)})
     return obj
+
+
+# -----------------------------------------------------------------------------
+# Users
+
+app.include_router(
+    fastapi_users_instance.get_auth_router(auth_backend,
+                                           requires_verification=True),
+    prefix="/user",
+    tags=["user"]
+)
+app.include_router(
+    fastapi_users_instance.get_register_router(UserRead, UserCreate),
+    prefix="/user",
+    tags=["user"],
+)
+app.include_router(
+    fastapi_users_instance.get_reset_password_router(),
+    prefix="/user",
+    tags=["user"],
+)
+app.include_router(
+    fastapi_users_instance.get_verify_router(UserRead),
+    prefix="/user",
+    tags=["user"],
+)
+app.include_router(
+    fastapi_users_instance.get_users_router(UserRead, UserUpdate,
+                                            requires_verification=True),
+    prefix="/user",
+    tags=["user"],
+)
 
 
 app = VersionedFastAPI(
--- a/api/main.py
+++ b/api/main.py
@@ -17,14 +17,8 @@ from fastapi import (
     HTTPException,
     status,
     Request,
-    Security,
-    Query,
 )
 from fastapi.responses import JSONResponse, PlainTextResponse
-from fastapi.security import (
-    OAuth2PasswordRequestForm,
-    SecurityScopes
-)
 from fastapi_pagination import add_pagination
 from fastapi_versioning import VersionedFastAPI
 from bson import ObjectId, errors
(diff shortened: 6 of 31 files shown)


gctucker:
@JenySadadia It looks like the scope of the unit tests isn't completely clear, so sometimes it uses the methods from `api.main` to get users and sometimes it seems to need a mock instead.  So maybe just always use a mock everywhere like before, and potentially add some tests to specifically cover the methods to get users in `api.main`?

On a side note, one thing I didn't figure out while looking at the code yesterday was how the `event_loop()` fixture was being used.  Is it a magic name or something that pytest can use?  At the end of the unit tests there seems to be some deadlock, I wonder if it's related and maybe that's why the GitHub checks are taking quite a long time to complete.

JenySadadia:
Fixed end-to-end tests.

JenySadadia:
> @JenySadadia It looks like the scope of the unit tests isn't completely clear, so sometimes it uses the methods from `api.main` to get users and sometimes it seems to need a mock instead. So maybe just always use a mock everywhere like before, and potentially add some tests to specifically cover the methods to get users in `api.main`?

We are not using methods from `api.main` but mocking them instead in `conftest.py`. The mock functions `mock_get_current_user` and `mock_get_current_admin_user` are used for getting regular user and admin user respectively.

> On a side note, one thing I didn't figure out while looking at the code yesterday was how the `event_loop()` fixture was being used. Is it a magic name or something that pytest can use? At the end of the unit tests there seems to be some deadlock, I wonder if it's related and maybe that's why the GitHub checks are taking quite a long time to complete.

Usually `pytest` uses default `event_loop` to run the tests. In case of async tests, the issue was default event loop was closing before all the test runs were completed. That's why `event_loop` fixture was implemented. Please see https://github.com/kernelci/kernelci-api/commit/732d6b90e0b585983b6bfb7b1c63e44a28d20f3f

> At the end of the unit tests there seems to be some deadlock, I wonder if it's related and maybe that's why the GitHub checks are taking quite a long time to complete.

I think that's because of `PubSub._keep_alive`. Need to investigate what happens there.
```
Task was destroyed but it is pending!
task: <Task pending name='Task-74' coro=<PubSub._keep_alive() running at /home/jeny/kernelCI/kernelci-api_env/src/kernelci-api/api/pubsub.py:81> wait_for=<Future pending cb=[_chain_future.<locals>._call_check_cancel() at /usr/lib/python3.10/asyncio/futures.py:385, Task.task_wakeup()]> cb=[_chain_future.<locals>._call_set_state() at /usr/lib/python3.10/asyncio/futures.py:392]>
```

gctucker:
> We are not using methods from `api.main` but mocking them instead in `conftest.py`. The mock functions `mock_get_current_user` and `mock_get_current_admin_user` are used for getting regular user and admin user respectively.

Ah yes sorry, I mean here:
```python
    app.dependency_overrides[get_current_user] = mock_get_current_user
    app.dependency_overrides[get_current_superuser] = mock_get_current_admin_user
```
The `mock_*` methods aren't actual fixtures any more, but some tests still expect them to be fixtures.  Maybe we should have these methods and actual mock fixtures to wrap around them?  That way tests that just need a test client will get them via the snippet above and those that require a mock method explicitly will get the actual mock.

JenySadadia:
> > We are not using methods from `api.main` but mocking them instead in `conftest.py`. The mock functions `mock_get_current_user` and `mock_get_current_admin_user` are used for getting regular user and admin user respectively.
> 
> Ah yes sorry, I mean here:
> 
> ```python
>     app.dependency_overrides[get_current_user] = mock_get_current_user
>     app.dependency_overrides[get_current_superuser] = mock_get_current_admin_user
> ```
> 
> The `mock_*` methods aren't actual fixtures any more, but some tests still expect them to be fixtures. Maybe we should have these methods and actual mock fixtures to wrap around them? That way tests that just need a test client will get them via the snippet above and those that require a mock method explicitly will get the actual mock.

Actually, I needed to drop `mock_get_current_user` argument from all the tests as it was not a fixture anymore. But some tests were accessing `return_value` from it. And I did not drop it as I wanted to fix it first.
Sorry for the confusion. I've dropped it from all the tests now.

As `get_current_user` is a dependency callable and not a normal function, we need to use `app.dependency_overrides` to override it. We don't need to call the mock explicitly in the function argument and also don't need to declare it as a fixture to use it. It will be overridden automatically and used when required by tests.

gctucker:
Alright I'll stop looking into this for now then, let me know if you need help with anything else.

I did however hit an issue when running unit tests locally as the environment variables aren't created, they're only defined in the GitHub `test.yml` file.  It would be good to address this, but it's a separate thing.  I'll create a GitHub issue about it.

JenySadadia:
> Alright I'll stop looking into this for now then, let me know if you need help with anything else.

The issue with unit tests involving fastapi-users routers (e.g. `/whoami`) are still not resolved yet. I would appreaciate if you can help.

JenySadadia:
Fixed all the unit tests.
To speed up the development, I used different `Settings` class for `EmailSender`. Also, I added env variables to Github workflow files.
I'll add a GH issue for further improvements in handling `env` variables.

.github/workflows/test.yml
```diff
@@ -30,6 +30,10 @@ jobs:
       - name: Export environment variables
         run: |
           echo "SECRET_KEY=$(openssl rand -hex 32)" > .env
+          echo "SMTP_HOST=smtp.gmail.com" >> .env
+          echo "SMTP_PORT=465" >> .env
+          echo "EMAIL_SENDER=test@kernelci.org" >> .env
+          echo "EMAIL_PASSWORD=random" >> .env
```
gctucker, line 36: On a side topic, one thing I've found is that the .env file needs to be populated by hand when running the tests locally.  As these aren't special values but can be anything arbitrary, I think it would be better to have them supplied directly when running the test instead.

templates/email-verification-successful.jinja2
```diff
@@ -0,0 +1,8 @@
+{# SPDX-License-Identifier: LGPL-2.1-or-later -#}
+
+Hello {{ username }},
+
+Your email address has been verified successfully.
+
+Thanks,
+KernelCI SysAdmin
```
gctucker, line 8: These email templates look great for a first implementation.  Then as a follow-up, I think it might be good to have a GitHub discussion about their contents and probably ask the people who will be working on the UX Analysis for the new web dashboard to provide some input there with best practices etc.

api/db.py
```diff
@@ -39,6 +41,15 @@ def __init__(self, service='mongodb://db:27017', db_name='kernelci'):
...
         self._db = self._motor[db_name]
 
+    async def initialize_beanie(self):
+        """Initialize Beanie ODM to use `fastapi-users` tools for MongoDB"""
+        await init_beanie(
+            database=self._db,
+            document_models=[
+                User,
+            ],
+        )
```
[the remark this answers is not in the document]
JenySadadia: This method is specific to DB-related initialization. Hence, I added it to the `Database` class.
Yes, we can refactore it later if required.

templates/email-verification-successful.jinja2
JenySadadia: Yes, makes sense. Thanks for pointing it out.
I'll create a GH discussion.

.github/workflows/test.yml
JenySadadia: Yes, I want to override the `Settings` class for unit tests. Something like described here: https://fastapi.tiangolo.com/advanced/settings/#settings-and-testing

JenySadadia: I have added env variables to `pytest` command. Also, created a follow-up PR for `Settings` class.

gctucker:
@JenySadadia Is this still pending some other changes before the `staging-skip` label can be removed?

api/db.py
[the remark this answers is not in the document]
gctucker: What I mean is that the `Database` class doesn't need the `fastapi-users` initialisation, the dependency is the other way round.  The `Authentication` class uses `fastapi-users` and as such it depends on `Database`.  At least that's my understanding.  But like I wrote, I think it's not a blocker here and this PR is large enough so let's try and have it merged based on functionality from a user point of view and see if any follow-up reworks should be done.

api/user_manager.py
```diff
@@ -0,0 +1,143 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2023 Collabora Limited
+# Author: Jeny Sadadia <jeny.sadadia@collabora.com>
+
+"""User Manager"""
```
gctucker, line 6: It might be a bit better to explain what this module is about.  Maybe docstrings and code clarity in general is something we can take care as a follow-up.

api/user_models.py
```diff
@@ -0,0 +1,54 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2023 Collabora Limited
+# Author: Jeny Sadadia <jeny.sadadia@collabora.com>
+
+# Disable flag as user models don't require any public methods
+# at the moment
+# pylint: disable=too-few-public-methods
+
+"""User model definitions"""
```
gctucker, line 10: It might be worth mentioning `fastapi-users` here to say this is the "glue" code for our particular API application.

gctucker (commented):
Just a few comments, LGTM otherwise.  I guess it'll be good to test this with the reworked `kci user` command line tool, I'll take a look at that too.

JenySadadia:
Added a migration file for `User` schema.

JenySadadia:
Eventually, we are going to drop `admin` group for identifying admin users, I have added `is_superuser=0` by default. I'll
update admin users from the DB shell directly once the migration is applied on staging.

JenySadadia:
Introduced below endpoints with the PR:

1. `/register`: Register a user
```
$ curl -X 'POST' \
  'http://localhost:8001/latest/user/register' \
  -H 'Content-Type: application/json' \
  -d '{
  "email": "user@example.com",
  "password": "test",
  "username": "test"
  "groups": ["kernelci"]
}'
{"id":"65116ac07e1037b63f946c08", "email": "user@example.com", "password": "test", "is_active": true,  "is_superuser": false, "is_verified": false, "username": "test", "groups":[{"id":"648ff894bd39930355ed16ad","name":"kernelci"}}
```
2. Verify email
`/user/request-verify-token` Request token for verification:
```
curl -X 'POST' \
  'http://localhost:8001/latest/user/request-verify-token' \
  -H 'Content-Type: application/json' \
  -d '{
  "email": "user@example.com"
}'
Reponse: null
```
Get token via email


Provide the verification token to `/verify` endpoint.
```
curl -X 'POST' \
  'http://localhost:8001/latest/auth/verify' \
  -H 'Content-Type: application/json' \
  -d '{
  "token": "TOKEN"
}'
{"id": null, "email": "user@example.com", "is_active": true, "is_superuser": false, "is_verified": true, "username": "test"}
```


3. Login
```
curl -X 'POST'   'http://localhost:8001/latest/user/login'   -H 'accept: application/json'   -H 'Content-Type: application/x-www-form-urlencoded'   -d 'username=user@example.com&password=test'
{"access_token":"TOKEN",
"token_type":"bearer"}
```


4. Logout
```
$ curl -X 'POST' \
  'http://localhost:8001/latest/user/logout' \
  -H 'Authorization: Bearer TOKEN'
Reponse: null
```
[Not implemented atm]

5. Reset password
```
`user/forgot-password` Get token for resetting password:
$ curl -X 'POST' \
  'http://localhost:8001/latest/user/forgot-password' \
  -H 'Content-Type: application/json' \
  -d '{
  "email": "user@example.com"
}'
Reponse: null
```
Get token via email


`user/reset-password` Reset the password using token:
```
$ curl -X 'POST' \
  'http://localhost:8001/latest/user/reset-password' \
  -H 'Content-Type: application/json' \
  -d '{
  "token": "TOKEN",
  "password": "jeny"
}'
```

6. Get user
```
$ curl -X 'GET'   'http://localhost:8001/latest/user/me'   -H 'accept: application/json'   -H 'Authorization: Bearer TOKEN'
{"id":"650bead8a4de36b741e3fbcf","email":"user@example.com","is_active":true,"is_superuser":false,"is_verified":true,"username":"test"}
```

7. Update user
```
$ curl -X 'PATCH'   'http://localhost:8001/latest/user/me'   -H 'accept: application/json'   -H 'Content-Type: application/json'   -H 'Authorization: Bearer TOKEN' -d '{
  "password": "test",
  "email": "user1@example.com",
  "username": "test"
}'
{"id":"650bead8a4de36b741e3fbcf","email":"user1@example.com","is_active":true,"is_superuser":false,"is_verified":true,"username":"test"}
```

8. Get user by id  (accessible by superuser only)
```
$ curl -X 'GET'   'http://localhost:8001/latest/user/650bead8a4de36b741e3fbcf'   -H 'accept: application/json'   -H 'Content-Type: application/json'   -H 'Authorization: Bearer TOKEN'
{"id":"650bead8a4de36b741e3fbcf","email":"user@example.com","is_active":true,"is_superuser":false,"is_verified":false,"username":"test"}
```

9. Update user by id (accessible by superuser only)
```
$ curl -X 'PATCH'   'http://localhost:8001/latest/user/650be1de418d00f4cc61cd2e'   -H 'accept: application/json'   -H 'Content-Type: application/json'   -H 'Authorization: Bearer TOKEN' -d '{"username": "test"}'
{"id":"650be1de418d00f4cc61cd2e","email":"user@example.com","is_active":true,"is_superuser":false,"is_verified":false,"username":"test"}
```

10. Delete user by id (accessible by superuser only)
```
$ curl -X 'DELETE'   'http://localhost:8001/latest/user/650d85d6d6456a3ba3643b05'   -H 'accept: application/json'   -H 'Content-Type: application/json'   -H 'Authorization: Bearer TOKEN'
Returns HTTP 204 No Content
```

JenySadadia:
Added `/user/update-password` endpoint to update password for a user account.

gctucker:
OK thanks.  Now I think it's all in place I'll test and review it on Monday.

gctucker:
The API service failed to start, I guess the migration didn't get applied automatically:
```
pymongo.errors.DuplicateKeyError: Index build failed: 1d9b688a-2f09-4192-955d-0ad38df3905d: Collection kernelci.user ( 8ce1d873-8042-4e69-8062-7f6e42423136 ) :: caused by :: E11000 duplicate key error collection: kernelci.user index: username_1 dup key: { username: null }, full error: {'ok': 0.0, 'errmsg': 'Index build failed: 1d9b688a-2f09-4192-955d-0ad38df3905d: Collection kernelci.user ( 8ce1d873-8042-4e69-8062-7f6e42423136 ) :: caused by :: E11000 duplicate key error collection: kernelci.user index: username_1 dup key: { username: null }', 'code': 11000, 'codeName': 'DuplicateKey', 'keyPattern': {'username': 1}, 'keyValue': {'username': None}}
```

migrations/20231102101356_user.py
gctucker, line 1: The commit message prefix is `migration:` but it should be `migrations:`.  I can fix this after merging with a force-push.

migrations/20231102101356_user.py
```diff
@@ -0,0 +1,36 @@
...
+                "is_superuser": 0,
+                "is_verified": 0,
+                "username": user['profile']['username'],
+                "groups": user['profile']['groups']
+            },
+        )
+
+
+def downgrade(db: "pymongo.database.Database"):
+    pass
```
gctucker, line 36: So `fastapi-users` is a point of no return :)  I guess that's fine as we're not in production yet.  It would be good to test a rollback at some point with the Early Access deployment though.

gctucker:
> The API service failed to start, I guess the migration didn't get applied automatically:
> 
> ```
> pymongo.errors.DuplicateKeyError: Index build failed: 1d9b688a-2f09-4192-955d-0ad38df3905d: Collection kernelci.user ( 8ce1d873-8042-4e69-8062-7f6e42423136 ) :: caused by :: E11000 duplicate key error collection: kernelci.user index: username_1 dup key: { username: null }, full error: {'ok': 0.0, 'errmsg': 'Index build failed: 1d9b688a-2f09-4192-955d-0ad38df3905d: Collection kernelci.user ( 8ce1d873-8042-4e69-8062-7f6e42423136 ) :: caused by :: E11000 duplicate key error collection: kernelci.user index: username_1 dup key: { username: null }', 'code': 11000, 'codeName': 'DuplicateKey', 'keyPattern': {'username': 1}, 'keyValue': {'username': None}}
> ```

Ah I see migrations need to be initiated manually:
```
docker-compose exec api /bin/sh -c 'pymongo-migrate migrate -u mongodb://db/kernelci -m migrations -v'
```
I was expecting them to run automatically when the API service starts, assuming it can keep track of migrations already applied.  I'll create a GitHub issue for this.

gctucker:
Actually this should be a `docker-compose run api ...` command as it can't be run with a running container since the container needs the migrations to have been applied in order to start.  Will also fix the docs :)

migrations/20231102101356_user.py
nuclearcat: We can deal with that different way (with mongodb snapshots and loss of some data that was submitted after upgrade), often it might be cheaper than implement complex downgrade (which might be also not always possible).

nuclearcat:
I prefer to avoid any exec, because with kubernetes it will make migration a bit non-trivial and too much manual. Imho on startup some script should check mongodb schema version and execute migration scripts accordingly.

gctucker:
No I mean, it should just be run on startup (see my previous comment).  But for running manually, it should be `docker-compose run` rather than `docker-compose exec`.

gctucker:
@nuclearcat Feel free to work on enabling the migrations to run during startup btw, I'll look just at this PR for now.  I've prepared an update in the docs for running the migrations manually but that's orthogonal I guess.

nuclearcat:
Ok, i will try to do that after will complete initial deployment scripts, on second revision it should be part of deployment

nuclearcat:
@JenySadadia we need to have a way (config option, environment variable, daemon flag, anything) to disable mandatory email verification, it is a bit difficult to implement this on staging for now.

api/main.py
gctucker, line 1: The commit summary should be in the present tense i.e. `rename` rather than `renamed`.  Also something that can be fixed after merging this PR.

.github/workflows/main.yml
```diff
@@ -54,13 +54,9 @@ jobs:
...
           pylint tests/unit_tests
 
-      - name: Export environment variables
-        run: |
-          echo "SECRET_KEY=$(openssl rand -hex 32)" >> $GITHUB_ENV
-
       - name: Run pytest
         run: |
-          pytest -v tests/unit_tests
+          SECRET_KEY=$(openssl rand -hex 32) SMTP_HOST=smtp.gmail.com SMTP_PORT=465 EMAIL_SENDER=test@kernelci.org EMAIL_PASSWORD=random pytest -vs tests/unit_tests/
```
gctucker, line 59: Also, ideally the key should be fixed rather than recreated every time the test is run.  In some particular corner cases, using truly random data in tests might cause false positives that can't be reproduced.

migrations/20231102101356_user.py
gctucker: The downgrade has the advantage of being usable with new data.

migrations/20231102101356_user.py
```diff
@@ -0,0 +1,36 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2023 Collabora Limited
+# Author: Jeny Sadadia <jeny.sadadia@collabora.com>
+
+"""
+Migration for User schema
```
gctucker, line 7: Here as well, it might be better to mention `fastapi-users`.

api/email_sender.py
gctucker, line 1: The commit message for fixing the staging bug doesn't explain what this does.  It looks like it's temporary so should it really be merged?  This looks like a staging configuration issue, or maybe the API code should be able to deal with such cases and not try to send emails when the sender is not valid.

gctucker (commented):
OK LGTM overall.  As it's a large PR there are a few things that could be improved.  I just used it locally and migrated my users fine, and could authenticate with the API.

Next I'll try to create / update / find users with `kci user` to test this completely.  If that works then I think it's good to go, with a few things to clean-up once it's merged.

gctucker (approved):
Tested OK with local instance.  A few things need to be tweaked, I'll send some follow-up PRs for that.

.github/workflows/main.yml
gctucker: See #402

api/admin.py
```diff
@@ -42,19 +43,18 @@ async def setup_admin_user(db, username, email, admin_group):
...
-    return await db.create(User(
-        profile=profile
+        groups=[admin_group],
+        is_superuser=1
     ))
 
 
 async def main(args):
     db = Database(args.mongo, args.database)
+    await db.initialize_beanie()
```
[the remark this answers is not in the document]
JenySadadia: We are accessing DB to create an admin user with `db.create`. Also, access DB for finding existing users with the provided `username` in `db.find_one_by_attributes`.
For these methods to work, we need to initialize Beanie first as they are accessing the collection `user` created by Beanie.

PR #383, opened 2023-10-16
README.md: fix `API Testing` section for unit tests

JenySadadia:
Test directory structure has been changed
to have two separate directories for unit tests and e2e tests named `unit_tests` and `e2e_tests` respectively. Fix command to run unit tests in the README.md file by renaming test directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ export $(cat .env)
 
 Run the below command from kernelci-api directory:
 ```
-pytest -v test/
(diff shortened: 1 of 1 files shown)


gctucker (approved):
And I see the GitHub workflow has already been updated so it looks all good.

By the way, this is one reason why having `make test` avoids updating several places when changing things with an extra level of indirection like in `kernelci-core`.

Issue #371, opened 2023-10-03
Define node types


hardboprobot:
Related to #370 and maybe to #204 as well

It's not clear how to distinguish different node types. For instance, how to tell if a node is a test result?

Example:
from this node:

```
$ kci show results 651162e28326c545a7810637
kci show results 651162e28326c545a7810637
Node
  path      checkout
  id        651162e28326c545a7810637
  parent    ----
  owner     nuclearcat
  created   2023-09-25 at 10:37:22
  state     done

Kernel
  tree      kernelci
  branch    staging-mainline
  commit    4abc3c0a156b6aabb774334652fbfd505550b525
  describe  staging-mainline-20230925.0

...

Results
checkout                                                        ----  651162e28326c545a7810637
  kbuild-gcc-10-x86                                             pass  651163a68326c545a7810638
    config                                                      pass  651164b18326c545a781078b
    kernel                                                      pass  651164b18326c545a781078c
    modules                                                     pass  651164b18326c545a781078d
    modules_install                                             pass  651164b18326c545a781078e
    modules_tarball                                             pass  651164b18326c545a781078f
    baseline-x86                                                ----  651164b18326c545a7810790
    baseline-x86                                                ----  651164b28326c545a7810791
  kbuild-gcc-10-x86                                             ----  651163a78326c545a7810639
  kver                                                          ----  651163a78326c545a781063a
  kunit-x86_64                                                  ----  651163a78326c545a781063b
  kunit-x86_64                                                  pass  651163a78326c545a781063c
    config                                                      pass  651164268326c545a781063e
    build                                                       pass  651164268326c545a781063f
    exec                                                        pass  651164268326c545a7810640
      time_test_cases                                           ----  651164268326c545a7810641
        time64_to_tm_test_date_range                            pass  651164268326c545a7810642
      hw_breakpoint                                             ----  651164268326c545a7810643
        test_one_cpu                                            ----  651164268326c545a7810644
        test_many_cpus                                          ----  651164268326c545a7810645
        test_one_task_on_all_cpus                               ----  651164268326c545a7810646
        test_two_tasks_on_all_cpus                              ----  651164268326c545a7810647
        test_one_task_on_one_cpu                                ----  651164268326c545a7810648
        test_one_task_mixed                                     ----  651164268326c545a7810649
        test_two_tasks_on_one_cpu                               ----  651164268326c545a781064a
        test_two_tasks_on_one_all_cpus                          ----  651164268326c545a781064b
[...]

PR #388, opened 2023-10-23
api.models: re-format Regression data field

hardboprobot:
Redefine the Regression data field so that it can be defined as a dict with a format that's validated by Pydantic.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -369,6 +369,12 @@ class Hierarchy(BaseModel):
 Hierarchy.update_forward_refs()
 
 
+class RegressionData(BaseModel):
+    """Model for the data field of a Regression"""
+    node_failed: PyObjectId
+    node_passed: PyObjectId
+
+
 class Regression(Node):
     """API model for regression tracking"""
 
@@ -377,18 +383,14 @@ class Regression(Node):
         description='Type of the object',
         const=True
     )
-    regression_data: List[Node] = Field(
-        description='Regression details'
+
+    data: RegressionData = Field(
+        description="Regression details"
     )
 
     _OBJECT_ID_FIELDS = Node._OBJECT_ID_FIELDS + [
-        'regression_data.parent',
-    ]
-    _TIMESTAMP_FIELDS = Node._TIMESTAMP_FIELDS + [
-        'regression_data.created',
-        'regression_data.updated',
-        'regression_data.timeout',
-        'regression_data.holdoff',
+        'data.node_failed',
+        'data.node_passed',
     ]
 
 


gctucker (commented):
The issue with this is I think it's going to break at the API level, as discussed previously.  You can't submit a subclass of Node via the `/node` API entry point, and it would seem like a better approach to rely on the `.kind` field to validate the contents of `.data`.

api/models.py
```diff
@@ -377,20 +382,24 @@ class Regression(Node):
...
     ]
 
+    # [TODO] Do we need these?
+    # _TIMESTAMP_FIELDS = Node._TIMESTAMP_FIELDS + [
+    #     'data.created',
+    #     'data.updated',
+    #     'data.timeout',
+    #     'data.holdoff',
+    # ]
+
```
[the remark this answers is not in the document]
hardboprobot: Ok, I'll remove that once we move past the review stage

gctucker:
Yes that's not a good implementation, I guess I'll create an issue or just make a PR.

README.md
```diff
@@ -67,7 +67,7 @@ export $(cat .env)
 
 Run the below command from kernelci-api directory:
 ```
-pytest -v test/
+pytest -v tests/unit_tests/
```
JenySadadia, line 70: This is already fixed.
Please update your fork.

api/models.py
```diff
@@ -369,6 +369,12 @@ class Hierarchy(BaseModel):
 Hierarchy.update_forward_refs()
 
 
+class RegressionData(BaseModel):
+    """Model for the data field of a Regression"""
+    testrun_failed: PyObjectId
+    testrun_passed: PyObjectId
```
JenySadadia, line 375: Field names are not aligned with https://github.com/kernelci/kernelci-pipeline/pull/343/

JenySadadia: The reason for having embedded models here rather than just an object ID was to reduce DB find operations for further processing.

api/models.py
```diff
@@ -377,18 +383,14 @@ class Regression(Node):
...
     _OBJECT_ID_FIELDS = Node._OBJECT_ID_FIELDS + [
-        'regression_data.parent',
-    ]
-    _TIMESTAMP_FIELDS = Node._TIMESTAMP_FIELDS + [
-        'regression_data.created',
-        'regression_data.updated',
-        'regression_data.timeout',
-        'regression_data.holdoff',
+        'data.testrun_failed',
+        'data.testrun_passed',
```
JenySadadia, line 393: These are not timestamp fields. `_TIMESTAMP_FIELDS` are defined to translate timestamp fields for querying nodes.
Please see `Node._translate_timestamps` method for the implementation.

JenySadadia:
> Redefine the Regression data field so that it can be defined as a dict with a format that's validated by Pydantic.

I am not sure about the exact requirement here.
`regression_data: List[Node]` can also be validated by Pydantic.

JenySadadia:
Added `staging-skip` as this broke regression node query on staging.
```
$ curl https://staging.kernelci.org:9000/latest/nodes?kind=regression
{"detail":"1 validation error for Regression\ndata\n  none is not an allowed value (type=type_error.none.not_allowed)"}
```

api/models.py
hardboprobot: No, they arent. These are added to `_OBJECT_ID_FIELDS`, not `_TIMESTAMP_FIELDS`. I actually removed `_TIMESTAMP_FIELDS` because I think we have to put more thought into that.

First of all, what did the original implementation do? Did the original `regression.data` (a List) have `created`, `updated`, `timeout` and `holdoff` fields?

JenySadadia:
> > > Redefine the Regression data field so that it can be defined as a dict with a format that's validated by Pydantic.
> > 
> > 
> > I am not sure about the exact requirement here. `regression_data: List[Node]` can also be validated by Pydantic.
> 
> This comes from a [discussion](https://github.com/kernelci/kernelci-api/issues/371) about how to define node types and how to differentiate them. One of the approaches was to differentiate them based on `node.kind` and have `node.data` as the specific "payload" for a node.
> 
> `regression_data` as a List can be validated but it gives no information about its contents at the API level. How many nodes are supposed to be in the list? What do they mean? Having the regression info stored in `node.data` (Dict) allows us to add arbitrary data as needed and have it validated with a specific format (field names and types).

Thanks for the issue link. The whole point makes sense now.
Bdw we can directly eliminate `Regression` model and just use appropriate `kind` and `data` for regression nodes.

api/models.py
JenySadadia: Oh, sorry. I made a mistake while reading `diff`.
 
> First of all, what did the original implementation do?

When you send GET `/nodes` query with attributes, all the request query param values will be received as strings by default. And those values will be used to find matching nodes with `db.find_by_attributes`. See https://github.com/kernelci/kernelci-api/blob/main/api/main.py#L493.
It is not an issue when attributes are also string type in `Node` model such as `kind` or `name`. But when we send attributes having a different type such as `ObjectId` or `datetime`, a conversion from string to actual data type will be required in order to get items from DB.
e.g. `https://staging.kernelci.org:9000/latest/nodes?created=2023-07-17T11:52:31.913000`

That's when `Node._translate_object_ids` and `Node._translate_timestamps` come into the picture.

JenySadadia:
> Were there any regression nodes in the DB already?

Yes, there are regression nodes in staging already. That's why the schema change from this PR is breaking staging.
We can have a migration file (see https://github.com/kernelci/kernelci-api/blob/main/migrations/20221014061654_set_timeout.py) or drop all the staging regression nodes.

api/models.py
hardboprobot: There's a tradeoff here: having the embedded node models will reduce the amount of DB queries but will increase the size of each Regression node. I opted to include the object IDs instead of the full nodes because it looked more coherent with how other models work (for instance, `node.parent` also saves an object ID rather than a full node).

api/models.py
hardboprobot: Thanks for the explanation. What I don't understand is that `_TIMESTAMP_FIELDS` was defined as 

```
_TIMESTAMP_FIELDS = Node._TIMESTAMP_FIELDS + [
        'regression_data.created',
        'regression_data.updated',
        'regression_data.timeout',
        'regression_data.holdoff',
```

with `regression_data` being a `List` of `Node`. The `Node` model implements the `created`, `updated`, `timeout` and `holdoff` fields, but how does accessing these on the _list_ rather than on an individual _node_ work?

JenySadadia: `collection.find( )` works with an array of embedded documents with query filters like this.
Please see https://www.mongodb.com/docs/manual/tutorial/query-array-of-documents/

Also, you can experiment with it in your `mongo` shell.
For instance, you can try out:
```
db.regression.find({"regression_data.created": ISODate("2023-05-29T11:55:05.462Z")})
```

hardboprobot: Great, thanks for the explanation. Don't worry about the changes anyway, this is just an experiment to raise a discussion. The definitive code changes will take this into account.

PR #389, opened 2023-10-23
HTTP redirect error handling

gctucker:
Return HTTP 400 when trying to use a version that doesn't exist e.g. `http://host/v1/`

Follow-up from #367

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -9,6 +9,7 @@
 """KernelCI API main module"""
 
 import os
+import re
 from typing import List, Union
 from fastapi import (
     Depends,
@@ -19,7 +20,7 @@ from fastapi import (
     Security,
     Query,
 )
-from fastapi.responses import JSONResponse
+from fastapi.responses import JSONResponse, PlainTextResponse
 from fastapi.security import (
     OAuth2PasswordRequestForm,
     SecurityScopes
@@ -46,7 +47,7 @@ from .pubsub import PubSub, Subscription
 # List of all the supported API versions.  This is a placeholder until the API
 # actually supports multiple versions with different sets of endpoints and
 # models etc.
-API_VERSIONS = ['latest', 'v0']
+API_VERSIONS = ['v0']
 
 app = FastAPI()
 db = Database(service=(os.getenv('MONGO_SERVICE') or 'mongodb://db:27017'))
@@ -711,9 +712,16 @@ for sub_app in app.routes:
 @app.middleware("http")
 async def redirect_http_requests(request: Request, call_next):
     """Redirect request with version prefix when no version is provided"""
+    response = None
     path = request.scope['path']
-    prefix = path.split('/')[1]
-    if prefix not in API_VERSIONS:
+    match = re.match(r'^/(v[\d.]+)', path)
+    if match:
+        prefix = match.group(1)
+        if prefix not in API_VERSIONS:
+            response = PlainTextResponse(
+                f"Unsupported API version: {prefix}",
+                status_code=status.HTTP_400_BAD_REQUEST
+            )
+    elif not path.startswith('/latest'):
         request.scope['path'] = '/latest' + path
-    response = await call_next(request)
-    return response
+    return response or await call_next(request)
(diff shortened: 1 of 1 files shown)


api/main.py
```diff
@@ -715,9 +712,16 @@ async def put_regression(regression_id: str, regression: Regression,
...
     path = request.scope['path']
-    prefix = path.split('/')[1]
-    if prefix not in API_VERSIONS:
+    match = re.match(r'^/(v\d+)', path)
+    if match:
+        prefix = match.group(1)
+        if prefix not in API_VERSIONS:
+            response = JSONResponse(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                content=f"Unsupported API version: {prefix}",
```
JenySadadia, line 723: It would be good to have `detail` as a key to enable pipeline service parse error messages correctly.
```suggestion
                content={"detail": f"Unsupported API version: {prefix}"},
```
Exception handler in pipeline:
```
except requests.exceptions.HTTPError as ex:
            detail = ex.response.json().get('detail')
            if detail:
                self.log.error(detail)
```

api/main.py
```diff
@@ -715,9 +712,16 @@ async def put_regression(regression_id: str, regression: Regression,
 @app.middleware("http")
 async def redirect_http_requests(request: Request, call_next):
     """Redirect request with version prefix when no version is provided"""
+    response = None
     path = request.scope['path']
-    prefix = path.split('/')[1]
-    if prefix not in API_VERSIONS:
+    match = re.match(r'^/(v\d+)', path)
```
JenySadadia, line 717: We should also consider optional minor number in the version as we might add it in the future.

gctucker: Yes but right now it's not supported and I think we agreed to only have major numbers, so I left that out explicitly.  It's true that adding support for it in this check right now shouldn't cause side-effects, in fact the current regex would already match things like `v1.2.3`.  So yeah I'll update this :)

api/main.py
gctucker: Yes it's actually coming from here:
https://github.com/encode/starlette/blob/master/starlette/middleware/exceptions.py#L64

So in this case, we can't raise an `HTTPException` in the FastAPI middleware so we need to emulate the `starlette` implementation basically.

gctucker: I've updated the code with a `PlainTextException`.  Tested OK manually with `curl`:
```
$ curl http://172.17.0.1:8001/v0/ -v
*   Trying 172.17.0.1:8001...
* Connected to 172.17.0.1 (172.17.0.1) port 8001 (#0)
> GET /v0/ HTTP/1.1
> Host: 172.17.0.1:8001
> User-Agent: curl/7.88.1
> Accept: */*
> 
< HTTP/1.1 200 OK
< date: Wed, 25 Oct 2023 14:19:56 GMT
< server: uvicorn
< content-length: 26
< content-type: application/json
< 
* Connection #0 to host 172.17.0.1 left intact
{"message":"KernelCI API"}
```
```
$ curl http://172.17.0.1:8001/v0.1/ -v
*   Trying 172.17.0.1:8001...
* Connected to 172.17.0.1 (172.17.0.1) port 8001 (#0)
> GET /v0.1/ HTTP/1.1
> Host: 172.17.0.1:8001
> User-Agent: curl/7.88.1
> Accept: */*
> 
< HTTP/1.1 400 Bad Request
< date: Wed, 25 Oct 2023 14:20:24 GMT
< server: uvicorn
< content-length: 29
< content-type: text/plain; charset=utf-8
< 
* Connection #0 to host 172.17.0.1 left intact
Unsupported API version: v0.1
```
I'll see if this can be added to the unit tests, probably once the `fastapi-users` changes have landed to avoid unnecessary conflicts.

JenySadadia: I don't think it is required here to change it to `PlainTextException` as It still worked as `JSONResponse` when I tested it locally with the pipeline. I suggested tweaking `content` though.

gctucker: No I mean I think it should have been `PlainTextResponse` everywhere here rather than `JSONResponse`.  `PlainTextResponse` has a `details` attribute and it's meant to be used for this kind of scenario, `JSONResponse` is a lower-level raw structure.

JenySadadia (approved):
Tested OK on staging.

PR #391, opened 2023-10-25
unit tests hacks

gctucker:
Based on top of https://github.com/kernelci/kernelci-api/pull/377

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,10 @@ jobs:
       - name: Check out source code
         uses: actions/checkout@v2
 
+      - name: Export environment variables
+        run: |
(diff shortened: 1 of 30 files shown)


.github/workflows/test.yml
```diff
@@ -29,7 +29,7 @@ jobs:
 
       - name: Export environment variables
         run: |
-          echo "SECRET_KEY=$(openssl rand -hex 32)" > .env
```
JenySadadia, line 32: This is used for e2e tests.
For uni tests, we export local `.env` file.
See https://github.com/kernelci/kernelci-api#api-testing

gctucker: I know

PR #392, opened 2023-10-30
Add `api.config` module for `Settings`

JenySadadia:
On top of https://github.com/kernelci/kernelci-api/pull/377

api: add `api.config` to define various module settings
    
Instead of defining multiple settings class in different API modules like `pubsub`, `auth`, `email_sender`, add `api.config` module to store all different settings classes.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -57,6 +57,10 @@ jobs:
       - name: Export environment variables
         run: |
           echo "SECRET_KEY=$(openssl rand -hex 32)" >> $GITHUB_ENV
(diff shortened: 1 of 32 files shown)


gctucker:
OK I've tested this which includes the changes from #377 and it's working with a few minor fixes and tweaks.  I'll prepare the follow-up PRs and merge these two so we should have all this in place today, as well as the related `kernelci-core` changes with `kci user` etc.

gctucker (commented):
Arguably this should have been called `api.settings` but that's a detail.

PR #399, opened 2023-11-09
doc: fix multiple config paths setting on Early Access page

gctucker:
The config settings (--yaml-config) can take several values, but the first one should be the standard one in /etc/kernelci/core when adding a custom file on top of it.  Fix this instead of relying on a local config/core copy of the files.

Fixes: e8e94267a669 ("doc: update Early Access page with latest kci")


nuclearcat (approved):
Looks good

PR #400, opened 2023-11-09
doc: add notes about Azure File share names

gctucker:
The Azure File name might not be exactly the same as the username so add a note about that.  Also add a related note about putting values in double quotes when they contain a dot so it gets parsed correctly in TOML with the dotted-key syntax.

### DIFF
--- a/doc/early-access.md
+++ b/doc/early-access.md
@@ -143,6 +143,10 @@ storage:
     share: <your-username>
 ```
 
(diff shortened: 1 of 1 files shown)


hardboprobot (approved):

gctucker:
Just noticed a typo in the commit message, it's a "note" and not a "node".  Sending an update now.

nuclearcat (approved):
Looks good

PR #403, opened 2023-11-10
api.config: add module to define various settings (was #392)

gctucker:
Instead of defining multiple settings class in different API modules like `pubsub`, `auth`, `email_sender`, add `api.config` module to store all different settings classes.

Replaces #392

### DIFF
--- a/api/pubsub.py
+++ b/api/pubsub.py
@@ -9,15 +9,8 @@ import asyncio
 
 import aioredis
 from cloudevents.http import CloudEvent, to_json
-from pydantic import BaseModel, BaseSettings, Field
-
-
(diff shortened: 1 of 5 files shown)


api/pubsub.py
```diff
@@ -49,7 +42,8 @@ async def create(cls, *args, **kwargs):
         return pubsub
 
     def __init__(self, host=None, db_number=None):
-        self._settings = Settings()
+        # self._settings = Settings()
```
JenySadadia, line 45: This comment can be dropped now.
I added it when I was testing the changes.

JenySadadia (approved):

PR #405, opened 2023-11-10
api.user_manager: create EmailSender only when required

gctucker:
Convert self._email_sender to an attribute and only create the object when required.  This way, the API can be run without email support as long as no endpoints that rely on email features are used.


gctucker:
Replacing with new PR with a few more fixes bundled together.

PR #406, opened 2023-11-10
Fix several issues related to user management

gctucker:
This PR fixes several issues found when testing the new user management implementation:

* `api.user_manager: call UserManager parent with all arguments`
  This is to avoid unexpected behaviour and fix hidden bugs.
* `api.main: rely on create_user_manager()`
  This is to have only one way of creating the UserManager objects and simplify the code
* `api.user_manager: create EmailSender only when required`
  This is to avoid requiring SMTP settings unless when actually trying to send an email
* `api.main: only look for duplicate user when required`
  This is to remove unwanted checks for duplicate users when no new username is being provided

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -220,12 +220,13 @@ async def update_me(request: Request, user: UserUpdate,
     Custom user update router handler will only allow users to update
     its own profile. Adding itself to 'admin' group is not allowed.
     """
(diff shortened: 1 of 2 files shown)


JenySadadia (approved):
LGTM.

JenySadadia:
Tested OK locally.

PR #407, opened 2023-11-10
api.admin: set is_verified=1 for the first superuser

gctucker:
Bypass the email verification step when creating the first superuser as this is a special case.  Administrators should be able to create this entry entirely non-interactively, and they should also have direct access to the database anyway where this flag could be set by hand.


JenySadadia (approved):
Tested OK locally.

PR #408, opened 2023-11-10
api.user_manager: add username in log messages

gctucker:
Show the username in log messages in addition to the user ID.  This makes it much easier to follow the logs when checking user activity.


JenySadadia (approved):
LGTM. Let's wait for staging results.

PR #412, opened 2023-11-15
api.admin: fix `setup_admin_user`

### LINKED ISSUE #401: User email field is missing unique index
With the new user management implementation based on `fastapi-users`, there doesn't appear to be any constraint for the user email addresses to be unique.  I've been able to create several users with the same email, then when doing things like a password reset which takes the email address as input a random user that had a matching email address was selected in the database.

JenySadadia:
Since `profile` field has been eliminated from `User` schema, update logic to check existing admin user by matching `username` field.

Fixes #401


nuclearcat (approved):
LGTM

PR #413, opened 2023-11-15
api.admin: fix method for setting node ownership

JenySadadia:
Fix `_set_node_ownership_recursively` method to use `User.username` to get username as per the latest `User` schema while setting node ownership.


nuclearcat (approved):
LGTM, tested on staging

Issue #390, opened 2023-10-25
Enable find users with boolean query filters


JenySadadia:
Once the `fastapi-users` integration is completed and https://github.com/kernelci/kernelci-api/pull/377/ is merged, enable
support for boolean query filters for `/users` endpoint.
e.g. `http://localhost:8001/latest/users?is_superuser=true` and `http://localhost:8001/latest/users?is_verified=false` should work fine.

- [ ] Add translation from string to boolean for database query in `api.db.Database`
- [ ] Tested OK on staging

PR #415, opened 2023-11-17
api.db: enable boolean query fields for GET requests

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/390

Enable support for querying boolean fields while getting objects from API such as nodes or models.
e.g. find all the active admin users using
`http://API_SERVER/users?is_active=true&is_superuser=true`


JenySadadia:
Tested OK on staging:
```
$ curl https://staging.kernelci.org:9000/latest/users?is_verified=true
{"items":[{"id":"62554673a68f9bac63a0ead2","email":"guillaume.tucker@collabora.com","is_active":true,"is_superuser":false,"is_verified":true,"username":"gtucker","groups":[]},{"id":"62f3ec222d0a223569d55be7","email":"kernelci-tsc@groups.io","is_active":true,"is_superuser":true,"is_verified":true,"username":"admin","groups":[{"id":"649190b01432ccfdb575a250","name":"admin"}]},{"id":"654b618eb5e6b3443a48e55b","email":"denys.f@collabora.com","is_active":true,"is_superuser":false,"is_verified":true,"username":"nuclearcat","groups":[]}],"total":3,"limit":50,"offset":0}
```

api/db.py
```diff
@@ -103,10 +108,18 @@ def _convert_int_values(cls, attributes):
             if isinstance(val, tuple) and len(val) == 2 and val[0] == 'int'
         }
 
+    def _convert_bool_values(self, attributes):
+        return {
+            key: self.BOOL_VALUE_MAP[val.lower()]
+            for key, val in attributes.items()
+            if isinstance(val, str) and val.lower() in self.BOOL_VALUE_MAP
+        }
```
gctucker, line 116: It's not the optimal way to implement it as it's converting the value to lower-case twice and doing a lookup in BOOL_VALUE_MAP twice - but that's not a big deal.  I think it could be improved with either a nested comprehension or a plain for-loop e.g.:
```python
for key, val in attributes.items():
    val_bool = self.BOOL_VALUE_MAP.get(val.lower())
    if val_bool is not None:
        ret[key] = val_bool
```
Also I believe all the attributes should be strings, or is FastAPI converting some values to other types?  The `_convert_int_values()` function is there specifically for this kind of reason.

gctucker (commented):
LGTM - some minor comments.

api/db.py
JenySadadia: > It's not the optimal way to implement it as it's converting the value to lower-case twice and doing a lookup in BOOL_VALUE_MAP twice

Fixed it.

nuclearcat:
LGTM, tested by curl.
Merging, but as follow-up i guess we need to support that in cli as well, and i guess it will be tricky, because at moment:
```
./kci user find is_active=true
[]
```

JenySadadia:
> LGTM, tested by curl. Merging, but as follow-up i guess we need to support that in cli as well, and i guess it will be tricky, because at moment:
> 
> ```
> ./kci user find is_active=true
> []
> ```

Works for me:
```
$ ./kci user find is_superuser=true
[{"id": "6526448e7d140ee220971a0e", "email": "admin@gmail.com", "is_active": true, "is_superuser": true, "is_verified": true, "username": "admin", "groups": [{"id": "648c07a70fccad400a122509", "name": "admin"}, {"id": "648ff894bd39930355ed16ad", "name": "kernelci"}]}]

```

PR #417, opened 2023-11-22
Enable support for unicast messages

JenySadadia:
Use Redis list for enabling support for unicast messages for load-balancing use-case.
Introduced endpoints `/push` and `/pop` for pushing messages to a Redis list and popping a message from the
list respectively.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -617,6 +617,21 @@ async def publish(raw: dict, channel: str,
     await pubsub.publish_cloudevent(channel, data, attributes)
 
 
+@app.post('/push/{list_name}')
+async def push(raw: dict, list_name: str,
+               user: User = Depends(get_current_user)):
+    """Push a message on the provided list"""
+    attributes = dict(raw)
+    data = attributes.pop('data')
+    await pubsub.push_cloudevent(list_name, data, attributes)
+
+
+@app.get('/pop/{list_name}')
+async def pop(list_name: str, user: User = Depends(get_current_user)):
+    """Pop a message from a given list"""
+    return await pubsub.pop(list_name)
+
+
 # -----------------------------------------------------------------------------
 # Regression
 
--- a/api/pubsub.py
+++ b/api/pubsub.py
@@ -7,7 +7,7 @@
 
 import asyncio
 
-import aioredis
+from redis import asyncio as aioredis
 from cloudevents.http import CloudEvent, to_json
 from pydantic import BaseModel, Field
 from .config import PubSubSettings
--- a/api/pubsub.py
+++ b/api/pubsub.py
@@ -7,6 +7,7 @@
 
 import asyncio
 
+import json
 from redis import asyncio as aioredis
 from cloudevents.http import CloudEvent, to_json
 from pydantic import BaseModel, Field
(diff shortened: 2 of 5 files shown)


JenySadadia:
I have been investigating unit test failure in `test_pubsub`.
Somehow after changing `fakeredis` version to `2.20.0`, redis mock stays the same for all the tests even if `mock_pubsub` fixture has the default scope `function`. That's why value of `_redis.ID_KEY` key becomes global among all the tests and `pubsub.subscribe` increments it based on the value from the previous test run.
```
@pytest.fixture
def mock_pubsub(mocker):
    """Mocks `_redis` member of PubSub class instance"""
    pubsub = PubSub()
    redis_mock = fakeredis.aioredis.FakeRedis()
    mocker.patch.object(pubsub, '_redis', redis_mock)
    return pubsub
```

First, `test_subscribe_single_channel` test will be run and `_redis.ID_KEY` will become 1 after the execution.
```
async def test_subscribe_single_channel(mock_pubsub):
    """
    Test Case: Subscribe for one channel with a PubSub.subscribe() method.

    Expected Result:
        Subscription-like object is returned from PubSub.subscribe() method
        with channel set to 'CHANNEL' and id set to 1.
        PubSub._subscriptions dict should have one entry. This entry's
        key should be equal 1.
    """
    result = await mock_pubsub.subscribe('CHANNEL')
    assert result.channel == 'CHANNEL'
    assert result.id == 1
    assert len(mock_pubsub._subscriptions) == 1
    assert 1 in mock_pubsub._subscriptions
```
When `test_subscribe_multiple_channels` executes after it, `mock_pubsub._redis` doesn't get re-instantiated and uses the same instance from the previous test. This results in test failure as `_redis.ID_KEY` has been set to 1 in the previous run and now it becomes 2.
```
async def test_subscribe_multiple_channels(mock_pubsub):
    """
    Test Case: Subscribe for three channels with subsequent calls of
    PubSub.subscribe() method.

    Expected Result:
        Subscription-like objects are returned for each call of
        PubSub.subscribe() method
        Subsequent calls should have channel names: 'CHANNEL1', 'CHANNEL2',
        'CHANNEL3' and should have ids 1, 2, 3 respectively.
        PubSub._subscriptions dict should have 2 entries. This entries'
        keys should be 1, 2, and 3.
    """
    channels = ((1, 'CHANNEL1'), (2, 'CHANNEL2'), (3, 'CHANNEL3'))
    for expected_id, expected_channel in channels:
        result = await mock_pubsub.subscribe(expected_channel)
        assert result.channel == expected_channel
        assert result.id == expected_id
    assert len(mock_pubsub._subscriptions) == 3
    assert (1, 2, 3) == tuple(mock_pubsub._subscriptions.keys())
``` 
This seems some kind of mock leak.

JenySadadia:
It fixed the issue when I reset `ID_KEY` in `test_subscribe_multiple_channels` but need to investigate the root cause.

```
await mock_pubsub._redis.set(mock_pubsub.ID_KEY, 0)`
```

gctucker (commented):
LGTM - The only issue is what we discussed in a meeting last week, there needs to be a way to close the queue or we'll have resource leaks in Redis.  That's why I suggested to have the open / close endpoints (and `kci` commands).  I'll create a GitHub issue for it.

gctucker (approved):
OK following the merge-and-fix approach to untangle the various pending PRs and conflicts with `kernelci-core` related changes.

Issue #402, opened 2023-11-10
Secret key shouldn't be randomly generated in tests


gctucker:
The tests run on GitHub currently have a secret key for the test instance generated randomly on the fly.  This is not best practice for automated tests as it means each run will be with a different key, which might lead to corner-case problems that can't be reproducible.  Instead, the key should just be hard-coded with a comment explaining how it was generated.

JenySadadia:
Yes, I agree. We should have all the settings for tests hard-coded.

PR #418, opened 2023-11-24
github/workflows: use hard-coded `SECRET_KEY` for tests

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/402

Instead of generating `SECRET_KEY` value for every test run, use hard-coded value to reduce test failures and easy debugging for automated unit tests and e2e tests.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,7 +56,7 @@ jobs:
 
       - name: Run pytest
         run: |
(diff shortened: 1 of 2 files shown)


gctucker (commented):
It would be good to be able to run the tests without the GitHub action though.  A follow-up would be to have the environment variables part of the code instead, in the unit tests.  I think there's already something going in this direction with the reworked settings class.

PR #419, opened 2023-11-28
api.main: [trivial] handle runtime errors properly in listen endpoint

hardboprobot:
Certain scenarios may trigger a runtime error on the <listen> endpoint, particularly during development and testing. Handle them so the request returns an error message instead of crashing.


nuclearcat:
Tested on staging

PR #420, opened 2023-11-28
api.models: define basic pydantic models for publish events

hardboprobot:


nuclearcat:
Looks like working fine on staging, fixes crash on corrupted data, thanks!

PR #421, opened 2023-11-29
Implement endpoint to delete a user group

JenySadadia:
implement DELETE `/group/<group-id>` endpoint to delete a user group.
Remove users from that group before deleting it. Also, check if the group is assigned to any node and remove it from user groups that are permitted to update the node.

### DIFF
--- a/api/db.py
+++ b/api/db.py
@@ -194,3 +194,10 @@ class Database:
         if res.matched_count == 0:
             raise ValueError(f"No object found with id: {obj.id}")
         return obj.__class__(**await col.find_one(ObjectId(obj.id)))
+
+    async def delete_by_id(self, model, obj_id):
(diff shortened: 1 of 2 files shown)


JenySadadia:
In order to use `db.find_by_attributes`, I needed to add `response_model=PageModel` to the endpoint event if the endpoint is not returning anything.
This is because the DB function is using `paginate` method from `fastapi-pagination` and I need to mention pagination model in the `response_model` as per the package requirements.
The new version of `fastapi-pagination` has fixed the issue. I'll create a GH issue to upgrade the package version.

gctucker (approved):
OK following the merge-and-fix approach to untangle the various pending PRs and conflicts with `kernelci-core` related changes.

PR #430, opened 2023-12-08
api.main: perform authorization for GET '/users'

JenySadadia:
Only allow authorized users to get existing users' details to protect user's personal information.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -318,7 +318,8 @@ async def authorize_user(node_id: str,
 @app.get('/users', response_model=PageModel, tags=["user"],
          response_model_exclude={"items": {"__all__": {
                                     "hashed_password"}}})
-async def get_users(request: Request):
+async def get_users(request: Request,
+                    current_user: User = Depends(get_current_user)):
     """Get all the users if no request parameters have passed.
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
It works! :)

gctucker:
But, this should only be returning public information.  I think it's a step backwards, being able to find user information publicly makes sense and all public web services do this.  Also, if all it takes is a user account then I would argue it's a false sense of security as it's not preventing any personal user data from being shared with other individuals.

nuclearcat:
Email is not public information, at least. I think it is better to move this discussion to Slack.

gctucker:
I can see your email address on your public profile here ;) https://github.com/nuclearcat

nuclearcat:
Yes, that intentional and github explicitly shows it will be published on github profile page.
But even this one good as example, it is my personal address for low priority stuff, but not the one i use for github authentication, notifications and important business emails.

gctucker:
Yes, I know :)  As discussed on Slack now.  Maybe a GitHub issue or discussion about how to manage users in general would be good actually.

PR #433, opened 2023-12-14
Refactor user-facing API models

hardboprobot:
Related to https://github.com/kernelci/kernelci-core/pull/2249

Move the definition of models used in user-facing endpoints to kernelci-core so they can be used simultaneously by code in kernelci-api and helpers and other utility functions in kernelci-core.

These changes shouldn't break anything when merged, but pending PRs might need to be rebased.


docker/api/requirements.txt
```diff
@@ -12,3 +12,4 @@ python-jose[cryptography]==3.3.0
 pyyaml==5.3.1
 redis==5.0.1
 uvicorn[standard]==0.13.4
+kernelci
```
JenySadadia, line 15: I'd suggest also mentioning the package version here.

hardboprobot: Yes, I'm just not sure about which one to specify since I guess this depends on the deployment of the kernelci package after merging https://github.com/kernelci/kernelci-core/pull/2249

gctucker: We need to create the versions though ;)  That's related to what I mentioned on Slack last week, we need a `pyproject.toml` file and get this uploaded to PyPI.  In the meantime, it's possible to use a Git branch e.g.:
```
kernelci@git+https://github.com/kernelci/kernelci-core@release-branch
```

api/auth.py
```diff
@@ -12,7 +12,7 @@
     BearerTransport,
     JWTStrategy,
 )
-from .config import AuthSettings
+from .models import AuthSettings
```
JenySadadia, line 15: Why the settings classes have been moved to `models`?

hardboprobot: I think we discussed this in more general terms in a prior conversation, this moves all the pydantic model definitions to the same file. The idea was to have the models separated from the logic as much as possible.

JenySadadia: These are `Settings` classes and not usual object models. That's why we have it in the `.config` file. These are specific to API only. That's why we should keep it in the API.

hardboprobot: ACK and fixed

JenySadadia:
Could you please rebase?

test-docker-compose.yaml
```diff
@@ -63,3 +63,19 @@ services:
...
+
+  unit_test:
+    container_name: 'kernelci-api-unittests'
+    build:
+      context: 'docker/api'
+      args:
+        - REQUIREMENTS=${REQUIREMENTS:-requirements-dev.txt}
+    volumes:
+      - './api:/home/kernelci/api'
+      - './tests:/home/kernelci/tests'
```
JenySadadia, line 75: This should only need `unit_tests` directory.
```suggestion
      - './tests:/home/kernelci/tests/unit_tests'
```

.github/workflows/main.yml
```diff
@@ -39,24 +39,44 @@ jobs:
...
+          docker-compose -f test-docker-compose.yaml exec --no-tty unit_test pylint api.pubsub
+          docker-compose -f test-docker-compose.yaml exec --no-tty unit_test pylint api.user_manager
+          docker-compose -f test-docker-compose.yaml exec --no-tty unit_test pylint api.user_models
+          docker-compose -f test-docker-compose.yaml exec --no-tty unit_test pylint tests/unit_tests
 
       - name: Run pytest
         run: |
-          SECRET_KEY=b6451545d2e4635148d768f07877aade3ad8e7e160d52962badd7587a4b9a150 SMTP_HOST=smtp.gmail.com SMTP_PORT=465 EMAIL_SENDER=test@kernelci.org EMAIL_PASSWORD=random pytest -vs tests/unit_tests/
+          docker-compose -f test-docker-compose.yaml exec --no-tty unit_test pytest -vs tests/unit_tests/
+
```
JenySadadia, line 75: I think we can move the step to run unit tests to the workflow `test.yml` now.

.github/workflows/main.yml
```diff
@@ -39,24 +39,44 @@ jobs:
...
+        run: |
+          echo "SECRET_KEY=b6451545d2e4635148d768f07877aade3ad8e7e160d52962badd7587a4b9a150" > .env
+          echo "SMTP_HOST=smtp.gmail.com" >> .env
+          echo "SMTP_PORT=465" >> .env
+          echo "EMAIL_SENDER=test@kernelci.org" >> .env
+          echo "EMAIL_PASSWORD=random" >> .env
+
+      - name: Build docker images
+        run: docker-compose -f test-docker-compose.yaml build
+
```
JenySadadia, line 52: Moving unit tests to `test.yml` would also remove duplication of this steps.

test-docker-compose.yaml
hardboprobot: Fixed.

.github/workflows/main.yml
hardboprobot: But even if we move the unit tests, the container will still be needed in `main.yml` to run the pylint tests, so these steps are necessary to populate the `.env` file for the API.

.github/workflows/main.yml
hardboprobot: Done

.github/workflows/main.yml
JenySadadia: Then `pylint` step can also be moved to `test.yml` or maybe create a separate workflow file for linter tools?
Also, using the same container as unit test for linter doesn't look good to me.

test-docker-compose.yaml
```diff
@@ -63,3 +63,20 @@ services:
       - '.env'
     environment:
       - 'KEEP_ALIVE_PERIOD=0'
+
+  unit_test:
```
JenySadadia, line 67: Container name for e2e tests should also be changed from `test` to `e2e_test` to distinguish it from unit tests.

.github/workflows/main.yml
hardboprobot: Ok, but can we do that in a different PR after we merge this one? Refactoring the github CI tests is out of the scope of this PR.

test-docker-compose.yaml
hardboprobot: Same as the comment above, can we collect these changes and do them in a separate PR instead? We're mixing goals here, adding more unrelated things to this PR, losing focus and taking way too much time to review it.

JenySadadia: Yes, it can be a follow-up PR.

JenySadadia:
Oh, yes, it doesn't depend on the python package anymore.
Let's wait for the next staging run and we'll be good to go.

JenySadadia:
I don't see these changes on staging https://github.com/kernelci/kernelci-api/tree/staging.kernelci.org. Maybe a merge conflict with some other PR.

JenySadadia:
Do you have access to the staging VM? Otherwise, I can verify it.

JenySadadia:
Pipeline services are not running atm. Taking a look.

JenySadadia:
The issue is related to Pub/Sub. This PR seems working well.

JenySadadia (approved):

PR #434, opened 2023-12-15
Move `Subscription` model definition 

JenySadadia:
Move `Subscription` model definition to module containing all the generic API models i.e. `api/models.py` to keep coding style consistent.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -35,10 +35,11 @@ from .models import (
     Regression,
     UserGroup,
     PublishEvent,
+    Subscription,
     get_model_from_kind
 )
 from .paginator_models import PageModel
-from .pubsub import PubSub, Subscription
+from .pubsub import PubSub
 from .user_manager import get_user_manager, create_user_manager
 from .user_models import (
     User,
--- a/api/models.py
+++ b/api/models.py
@@ -383,3 +383,17 @@ class PublishEvent(BaseModel):
     attributes: Optional[Dict] = Field(
         description="Extra Cloudevents Extension Context Attributes"
     )
+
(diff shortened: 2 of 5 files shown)


hardboprobot (commented):
Although I can understand the rationale of moving all the model definitions to the same file, this clashes with the refactoring in https://github.com/kernelci/kernelci-api/pull/433, which differentiates models that are common to the API and clients (and are then moved to kernelci-core) from models that are meant to be used internally by the API code only, and I think the Subscription model is an example of the latter if I'm not mistaken. Is there (or could it be) any client code that needs to deal with the internals of subscriptions?

JenySadadia:
> Although I can understand the rationale of moving all the model definitions to the same file, this clashes with the refactoring in #433, which differentiates models that are common to the API and clients (and are then moved to kernelci-core) from models that are meant to be used internally by the API code only, and I think the Subscription model is an example of the latter if I'm not mistaken. Is there (or could it be) any client code that needs to deal with the internals of subscriptions?

Yes, this model is specific to API and isn't required to be accessed from the client side. We can merge this one and you can rebase your PR later on if that's okay for you.

JenySadadia:
> I guess another option would be to put all the server specific models back into a models.py file so that they're all in the same place. That is, merge all the models from user_models.py and pubsub.py into a new models.py file (the contents of the current models.py file are all moved to kernelci-core). What do you think?

Yes, after looking at #433,  it makes a lot of sense. I'll close this PR and we'll move API-specific models to `models.py` later on. Maybe that can be covered in #433 itself.

PR #435, opened 2023-12-15
api/main.py: Add permissive CORS policy

nuclearcat:
As we might experiment with 3rd party web applications we need CORS policy to be defined.
Probably it need to be more strict in the future.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -19,6 +19,7 @@ from fastapi import (
     Request,
     Form
 )
+from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse, PlainTextResponse
 from fastapi.security import OAuth2PasswordRequestForm
 from fastapi_pagination import add_pagination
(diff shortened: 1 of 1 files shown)


api/main.py
```diff
@@ -704,11 +705,18 @@ async def put_regression(regression_id: str, regression: Regression,
         ]
     )
 
+versioned_app.add_middleware(
+    CORSMiddleware,
+    allow_origins=["*"],
```
hardboprobot, line 710: Do we want this to be any origin? I guess eventually we'll have to define concrete origins explicitly here, right?
Also, according to https://fastapi.tiangolo.com/tutorial/cors/ , setting `['*']` here will allow only certain types of communication, excluding everything that involves credentials.

hardboprobot: I haven't tested it, will that limitation (restricted to non-credentials-based communication) break any of the ongoing work or is that something that will only have an effect when we have a proper font-end app?

nuclearcat: As i tested on staging - everything works fine, and looked to fastapi code - it should affect only web apps, where for example cookies are used, might also affect basic auth (which we dont use too).
But i will check more now, what exactly it does.

nuclearcat:
I will deprecate this PR, and probably will write this week small web interface hosted on API instance itself

PR #436, opened 2023-12-18
Remove `admin` user group

JenySadadia:
With `fastapi-users` integration, `User.is_superuser` flag has been introduced to identify admin users. 
This eliminates the requirement for creating any specific user group for admin users. Hence, discard the method to setup `admin` user group from the script.

### DIFF
--- a/tests/e2e_tests/test_pipeline.py
+++ b/tests/e2e_tests/test_pipeline.py
@@ -17,7 +17,7 @@ from .test_node_handler import create_node, get_node_by_id, update_node
     depends=[
         'e2e_tests/test_subscribe_handler.py::test_subscribe_node_channel'],
     scope='session')
-@pytest.mark.order(5)
+@pytest.mark.order(4)
 @pytest.mark.asyncio
 async def test_node_pipeline(test_async_client):
     """
--- a/api/admin.py
+++ b/api/admin.py
@@ -17,27 +17,19 @@ import getpass
 
 from .auth import Authentication
 from .db import Database
-from .models import UserGroup
 from .user_models import User
 
 
-async def setup_admin_group(db, admin_group):
-    group_obj = await db.find_one(UserGroup, name=admin_group)
-    if group_obj is None:
-        print(f"Creating {admin_group} group...")
-        group_obj = await db.create(UserGroup(name=admin_group))
-    return group_obj
-
-
-async def setup_admin_user(db, username, email, admin_group):
+async def setup_admin_user(db, username, email):
+    """Create an admin user"""
     user_obj = await db.find_one_by_attributes(User,
                                                {'username': username})
     if user_obj:
         print(f"User {username} already exists, aborting.")
         print(user_obj.json())
         return None
-    password = getpass.getpass(f"Password for user '{args.username}': ")
-    retyped = getpass.getpass(f"Retype password for user '{args.username}': ")
+    password = getpass.getpass(f"Password for user '{username}': ")
+    retyped = getpass.getpass(f"Retype password for user '{username}': ")
     if password != retyped:
         print("Sorry, passwords do not match, aborting.")
         return None
@@ -47,7 +39,6 @@ async def setup_admin_user(db, username, email, admin_group):
         username=username,
         hashed_password=hashed_password,
         email=email,
-        groups=[admin_group],
         is_superuser=1,
         is_verified=1,
     ))
@@ -56,8 +47,7 @@ async def setup_admin_user(db, username, email, admin_group):
 async def main(args):
     db = Database(args.mongo, args.database)
     await db.initialize_beanie()
-    group = await setup_admin_group(db, args.admin_group)
-    user = await setup_admin_user(db, args.username, args.email, group)
+    await setup_admin_user(db, args.username, args.email)
     return True
 
 
@@ -67,8 +57,6 @@ if __name__ == '__main__':
                         help="Mongo server connection string")
     parser.add_argument('--username', default='admin',
                         help="Admin username")
-    parser.add_argument('--admin-group', default='admin',
-                        help="Admin group name")
     parser.add_argument('--database', default='kernelci',
                         help="KernelCI database name")
     parser.add_argument('--email', required=True,
--- a/api/main.py
+++ b/api/main.py
@@ -220,7 +220,7 @@ async def update_me(request: Request, user: UserUpdate,
     """User update route
 
     Custom user update router handler will only allow users to update
-    its own profile. Adding itself to 'admin' group is not allowed.
+    its own profile.
     """
     if user.username and user.username != current_user.username:
         existing_user = await db.find_one(User, username=user.username)
@@ -232,10 +232,6 @@ async def update_me(request: Request, user: UserUpdate,
     groups = []
     if user.groups:
         for group_name in user.groups:
-            if group_name == "admin":
-                raise HTTPException(
-                    status_code=status.HTTP_400_BAD_REQUEST,
-                    detail="Unauthorized to add user to 'admin' group")
             group = await db.find_one(UserGroup, name=group_name)
             if not group:
                 raise HTTPException(
--- a/tests/e2e_tests/test_subscribe_handler.py
+++ b/tests/e2e_tests/test_subscribe_handler.py
@@ -12,7 +12,7 @@ import pytest
 @pytest.mark.dependency(
     depends=['e2e_tests/test_user_creation.py::test_create_regular_user'],
     scope='session')
-@pytest.mark.order(4)
+@pytest.mark.order(3)
 def test_subscribe_node_channel(test_client):
     """
     Test Case : Test KernelCI API '/subscribe' endpoint with 'node' channel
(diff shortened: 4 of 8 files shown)


tests/e2e_tests/test_pipeline.py
```diff
@@ -17,7 +17,7 @@
     depends=[
         'e2e_tests/test_subscribe_handler.py::test_subscribe_node_channel'],
     scope='session')
-@pytest.mark.order(5)
+@pytest.mark.order(4)
```
hardboprobot, line 20: Is there any other way to define the test order that doesn't involve going through all tests in all files manually adjusting this number when we remove or add tests in the future? This looks like a pain to maintain.

JenySadadia: Yes, it is cumbersome to maintain in the long run. But I don't think there is much we can do about it.
Maybe using `before` and `after` marker attributes will optimize this a bit. I'll try and test it.

JenySadadia: OK. So, I tested with [markers](https://pytest-order.readthedocs.io/en/stable/usage.html#order-relative-to-other-tests) used for reletive tests. However, it doesn't work as expected and multiple tests are being skipped because of the tests dependencies.
```
kernelci-api-e2e-tests | e2e_tests/test_user_creation.py::test_create_admin_user PASSED           [  5%]
kernelci-api-e2e-tests | e2e_tests/test_count_handler.py::test_count_nodes SKIPPED (test_coun...) [ 11%]
kernelci-api-e2e-tests | e2e_tests/test_count_handler.py::test_count_nodes_matching_attributes SKIPPED [ 17%]
kernelci-api-e2e-tests | e2e_tests/test_pipeline.py::test_node_pipeline SKIPPED (test_node_pi...) [ 23%]
kernelci-api-e2e-tests | e2e_tests/test_pubsub_handler.py::test_pubsub_handler SKIPPED (test_...) [ 29%]
kernelci-api-e2e-tests | e2e_tests/test_regression_handler.py::test_regression_handler SKIPPED    [ 35%]
kernelci-api-e2e-tests | e2e_tests/test_root_handler.py::test_root_endpoint PASSED                [ 41%]
kernelci-api-e2e-tests | e2e_tests/test_user_creation.py::test_create_regular_user PASSED         [ 47%]
kernelci-api-e2e-tests | e2e_tests/test_subscribe_handler.py::test_subscribe_node_channel PASSED  [ 52%]
kernelci-api-e2e-tests | e2e_tests/test_subscribe_handler.py::test_subscribe_test_channel PASSED  [ 58%]
kernelci-api-e2e-tests | e2e_tests/test_subscribe_handler.py::test_subscribe_user_group_channel PASSED [ 64%]
kernelci-api-e2e-tests | e2e_tests/test_user_creation.py::test_whoami PASSED                      [ 70%]
kernelci-api-e2e-tests | e2e_tests/test_user_creation.py::test_create_user_negative PASSED        [ 76%]
kernelci-api-e2e-tests | e2e_tests/test_user_group_handler.py::test_create_and_get_user_group PASSED [ 82%]
kernelci-api-e2e-tests | e2e_tests/test_password_handler.py::test_password_endpoint PASSED        [ 88%]
kernelci-api-e2e-tests | e2e_tests/test_unsubscribe_handler.py::test_unsubscribe_node_channel PASSED [ 94%]
kernelci-api-e2e-tests | e2e_tests/test_unsubscribe_handler.py::test_unsubscribe_test_channel PASSED [100%]
kernelci-api-e2e-tests | ======== 12 passed, 5 skipped, 13 warnings in 1.19s==================
```
So, I guess ordering with numbers is the only way here.

hardboprobot: But aren't test dependencies supposed to be defined with pytest-dependency? What's the purpose of pytest-order here then? Sorry if the answer is obvious, I'm still getting familiar with this.

JenySadadia: Yes, dependencies are defined by `pytest-dependency`. In some cases, we need to execute tests in a specific sequence. For instance, we need to create users before sending authenticated requests. So, instead of adding the dependency marker to all the authenticated tests, we can add order marker to make sure the test creating user executes before any another tests.

hardboprobot: I understand, what I wanted to know is what's the advantage of using explicit test ordering here vs using dependencies and fixtures. It seems to me like it's overused.

The way I see it, the semantically correct way to use these could be:

- fixtures: what needs to be run *before* the test starts, as a preparation. Or *after*, at the end, as a teardown.
- dependencies: when a test should run or not depending on the status of a previous test or tests.
- order: to define a specific ordering in tests with the same constraints for other reason, but I can't think of a reasonable use case that can't be implemented with fixtures and dependencies.

In these tests, I'm seeing both dependency **and** order markers to create a:

`test_create_user_groups` -> `test_create_admin_user` -> `test_create_regular_user` -> `test_subscribe_node_channel`, `test_subscribe_test_channel` and `test_subscribe_user_group_channel` -> etc

What's the point of the order markers then? Couldn't the same thing be done with the dependency markers only?

JenySadadia: The point of using order markers is, that if you just use dependencies, it does not guarantee that a test will only execute after its dependency. Tests will execute in any order and see if the dependency is `PASSED` otherwise the test will be `SKIPPED`. It does not make sure that dependencies will execute before the dependent. That's when `order` comes into the picture.

For example, when I removed `order` from all the tests except user and user group creation, this is what I got:
```
kernelci-api-e2e-tests | 
kernelci-api-e2e-tests | e2e_tests/test_user_group_handler.py::test_create_user_groups PASSED     [  5%]
kernelci-api-e2e-tests | e2e_tests/test_user_creation.py::test_create_admin_user PASSED           [ 10%]
kernelci-api-e2e-tests | e2e_tests/test_user_creation.py::test_create_regular_user PASSED         [ 15%]
kernelci-api-e2e-tests | e2e_tests/test_count_handler.py::test_count_nodes SKIPPED (test_coun...) [ 21%]
kernelci-api-e2e-tests | e2e_tests/test_count_handler.py::test_count_nodes_matching_attributes SKIPPED [ 26%]
kernelci-api-e2e-tests | e2e_tests/test_password_handler.py::test_password_endpoint PASSED        [ 31%]
kernelci-api-e2e-tests | e2e_tests/test_pipeline.py::test_node_pipeline SKIPPED (test_node_pi...) [ 36%]
kernelci-api-e2e-tests | e2e_tests/test_pubsub_handler.py::test_pubsub_handler SKIPPED (test_...) [ 42%]
kernelci-api-e2e-tests | e2e_tests/test_regression_handler.py::test_regression_handler SKIPPED    [ 47%]
kernelci-api-e2e-tests | e2e_tests/test_root_handler.py::test_root_endpoint PASSED                [ 52%]
kernelci-api-e2e-tests | e2e_tests/test_subscribe_handler.py::test_subscribe_node_channel PASSED  [ 57%]
kernelci-api-e2e-tests | e2e_tests/test_subscribe_handler.py::test_subscribe_test_channel PASSED  [ 63%]
kernelci-api-e2e-tests | e2e_tests/test_subscribe_handler.py::test_subscribe_user_group_channel PASSED [ 68%]
kernelci-api-e2e-tests | e2e_tests/test_unsubscribe_handler.py::test_unsubscribe_node_channel PASSED [ 73%]
kernelci-api-e2e-tests | e2e_tests/test_unsubscribe_handler.py::test_unsubscribe_test_channel PASSED [ 78%]
kernelci-api-e2e-tests | e2e_tests/test_user_creation.py::test_whoami PASSED                      [ 84%]
kernelci-api-e2e-tests | e2e_tests/test_user_creation.py::test_create_user_negative PASSED        [ 89%]
kernelci-api-e2e-tests | e2e_tests/test_user_group_handler.py::test_get_user_group PASSED         [ 94%]
kernelci-api-e2e-tests | e2e_tests/test_user_group_handler.py::test_create_and_get_user_group PASSED [100%]
kernelci-api-e2e-tests | e2e_tests/test_user_group_handler.py::test_create_and_get_user_group ERROR [100%]
kernelci-api-e2e-tests | 
kernelci-api-e2e-tests | ============= 14 passed, 5 skipped, 13 warnings, 1 error in 1.32s ====
```

Still, if you feel that you can come up with a better solution, that'd be welcome.

hardboprobot: Well crap. I read more documentation about pytest-dependency and it looks like it doesn't enforce any true ordering at all. I wonder what's the point of implementing test dependencies if they don't translate to an execution order.

Anyway, there's the [`--order-dependencies`](https://pytest-dev.github.io/pytest-order/stable/configuration.html#order-dependencies) option that combines the order and dependency plugins and does exactly what we want: test ordering based on dependency markers. But it doesn't work in our case:

```
kernelci-api-e2e-tests | WARNING: cannot execute test relative to others: e2e_tests/test_pipeline.py::test_node_pipeline e2e_tests/test_pipeline.py::test_node_pipeline e2e_tests/test_pipeline.py::test_node_pipeline e2e_tests/test_subscribe_handler.py::test_subscribe_node_channel e2e_tests/test_subscribe_handler.py::test_subscribe_test_channel e2e_tests/test_user_creation.py::test_create_regular_user e2e_tests/test_user_creation.py::test_create_regular_user - ignoring the marker.
```

Maybe it doesn't play well with async functions, who knows. I guess we'll have to keep hacking the order manually.

hardboprobot (approved):

PR #438, opened 2023-12-20
migrations: fix existing migrations scripts

hardboprobot:
Remove the "set_timeout" migration script, as it won't be needed. Make some fixes to the "user" script:

  - Keep "admin" users, converting from the old format (pertenence to the "admin" group) to the new one ("is_superuser" field)
  - Adjust the "is_verified" and "is_active" fields accordingly
  - Make the script more robust: run sanity checks on the data to apply the migrations only when needed
  - On upgrade, remove the "admin" group
  - Implement a symmetric downgrade function


migrations/20231102101356_user.py
```diff
@@ -23,14 +57,64 @@ def upgrade(db: "pymongo.database.Database"):
                 "_id": user['_id'],
                 "email": user['profile']['email'],
                 "hashed_password": user['profile']['hashed_password'],
-                "is_active": True,
-                "is_superuser": False,
-                "is_verified": False,
+                "is_active": user['active'],
+                "is_superuser": is_superuser,
+                "is_verified": is_superuser,
```
JenySadadia, line 62: I think this should only be enabled for the initial admin user. Not all the admin users should be automatically verified.

migrations/20231102101356_user.py
```diff
@@ -23,14 +57,64 @@ def upgrade(db: "pymongo.database.Database"):
...
                 "username": user['profile']['username'],
                 "groups": user['profile']['groups']
             },
         )
+    # Sanity check: check if there are any old-format users in the
+    # "admin" group. Remove the group if there aren't any
+    remaining_admins = db.user.count(
```
JenySadadia, line 69: I don't think there will be any remaining admin users that are still in `admin` group after migration.
Do we have any specific use-case for this?

migrations/20231102101356_user.py
hardboprobot: The problem with that is how to tell who the initial admin user is. AFAICT this information isn't in the DB, so we can choose between having all admin users verified or none of them. The first option has the risk of automatically considering as verified admin users that weren't verified before the migration, the second one has the drawback of having to manually go through the admin users and verifying only the ones that were verified before the migration. But then again, that information wasn't in the DB before either, was it?

migrations/20231102101356_user.py
```diff
@@ -4,17 +4,51 @@
...
+
+    api.main: update/add user management routes
+
 """
 
 name = '20231102101356_user'
-dependencies = ['20221014061654_set_timeout']
+dependencies = []
+
+def user_upgrade_needed(user):
```
JenySadadia, line 17: Even though there is no restriction on function names returning boolean values, following the `is_*` pattern would help developers easily predict its return value.
```suggestion
def is_upgrade_needed(user):
```
I removed `user` from the name as this migration file is specific to user migrations.

migrations/20231102101356_user.py
hardboprobot: No, but this is a sanity check, it's meant to test if the migration ran as expected. If for any reason a user slips through the cracks and isn't migrated correctly (example corner case: old, manually-forged user object that didn't make the migration filter), this will give a warning.

migrations/20231102101356_user.py
hardboprobot: I actually find that kind of misleading, since in python the `is` prefix is normally used for predicates that describe a *property* or a characteristic (`isistance`, `issubclass`). The name I chose sounds quite clear to me for a conditional action.

JenySadadia: Okay. That was just a suggestion :+1:

migrations/20231102101356_user.py
JenySadadia: In the former user schema, we didn't have any mechanism to verify the user. So, basically, with the migrations, we should make all the users unverified. As a part of adopting the process of a new schema, the user should undertake verification even if it's the initial admin user.  So, we should use `"is_verified":  False` in the migration.
What do you think?

hardboprobot: I guess that'll have to be the way, I'll make the fix. Thanks!

PR #440, opened 2023-12-27
env.sample: add email config variables

JenySadadia:
API sends emails for some user related operations such as verifying user, and password reset.
Add email configuration variable to sample file for environment variables.

### DIFF
--- a/env.sample
+++ b/env.sample
@@ -1,3 +1,7 @@
 SECRET_KEY=
 #algorithm=
 #access_token_expire_minutes=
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
Thanks, thats very useful, as i was puzzled a bit during production update about this environment variables. We have minor inconsistency, i think in pipeline we have EMAIL_USER, while same thing here is EMAIL_SENDER, maybe we need to make it same name in near future (but we will have to migrate production/staging as usual).

PR #442, opened 2023-12-28
kube/minikube: add email configuration variables

JenySadadia:
Define `ConfigMap` variables for email-related configurations. Update `api` deployment file to use newly added configurations.


nuclearcat (approved):
LGTM

PR #443, opened 2023-12-28
api.email_sender: catch SMTP related exceptions

JenySadadia:
Add `try-catch` block to `EmailSender._send_email` method to catch exception while sending emails.


nuclearcat (approved):
Thanks! Hopefully on next issue it will help us to catch errors, meanwhile all works fine on staging

PR #447, opened 2024-01-02
api.pubsub: shutdown pubsub connection

JenySadadia:
Shut down `PubSub` connection in `unsubscribe` method for proper cleanup.


hardboprobot (approved):
Have you noticed any behavioral change after this? If everything still works the same, then it LGTM

JenySadadia:
Staging looks OK.

Issue #426, opened 2023-12-05
Enable pylint on all the files under `api/`


JenySadadia:
At the moment, `pylint` is running on most of the source files under `api/` directory in the Github workflow.
Enable it on all the remaining files to make sure that developers and maintainers follow expected coding standards.

- [ ] Check and fix pylint errors for all the source files under `api/` directory
- [ ] Update job to run pylint in the workflow file `main.yml`
- [ ] Test with Github Actions

gctucker:
Yes I think `api.client` can be deleted, I doubt it's still working anyway.  We should just provide some sample ways to use the raw API in the documentation as it shouldn't depend on the `kci` command line tool - the dependency goes the other way.  In fact the generated documentation probably already does that so maybe we can just delete that file, then we'll be able to run pylint on all the API files.

JenySadadia:
Okay. Thanks.

PR #449, opened 2024-01-04
api: remove `client` module

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/426

`api.client` was implemented as a command-line utility for sending and receiving Pub/Sub events. `kci` CLI tool has been implemented to use API in `kernelci-core`. The tool can be used to send all kinds of requests to API. Hence, `api.client`
can be dropped and `kci` can be used instead.


JenySadadia:
Added a commit to update the documentation.

nuclearcat (approved):
Looks good: https://static.staging.kernelci.org/docs/api/api-details/

PR #453, opened 2024-01-10
Assorted e2e test updates

hardboprobot:
- Fix pipeline test (sync it to Node model changes)
- The "test" service of test-docker-compose.yaml no longer launches the tests on startup. Update the run_e2e_tests.sh script to start the service and run the tests.
- Fix inconsistencies in pytest dependencies

### DIFF
--- a/test-docker-compose.yaml
+++ b/test-docker-compose.yaml
@@ -56,6 +56,7 @@ services:
     volumes:
       - './api:/home/kernelci/api'
       - './tests:/home/kernelci/tests'
+      - './tests/e2e_tests:/home/kernelci/e2e_tests'
     depends_on:
       - api
     env_file:
--- a/run_e2e_tests.sh
+++ b/run_e2e_tests.sh
@@ -9,5 +9,5 @@
 
 docker-compose -f test-docker-compose.yaml build
 docker-compose -f test-docker-compose.yaml up -d api db redis storage ssh test
-docker-compose -f test-docker-compose.yaml exec test pytest -v tests/e2e_tests
+docker-compose -f test-docker-compose.yaml exec test pytest -v e2e_tests
 docker-compose -f test-docker-compose.yaml down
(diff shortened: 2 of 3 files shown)


JenySadadia (changes requested):
Failing locally:
```
WARNING: Cannot resolve the dependency marker 'e2e_tests/test_user_creation.py::test_create_regular_user' - ignoring it.
WARNING: Cannot resolve the dependency marker 'e2e_tests/test_subscribe_handler.py::test_subscribe_node_channel' - ignoring it.
WARNING: Cannot resolve the dependency marker 'e2e_tests/test_subscribe_handler.py::test_subscribe_test_channel' - ignoring it.

```

JenySadadia:
One solution is to mount `tests/e2e_tests` directly in `test-docker-compose.yaml`.
```
- './tests/e2e_tests:/home/kernelci/e2e_tests'
```
Another solution is to update all the dependecy markers to use `test/e2e_tests/*`  instead of `e2e_tests/*`
```
@pytest.mark.dependency(
    depends=['tests/e2e_tests/test_user_creation.py::test_create_regular_user'],
    scope='session')
```
I think the first one looks easier than updating all the tests.

test-docker-compose.yaml
```diff
@@ -56,6 +56,7 @@ services:
     volumes:
       - './api:/home/kernelci/api'
       - './tests:/home/kernelci/tests'
```
JenySadadia, line 58: Thanks for the fix.
We can drop `tests` mount now.
```suggestion
```

hardboprobot: No, we can't because this container is also used for unit tests (https://github.com/kernelci/kernelci-api/blob/1e7a0be392aa19daaaa085752aa5dbb30b6494e7/.github/workflows/test.yml#L47). We could do two different volume binds, one for each type of test, and also change the github yml files, but that has nothing to do with this PR

JenySadadia: Yes, that can be a follow-up PR.

JenySadadia (approved):
Tested OK.

PR #454, opened 2024-01-22
migrations: fix regression node migrations

hardboprobot:
Fixes a bug in the migration of regressions (https://github.com/kernelci/kernelci-api/commit/d9176e8724fe25dcaeb56b2b53f983150caed97a)


nuclearcat (approved):
I guess this one wont be easy to test on staging, so approving as is.
@JenySadadia can you take a look as well?

JenySadadia:
Yes, I'll test it locally.

JenySadadia (approved):
Tested OK.

PR #455, opened 2024-01-22
docker-compose.yaml: Add git repo core_rev

nuclearcat:
To test on staging we need to able to set different branch, using environment variables.

Depends on https://github.com/kernelci/kernelci-core/pull/2286

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -12,6 +12,7 @@ services:
       context: 'docker/api'
       args:
         - REQUIREMENTS=${REQUIREMENTS:-requirements.txt}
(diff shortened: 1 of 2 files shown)


JenySadadia (approved):
Tested OK locally.

nuclearcat:
Thanks Jeny, we also tested yesterday issue with model restrictions, now staging works properly.

JenySadadia:
> Thanks Jeny, we also tested yesterday issue with model restrictions, now staging works properly.

But it should not have been merged without merging https://github.com/kernelci/kernelci-core/pull/2286.

PR #456, opened 2024-01-23
README: fix links

a-wai:
The "Getting Started" page has been replaced with "Local Instance", so update all the corresponding links. While at it, add minor updates:
* properly terminate a link entry
* drop the full URL from the internal "Authentication" link so it always point to the current document (avoid potential future breakages)


JenySadadia (approved):
Tested OK on staging https://github.com/kernelci/kernelci-api/tree/staging.kernelci.org.

PR #457, opened 2024-01-23
doc/api-details: fix reference to `Getting started` page

JenySadadia:
As the `Getting started` page has been renamed to `Local instance`, update the link to the section for creating environment variables.

Fixes: eefbece ("doc: rename getting-started to local-instance")


a-wai (approved):
Ah, I missed that one, good catch, thanks!

PR #458, opened 2024-01-25
viewer: Add HTML helper to view nodes

nuclearcat:
This is PoC of API viewer for development purposes.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -714,6 +714,26 @@ async def stats(user: User = Depends(get_current_superuser)):
     return await pubsub.subscription_stats()
 
 
+@app.get('/viewer')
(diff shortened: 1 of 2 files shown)


nuclearcat:
I think we might need to merge it, unless there is objections, cause already i have new experimental updates planned for next stage

a-wai:
Agreed, it's already extremely helpful!

If I may request a feature for the next iteration: adding an option to download the node as raw json would be nice :)

a-wai (approved):
Approving based on the usefullness, my JS skills are basically null

nuclearcat:
Merging, as i need to add few more features to it and create new PR.

PR #459, opened 2024-01-29
doc/early-access: use correct argument for email verification

pawiecz:
Tool "kci" expects email as an argument (not the user name):

```
$ ./kci user verify --help

Usage: kci user verify [OPTIONS] EMAIL

(snip)
```

### DIFF
--- a/doc/early-access.md
+++ b/doc/early-access.md
@@ -89,7 +89,7 @@ Retype new password:
 sent to your email:
 
 ```sh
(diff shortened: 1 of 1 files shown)


JenySadadia (approved):
Thanks for the fix. LGTM.

JenySadadia:
Tested OK on staging: https://static.staging.kernelci.org/docs/api/early-access/

PR #460, opened 2024-01-31
api.db: add support for regexes in mongodb queries

hardboprobot:
Introduce a new API query operator (re) to match query parameters using mongodb-compatible regular expressions.

Mongodb has built-in regex capabilities for queries, so we get those for free if we simply add support for an additional query operator.

### DIFF
--- a/api/db.py
+++ b/api/db.py
@@ -34,6 +34,7 @@ class Database:
         'gt': '$gt',
         'gte': '$gte',
         'ne': '$ne',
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
Tested on staging, works!

JenySadadia:
Thanks for the addition @hardboprobot . This should be documented here https://kernelci.org/docs/api/api-details/#getting-nodes-back

PR #461, opened 2024-02-01
doc/api-details: update Node examples, model format and operators

hardboprobot:
Sync the docs to the current model definitions and mention the `__re` operator, including an example.

### DIFF
--- a/doc/api-details.md
+++ b/doc/api-details.md
@@ -82,13 +82,18 @@ Another way of creating users is to use `kci user add` tool from kernelci-core.
 
 ## Nodes
 
-As a proof-of-concept, an object model called `Node` is defined in this API.
-It's possible to create new objects and retrieve them via the API.
+`Node` objects form the basis of the API models to represent tests runs,
+kernel builds, regressions and other test-related entities and their
+relationships. See [the model definitions in
+kernelci-core](https://github.com/kernelci/kernelci-core/blob/main/kernelci/api/models.py)
+for details on the `Node` model and its subtypes. It's possible to
+create new objects and retrieve them via the API.
 
 ### Create a Node
 
-To create an object of `Node` model, a `POST` request should be made along with
-the Node attributes. This requires an authentication token:
+To create a `Node` or a `Node` subtype object, for instance, a
+`Checkout` node, a `POST` request should be made along with the Node
+attributes. This requires an authentication token:
 
 ```
 $ curl -X 'POST' \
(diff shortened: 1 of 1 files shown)


doc/api-details.md
```diff
@@ -97,14 +102,48 @@ $ curl -X 'POST' \
...
+    "data": {
+      "kernel_revision": {
+        "tree": "mainline",
+        "url": "https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git",
+        "branch": "master",
+        "commit": "2a987e65025e2b79c6d453b78cb5985ac6e5eb26",
+        "describe": "v5.16-rc4-31-g2a987e65025e"
+      }
+    }
+  }' | jq
```
JenySadadia, line 117: I am wondering if we should have one-liner instruction for installing `jq` package as all the sample commands are using it.

JenySadadia (changes requested):
As you are already taking care of renaming `revision` field to `kernel_revision` in the PR, can you also add a fix for the response https://github.com/kernelci/kernelci-api/blob/main/doc/api-details.md?plain=1#L152 ?

JenySadadia:
> @JenySadadia gentle ping for this

Sorry for the delay. Just added a few minor comments otherwise LGTM.

doc/api-details.md
```diff
@@ -114,36 +153,220 @@ Reading Node doesn't require authentication, so plain URLs can be used.
...
+      "timeout": "2024-01-14T10:11:32.922000",
+      "holdoff": "2024-01-14T09:40:19.284000",
+      "owner": "admin",
+      "user_groups": []
+    }
+    ...
 ```
 
 Attributes along with comparison operators are also supported for the
 `/nodes` endpoint. The attribute name and operator should be separated by `__` i.e. `attribute__operator`. Supported operators are `lt`(less than), `gt`(greater than), `lte`(less than or equal to), and `gte`(greater than or equal to).
```
JenySadadia, line 353: Please add `re` here:
```suggestion
`/nodes` endpoint. The attribute name and operator should be separated by `__` i.e. `attribute__operator`. Supported operators are `lt`(less than), `gt`(greater than), `lte`(less than or equal to), `gte`(greater than or equal to), and `re`(regular expression).
```

doc/api-details.md
hardboprobot: For which distro / environment? These examples don't assume anything in particular.

I think `jq` is "external" to kernelci and not a requirement for anything, most of these examples aren't meant to be copied and pasted anyway, `jq` is there only to make it easier to see the result. If anyone wants to use it they know they'll need to have it installed, just like `curl`.

doc/api-details.md
hardboprobot: Done, thanks

doc/api-details.md
JenySadadia: Yes, I understand that it's optional.
Usually, developers/users follow the instruction guide in the ramp-up phase, and copy-paste commands to try them. In that case, it will generate an error like:
```
Command 'jq' not found, but can be installed with:
...
curl: (23) Failure writing output to destination
```
And then the user will need to check the correct command for installation.
I suggest if we want to keep it in the sample commands, it doesn't hurt to just add a line like `The sample commands below use `jq` package for output formatting and should be installed first to use it.` And we can put a link to its installation guide.

JenySadadia: @hardboprobot This conversation was still unresolved. Anyway it was just a suggestion :+1:

hardboprobot: Oops sorry, I missed this

PR #468, opened 2024-02-07
docker-compose.yaml: Update API to use docker image

nuclearcat:
Instead of doing local build - use docker image from registry.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -8,20 +8,17 @@ services:
 
   api:
     container_name: 'kernelci-api'
-    build:
-      context: 'docker/api'
-      args:
-        - REQUIREMENTS=${REQUIREMENTS:-requirements.txt}
-        - core_rev=${CORE_REV:-main}
-    volumes:
-      - './api:/home/kernelci/api'
-      - './tests:/home/kernelci/tests'
-      - './migrations:/home/kernelci/migrations'
-      - './templates:/home/kernelci/templates'
+    image: ${KERNELCI_API_IMAGE:-kernelci/staging-api}:${KERNELCI_API_TAG:-latest}
     ports:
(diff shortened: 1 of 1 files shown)


pawiecz:
I believe having a docker-compose doing a local build could still be useful for further development.

Do you think the main `docker-compose.yml` would be more frequently used for sample deployments and the one with adjustments for local builds should be renamed to e.g. `docker-compose-dev.yml` or would the development purposes be the main use case and there should be `docker-compose-production.yml` with image pulled from registry?

nuclearcat:
current -dev is broken, this is what caused me to use docker image. But definitely i can rename old one to -dev and we can fix it for local development.

pawiecz:
> current -dev is broken, this is what caused me to use docker image.

I'm a bit confused on what exactly was broken but it might not be relevant anymore.

> But definitely i can rename old one to -dev and we can fix it for local development.

Sure, let's keep them both (having the one with images from the registry as the default one). Later we could align them to keep only necessary overrides in the `-dev` (and call them [like on staging](https://github.com/kernelci/kernelci-deploy/blob/065097c9faa2227bbbfe99ccde891fcf1c059352/staging.kernelci.org#L358)).

pawiecz:
Unit test were out-of-sync with current models in `kernelci-core`, hence failure. It should be all fine once this branch is rebased on current `main`.

JenySadadia (approved):
Tested OK locally.

JenySadadia:
@nuclearcat Could you please rebase the PR? Hopefully, that would fix the unit tests.

nuclearcat:
Done, i think it can be merged?

JenySadadia:
> Done, i think it can be merged?

Yes, thanks.

PR #475, opened 2024-02-08
tests/unit_tests: sync unit tests after model changes (error_code field)

pawiecz:
[Recent changes](https://github.com/kernelci/kernelci-core/pull/2348) to the model caused unit tests to get out-of-sync with its new state.

This results in [false positive CI workflow failures](https://github.com/kernelci/kernelci-api/actions/runs/7827783619/job/21356327252?pr=473) which were not caused by the code introduced in the PR.

Fixes: #476


JenySadadia:
It is considered to be a good practice to have a commit description for every commit you make. I'd appreciate it if you could add it with `Fixes` tag.

JenySadadia (approved):
LGTM.

PR #481, opened 2024-02-19
pyproject.toml: add package directories such as templates and scripts

JenySadadia:
Only the `api` directory is included with the python package atm. Add directories such as templates and scripts to run API container properly to the package.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,4 +51,8 @@ Documentation = "https://kernelci.org/docs"
 Repository = "https://github.com/kernelci/kernelci-api"
 
 [tool.setuptools]
-packages = ["api"]
+packages = ["api", "scripts", "templates"]
+
+[tool.setuptools.package-data]
+scripts = ["*"]
+templates = ["*.jinja2", "*.html"]


pyproject.toml
```diff
@@ -51,4 +51,8 @@ Documentation = "https://kernelci.org/docs"
 Repository = "https://github.com/kernelci/kernelci-api"
 
 [tool.setuptools]
-packages = ["api"]
+packages = ["api", "scripts", "templates"]
+
+[tool.setuptools.package-data]
+scripts = ["*"]
+templates = ["*.jinja2"]
```
pawiecz, line 58:
```suggestion
templates = ["*.jinja2", "*.html"]
```

See https://github.com/kernelci/kernelci-api/pull/458

pawiecz (changes requested):

pyproject.toml
JenySadadia: Resolved. Thanks.

JenySadadia:
Instead of using `[tool.setuptools.package-data]` I could have added `include_package_data = true` flag and that would include all the non-py files of the directory but then we would lose control over what is being shared in the package.

pawiecz (approved):
Verified with imported `templates/viewer.html` from #458:

```sh
$ pip install .
$ pip show -f kernelci-api
Files:
(snip)
  scripts/setup_admin_user
  templates/email-verification-successful.jinja2
  templates/email-verification.jinja2
  templates/reset-password-successful.jinja2
  templates/reset-password.jinja2
  templates/viewer.html
```

nuclearcat:
It works, thanks!

PR #484, opened 2024-02-21
viewer: add ability to download raw node and open LAVA job page

a-wai:
Those features can be useful when debugging, as it will allow to easily:
* download the raw node as a local JSON file for testing/debugging purpose
* open the job page of a finished LAVA job to analyse the results and/or failures


nuclearcat:
Thanks, merging, tested on staging

nuclearcat:
Tested on staging

PR #485, opened 2024-02-22
Move templates to `api` directory

JenySadadia:
In order to access templates while running API from python package, move `templates` to `api` directory.
Update `pyproject.toml` to include `api/templates` directory to the python package.
Update `api.user_manager` to use `PackageLoader` to load templates from the python package.


JenySadadia:
> In order to access templates while running API from python package, move `templates` to `api` directory. Update `pyproject.toml` to include `api/templates` directory to the python package. Update `api.user_manager` to use `PackageLoader` to load templates from the python package.

PR #488, opened 2024-02-23
test-docker-compose: use `kernelci/staging-api` image

JenySadadia:
Use docker image for building api container for running tests.


nuclearcat:
Should be safe to merge, thanks!

JenySadadia:
Opps, I was still figuring out how it worked without `dev` packages in the `kernelci/staging-api` image.

PR #491, opened 2024-02-26
Add build argument for `e2e` tests

JenySadadia:
Created on top of https://github.com/kernelci/kernelci-api/pull/490

scripts/run_e2e_tests: add build argument
    
Add `no-cache` build argument while building images for tests.


a-wai (approved):
Maybe close #490 in favor of this one? As it touches the same area and just adds one more commit.

PR #496, opened 2024-02-28
Remove `Dockerfile` for `api` container

JenySadadia:
As `kernelci/staging-api` docker image has been used to build `api` and `tests` docker containers, remove the docker recipe from `docker/api` directory.


JenySadadia:
> Makes sense, however it seems I can't run my local instance since [9a569d6](https://github.com/kernelci/kernelci-api/commit/9a569d6f05d6619690417454621775e115d2a591) got merged, so I still need that file until I figure out how to fix things.

One way is to build a docker image with different `api_rev` and `core_rev` and use that image in your local setup.

a-wai:
> One way is to build a docker image with different `api_rev` and `core_rev` and use that image in your local setup.

Thanks for the hint! My idea was rather to use volumes so I can use the published container but still get my local versions of api and core. I've now found the proper settings, so feel free to go ahead and merge this.

a-wai (approved):

PR #499, opened 2024-03-06
api.db: translate query params received with a dict form

JenySadadia:
Update `Database` translation for operators and integer fields to use translated parameters in a dictionary form. For example: `{field: {operato1: value}}` will be converted to `{field: {operator1_mapping: value}}`.


hardboprobot (approved):
Thanks for fixing this. If I may add a suggestion, I think this translation scheme could be briefly documented somewhere in these functions, at least the data structures used and what results these functions produce. It doesn't need to be done in this PR, though.

JenySadadia:
Tested on staging.
Seems like below doesn't work:
```
 ./kci node find kind=regression state=done created__gt=2024-03-04 created__lt=2024-03-06 group__re=tast
```
But this works:
```
./kci node find kind=regression state=done created__lt=2024-03-06 created__gt=2024-03-04 group__re=tast
```

JenySadadia:
> Thanks for fixing this. If I may add a suggestion, I think this translation scheme could be briefly documented somewhere in these functions, at least the data structures used and what results these functions produce. It doesn't need to be done in this PR, though.

Yes, I'll add details in the docstrings in a follow-up PR.

JenySadadia:
Yes, I tested again and both commands are working fine. 
That's weird because previously it was giving nodes with the incorrect range.
Anyway, it seems working OK on staging now.

PR #501, opened 2024-03-07
Add `docker-compose` file for developer mode

JenySadadia:
Enable developer to use local volume binds by adding `dev-docker-compose.yaml` file. Also, add `Dockerfile` to provide the ability to build `api` container locally. This file is intended for developer use only.

### DIFF
--- a/dev-docker-compose.yaml
+++ b/dev-docker-compose.yaml
@@ -1,0 +1,61 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2024 Collabora Limited
+# Author: Jeny Sadadia <jeny.sadadia@collabora.com>
+
+version: '3'
+services:
+
+  api:
+    container_name: 'kernelci-api'
+    build:
+      context: 'docker/api'
+      args:
+        - REQUIREMENTS=${REQUIREMENTS:-requirements.txt}
+        - core_rev=${CORE_REV:-main}
+    volumes:
+      - './api:/home/kernelci/api'
+      - './tests:/home/kernelci/tests'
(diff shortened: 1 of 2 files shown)


JenySadadia:
> I still fail to see how different this would be from using the published container and mounting the local repo(s) as docker volume(s).

The whole point here is to enable development by using a local API setup. This isn't just about using local volume. What about when we need to add a Python package as a requirement to build some functionality? We can't use the published container for that. We should enable API development in a fully local mode as this is what we generally require while building something new or debugging in the API source code.
Rather than editing `docker-compose` file manually every time, we can directly provide a separate compose file for dev mode.

> My main concerns here are:
> - we end up with a duplicate-ish Dockerfile we must remember to update when we change the main container

That is required when we need to build and test `api` container with different requirement packages. That's true that we need to remember to update it but the docker recipe will have to change once in a while when `api.jinja2` template is updated in the `core`.

> - this only solves one problem (using the local API copy), but kernelci-core is still fetched from a git repo (while it is useful to use a local copy during development)

We can add that volume as well but not every use case will need to use local copy of `kernelci` module. Hence, we can consider adding a comment on binding volume for `core`.

JenySadadia:
> Personally I work with a docker-compose.override.yml file which sets the volumes I need, and would rather have a similar file in the repo (named docker-compose.override.sample.yml for example, so it doesn't get used by default).

By default `docker-compose` command tries to find `docker-compose.yaml` or `compose.yaml`. So, I guess adding the new yaml file won't create problem. Bdw, yes we can add `dev-*.sample.yaml` too.

a-wai (approved):

nuclearcat:
Thanks, sounds great feature!

PR #503, opened 2024-03-19
api.main: handle new Node path format in node submission

r-c-n:
When using a /-separated path string for Node paths instead of a list of strings, we can encode whether a Node is a leaf node or an intermediate node in a tree by ending the path string with a trailing slash character ('/') for intermediate nodes.

Implement the appropriate checks and logic in the node submission endpoints. For single nodes, adapt the path string of the parent node. For node trees, normalize the paths for the whole tree, adding trailing slashes when needed for all intermediate nodes.

**DEPENDS ON** https://github.com/kernelci/kernelci-core/pull/2466, the unit tests will fail until that PR is merged

**NOTE** These changes require running the DB migration script included.


JenySadadia:
LGTM. I'll mark it approved once we successfully test this on staging with all the related PRs.
Is this tested OK locally? @hardboprobot

Issue #479, opened 2024-02-15
Upgrade API to use Python 3.11


JenySadadia:
Atm the API uses `Python 3.10` and doesn't work when we try to build it with `Python 3.11`:
```
Created wheel for rx: filename=Rx-1.6.3-py2.py3-none-any.whl size=182331 sha256=f211ea5e2d444bf23b3b3026f84c9c7f08b192c140b3db146ce8f99a8dfb1251
  Stored in directory: /home/kernelci/.cache/pip/wheels/56/33/54/2b494f0f170cb849268dc65c53e63ade6c231cb2d6d914de6e
Successfully built MarkupSafe pymongo-migrate pyyaml python-multipart websockets ujson promise rx
Failed to build httptools
ERROR: Could not build wheels for httptools, which is required to install pyproject.toml-based projects
The command '/bin/sh -c pip install --requirement ${REQUIREMENTS}' returned a non-zero code: 1
ERROR: Service 'api' failed to build : Build failed
```

Eventually, we need to migrate it to Python 3.11 as `kernelci` package is already using it and API uses model definitions from the package. This is causing python package compatibility issues between both packages:
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
kernelci 1.1 requires click==8.1.3, but you have click 7.1.2 which is incompatible.
kernelci 1.1 requires jinja2==3.1.3, but you have jinja2 2.11.3 which is incompatible.
kernelci 1.1 requires pyyaml==6.0, but you have pyyaml 5.3.1 which is incompatible.
kubernetes 26.1.0 requires pyyaml>=5.4.1, but you have pyyaml 5.3.1 which is incompatible.
```

PR #504, opened 2024-03-19
T40464 Upgrade API to python 3.11

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/479

Upgrade Python version to 3.11 for API. Update docker recipe that is being used for Developer mode to use `Python 3.11` image.
Update requirements version to enable support for Python 3.11.
Below are the required changes:
- `httptools` that is a requirement for `uvicorn`, failed to
    build for existing package version that was `0.13.4`.
```
    gcc -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -I/usr/local/include/python3.11 -I/tmp/pip-install-o_24u7rm/httptools_8955220ea3f34d0bae14ed2b27b468ae/vendor/http-parser -c httptools/parser/parser.c -o build/temp.linux-x86_64-cpython-311/httptools/parser/parser.o -O2
    httptools/parser/parser.c:212:12: fatal error: longintrepr.h: No such file or directory
    212 |   #include "longintrepr.h"
           |            ^~~~~~~~~~~~~~~
    compilation terminated.
    error: command '/usr/bin/gcc' failed with exit code 1
    [end of output]    
    note: This error originates from a subprocess, and is likely not a problem with pip.
    ERROR: Failed building wheel for httptools
```
Fix it by using `uvicorn` version with Python 3.11 support i.e. `0.29.0`.
- Since `uvicorn` is one of the requirements of `fastapi`, we also need to update its version to make it compatible with `uvicorn`. The fastapi version `0.99.1` has been selected to avoid dependency conflicts with
existing `fastapi-users` version i.e. `10.4.0`. Otherwise the below build failure will take place:
```
    ERROR: Cannot install fastapi-users, fastapi-users[beanie,oauth]==10.4.0 and fastapi[all]==0.110.0 because
    these package versions have conflicting dependencies.
```
- `fastapi==0.99.1` also removes the requirement of pinning `pyyaml` version.
Dropped: 473bd7b ("docker/api: pin PyYAML version to 5.3.1")
- Also, upgrade `motor` version to use it with python 3.11.
- Upgrade `pylint` to the latest version i.e. `3.1.0`to resolve below incompatibility issue from `wrapt` package:
```
    File "/usr/local/lib/python3.11/site-packages/wrapt/decorators.py", line 34, in <module>
        from inspect import ismethod, isclass, formatargspec
    ImportError: cannot import name 'formatargspec' from 'inspect' (/usr/local/lib/python3.11/inspect.py)
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,7 +20,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ['3.10'] # Only supported one at the moment
+        python-version: ['3.10', '3.11']
 
     steps:
 
--- a/tests/e2e_tests/test_unsubscribe_handler.py
+++ b/tests/e2e_tests/test_unsubscribe_handler.py
@@ -9,18 +9,19 @@
 import pytest
 
 
+@pytest.mark.asyncio
 @pytest.mark.dependency(
     depends=[
         'tests/e2e_tests/test_subscribe_handler.py::test_subscribe_node_channel'],
     scope='session')
 @pytest.mark.order("last")
-def test_unsubscribe_node_channel(test_client):
+async def test_unsubscribe_node_channel(test_async_client):
     """
     Test Case : Test KernelCI API '/unsubscribe' endpoint with 'node' channel
     Expected Result :
         HTTP Response Code 200 OK
     """
-    response = test_client.post(
+    response = await test_async_client.post(
         f"unsubscribe/{pytest.node_channel_subscription_id}",  # pylint: disable=no-member
         headers={
             "Authorization": f"Bearer {pytest.BEARER_TOKEN}"  # pylint: disable=no-member
@@ -29,18 +30,19 @@ def test_unsubscribe_node_channel(test_client):
     assert response.status_code == 200
 
 
+@pytest.mark.asyncio
 @pytest.mark.dependency(
     depends=[
         'tests/e2e_tests/test_subscribe_handler.py::test_subscribe_test_channel'],
     scope='session')
 @pytest.mark.order("last")
-def test_unsubscribe_test_channel(test_client):
+async def test_unsubscribe_test_channel(test_async_client):
     """
     Test Case : Test KernelCI API '/unsubscribe' endpoint with 'test_channel'
     Expected Result :
         HTTP Response Code 200 OK
     """
-    response = test_client.post(
+    response = await test_async_client.post(
         f"unsubscribe/{pytest.test_channel_subscription_id}",  # pylint: disable=no-member
         headers={
             "Authorization": f"Bearer {pytest.BEARER_TOKEN}"  # pylint: disable=no-member
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,7 +20,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ['3.10'] # Only supported one at the moment
+        python-version: ['3.10', '3.11']
 
     steps:
 
--- a/api/main.py
+++ b/api/main.py
@@ -10,7 +10,7 @@
 
 import os
 import re
-from typing import List
+from typing import List, Union
 from fastapi import (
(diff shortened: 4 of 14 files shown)


JenySadadia:
Seems like API tests are not happy with this upgrade. Looking into the issues.

pawiecz:
As discussed offline today `fastapi` will be upgraded to the most recent version (resolving #321) as a next step. For now an updated and *compatible with other components* version was selected.

(btw we run into [`uvicorn` release window](https://pypi.org/project/uvicorn/#history) :smile:)

.github/workflows/test.yml
```diff
@@ -20,7 +20,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ['3.10'] # Only supported one at the moment
+        python-version: ['3.10', '3.11'] # Only supported one at the moment
```
pawiecz, line 23: nit: should this comment be dropped now (like above)?

```suggestion
        python-version: ['3.10', '3.11']
```

JenySadadia: Yes, I changed it locally but forgot to push.

JenySadadia:
> As discussed offline today `fastapi` will be upgraded to the most recent version (resolving #321) as a next step. For now an updated and _compatible with other components_ version was selected.
> 
> (btw we run into [`uvicorn` release window](https://pypi.org/project/uvicorn/#history), it has just been bumped to 0.29 😄)

Nice, let's use it then. I'll test it.

JenySadadia:
> As discussed offline today `fastapi` will be upgraded to the most recent version (resolving #321) as a next step. For now an updated and _compatible with other components_ version was selected.

Added description to commit message.

tests/e2e_tests/test_unsubscribe_handler.py
```diff
@@ -9,18 +9,19 @@
...
 
 
+@pytest.mark.asyncio
 @pytest.mark.dependency(
     depends=[
         'tests/e2e_tests/test_subscribe_handler.py::test_subscribe_node_channel'],
     scope='session')
 @pytest.mark.order("last")
-def test_unsubscribe_node_channel(test_client):
+async def test_unsubscribe_node_channel(test_async_client):
```
pawiecz, line 18: Does the client have to subscribe to the channel before unsubscribing or is this endpoint idempotent (essentially noop) if there was no subscription beforehand?

This could have been the reason why these tests were using synchronous client.

tests/e2e_tests/test_unsubscribe_handler.py
```diff
@@ -29,18 +30,19 @@ def test_unsubscribe_node_channel(test_client):
...
 
 
+@pytest.mark.asyncio
 @pytest.mark.dependency(
     depends=[
         'tests/e2e_tests/test_subscribe_handler.py::test_subscribe_test_channel'],
     scope='session')
 @pytest.mark.order("last")
-def test_unsubscribe_test_channel(test_client):
+async def test_unsubscribe_test_channel(test_async_client):
```
pawiecz, line 39: See `node channel` comment

tests/e2e_tests/test_unsubscribe_handler.py
JenySadadia: See the dependency marker for the test.
The `unsubscribe` test will only execute after a successful execution of `test_subscribe_handler.py::test_subscribe_node_channel` test.

pawiecz: You're right, I missed that - thanks!

tests/e2e_tests/test_unsubscribe_handler.py
pawiecz: (resolved for the same reason as `node` channel)

pawiecz (approved):
LGTM

JenySadadia:
Tested OK on staging.

PR #505, opened 2024-03-20
api/main: Add dashboard

nuclearcat:
Add mini-dashboard that allows to see quickly results of builds.


nuclearcat:
I am merging this PR, as i need to create one more endpoint for custom checkout submissions

PR #507, opened 2024-03-25
doc: update Atlas dashboard link and filtered graph

gctucker:
Fix the wording about the Atlas public dashboard and replace the embedded graph with the one showing filtered data over a 30-day rolling period.

### DIFF
--- a/doc/_index.md
+++ b/doc/_index.md
@@ -28,11 +28,13 @@ step that produced the node passed or failed.  Finally there's a list of
 artifacts with URLs to know where to find all the related files (binaries,
 logs, generated results etc.).
 
-Here's a publicly available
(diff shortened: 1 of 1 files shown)


gctucker:
Verified on staging: https://static.staging.kernelci.org/docs/api/#database-abstraction

(except there doesn't appear to be any new nodes since the beginning of the month in Atlas)

gctucker:
@nuclearcat FYI In fact this might need to be updated again if the current Atlas account is closed.  If we can "move" it to the new Azure subscription (or another Cloud vendor, up to sysadmins to decide) then maybe we can keep this dashboard as-is and the same link.

gctucker:
As discussed last week with @nuclearcat, the plan is to take a dump of the current database and recreate a new one.  Then the current Atlas one will be shutdown, so this Atlas dashboard won't be available or useful any more.  Closing now as a result.

PR #510, opened 2024-04-01
docker-compose.yaml: Ramp up staging mongo RAM

nuclearcat:
We need to increase a bit staging ram, as dataset keep growing.


nuclearcat:
Tested on staging

PR #514, opened 2024-04-05
api: Add manage endpoint, checkout manual submit

nuclearcat:
Checkout manual submit with job filter will allow to submit jobs for regressions.


nuclearcat:
As initial version should work fine, tested on staging

PR #516, opened 2024-04-09
api/config: Add token expiration

nuclearcat:
To have at least some token variability we need to add more entropy. Setting expiration to 10 years might be enough, as it will force token signed data to include this information.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -13,7 +13,7 @@ class AuthSettings(BaseSettings):
     secret_key: str
     algorithm: str = "HS256"
     # Set to None so tokens don't expire
-    access_token_expire_seconds: float = None
+    access_token_expire_seconds: float = 315360000
 
 
 class PubSubSettings(BaseSettings):


JenySadadia:
Yes, I understand adding token expiry is important.
But why 10 years? Will it serve the purpose? I think we should reduce it to something that can actually be useful.

nuclearcat:
As before token was permanent, i kept something equivalent. Setting it to less term might have undesired consequences (downtime), as unlikely we will have soon any mechanism to track token expiration and warn about expiring ones.

JenySadadia:
Okay. If we want something equivalent, why don't we keep it as-is for now and add a new token expiry time once we have the mechanism for handling expired tokens in place?
Sorry, but I just want to understand why we are adding the token expiry time as 10 yrs.

nuclearcat:
> Okay. If we want something equivalent, why don't we keep it as-is for now and add a new token expiry time once we have the mechanism for handling expired tokens in place? Sorry, but I just want to understand why we are adding the token expiry time as 10 yrs.

Right now, without any signed data variability in token - token will be always the same. This means you cannot generate 2 or more unique tokens.

JenySadadia:
Yes, that's true.
I tried this locally and even after setting an expiry time, it generates the same token every time. Did it behave differently when you tested?

nuclearcat:
Yes, with expiry - token should have to sign expiry time, so token should be different, i tested that.
```
./kci-staging.sh  user token xxx
Password: 
<hidden>.1ap1iIHCy7HHhBTm<hidden>
./kci-staging.sh  user token xxx
Password: 
<hidden>.vTU7vjDKXX86I813<hidden>
```
First part of token looks same, but final symbols are different (expiration date and signature).

JenySadadia:
You're right. The change in the expiry time wasn't reflected in my case as I wasn't using the API in dev mode.
Thanks for clarifying.

JenySadadia (approved):
Could you please rebase it to fix GH checks?

nuclearcat:
Done

PR #511, opened 2024-04-02
Use `kernelci` tag for `kernelci/staging-api` image

JenySadadia:
Depends on https://github.com/kernelci/kernelci-core/pull/2485

Template for building `api` docker image has been updated to use `kernelci` fragment for using `kernelci` python package.
Hence, use `kernelci` tag for `kernelci/staging-api` image.

PR #517, opened 2024-04-11
tests: add new model field

nuclearcat:
Add model field to tests.


JenySadadia (approved):
The GH checks will be passed once https://github.com/kernelci/kernelci-api/pull/511 is merged.

PR #527, opened 2024-05-07
api.main: publish events on `build` channel

JenySadadia:
Open a new PubSub channel to publish kernel build nodes i.e. `build`. The events on this channel would not have `owner` information associated. Hence, subscribers will receive all the build events and not just filtered events by node owners.


padovan:
This is not necessary anymore.

PR #528, opened 2024-05-08
viewer: Add more intuitive name

nuclearcat:
Many people don't spot what is "->", i should have given it logical name.


JenySadadia (approved):
Thanks, that was quick :)

PR #529, opened 2024-05-09
api.main: add `is_hierarchy` to node event data

JenySadadia:
Add a boolean key named `is_hierarchy` node event data that denotes if the node event represents submission of node hierarchy. It would be required to distinguish such events from single node create/update events.


JenySadadia:
Tested OK on staging.

PR #532, opened 2024-05-23
api: Implement pub/sub promisc mode

nuclearcat:
Some labs need to subscribe to all events, so they can receive other users kernel builds. Implement optional "promiscuous" option to make subscription listen to all users events.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -11,7 +11,7 @@
 import os
 import re
 import hashlib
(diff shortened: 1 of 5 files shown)


nuclearcat:
In this particular case we just need kernel build nodes.

nuclearcat:
Adding so production will have this feature

PR #533, opened 2024-05-29
Re-structure API docs

JenySadadia:
Related to https://github.com/kernelci/kernelci-project/issues/338


doc/design.md
```diff
@@ -89,7 +89,7 @@ graph LR
 ```
 
 Each step is a client for the API and related storage solutions.  They are all
-implemented in Python, rely on the [KernelCI Core](/docs/core) tools and run in
+implemented in Python, rely on the [KernelCI Core](/docs/../core) tools and run in
```
pawiecz, line 92: nit: `/docs/` prefix was dropped from all other links - it there any particular reason to leave this one?

JenySadadia: All other links refer to the pages in the same directory i.e. `api`.
The Core tool doc is from `legacy` section.
`kernelci-project` has this directory structure `docs/legacy/core` and `docs/api_pipeline/api`.
That's why the link is different.

pawiecz (approved):
LGTM

Issue #534, opened 2024-05-30
Bug: Response error in user password update request


JenySadadia:
Reported by @mwasilew

`kci` tool for a user password update command emits an error:
```
$ ./kci user password update test
Current password: 
New password: 
Retype new password: 
Error: 400 Client Error: Bad Request for url: http://172.17.0.1:8001/latest/user/update-password
1 validation error for UserRead
response
  none is not an allowed value (type=type_error.none.not_allowed)
``` 

API logs:
```
kernelci-api | User 65f940b449418a6c735eb580 test has been updated.
kernelci-api | INFO:     172.19.0.1:46820 - "POST /user/update-password HTTP/1.1" 400 Bad Request
```

JenySadadia:
Reproduced it with API endpoint:
```
 curl -X 'POST'   'http://localhost:8001/latest/user/update-password'   -H 'Content-Type: application/x-www-form-urlencoded'    -d 'username=test&password=pwd&new_password=new_pwd'
{"detail":"1 validation error for UserRead\nresponse\n  none is not an allowed value (type=type_error.none.not_allowed)"}
```

PR #535, opened 2024-05-30
api.main: fix POST `/user/update_password` endpoint

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/534

Fix the below validation error while requesting user password update:
```
{"detail":"1 validation error for UserRead\nresponse\n  none is not an
allowed value (type=type_error.none.not_allowed)"}
```
The endpoint returns null on successful request even if response model is specified (and it's not None).

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -344,7 +344,7 @@ async def get_users(request: Request,
     return paginated_resp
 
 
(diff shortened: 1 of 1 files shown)


JenySadadia:
Tested OK on staging with `kci` tool.

PR #543, opened 2024-08-22
fix(db): Do not allow creating objects with wrong parent

nuclearcat:
If parent is set to same id as object it might cause loops in some of pipeline programs. Block such updates.

### DIFF
--- a/api/db.py
+++ b/api/db.py
@@ -143,6 +143,8 @@ class Database:
             obj.parent = parent.id
         if obj.id:
             obj.update()
+            if obj.parent == obj.id:
+                raise ValueError("Parent cannot be the same as the object")
             res = await col.replace_one(
                 {'_id': ObjectId(obj.id)}, obj.dict(by_alias=True)
             )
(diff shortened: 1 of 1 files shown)


JenySadadia (commented):
I am wondering if this check should be in the endpoint handler instead, in `api.main` just like other checks we have e.g. `_verify_user_group_existence`.
As this check doesn't concern DB abstraction layer.
What do you think? @nuclearcat

nuclearcat:
Yes, it make sense, at least i will try to rewrite and add it to endpoint.

nuclearcat:
@JenySadadia after returning to this question i remember why i moved to db layer. When we submit multiple nodes over PUT /nodes call, we are not unrolling nodes in api.main: https://github.com/kernelci/kernelci-api/blob/ce9ea25f052253814fa1fe3845d5f113b72650ae/api/main.py#L666 so i dont have access to all individual nodes to verify them easily.
But rather forward it to db layer, this is why i moved parent verification in db layer, as otherwise i will have to duplicate same code and unroll hierarchy twice - in main code too.

JenySadadia:
> @JenySadadia after returning to this question i remember why i moved to db layer. When we submit multiple nodes over PUT /nodes call, we are not unrolling nodes in api.main:
> 
> https://github.com/kernelci/kernelci-api/blob/ce9ea25f052253814fa1fe3845d5f113b72650ae/api/main.py#L666
> 
> so i dont have access to all individual nodes to verify them easily.
> But rather forward it to db layer, this is why i moved parent verification in db layer, as otherwise i will have to duplicate same code and unroll hierarchy twice - in main code too.

Okay. Usually we create sub-nodes rather than updating them with `PUT /nodes` request. But, yes, it's good to have validation there as well.
Let's go with `db` layer in this case.

PR #544, opened 2024-08-23
api: templates: viewer: add missing LAVA URLs

a-wai:
Since the "LAVA Job" button was first implemented, a few LAVA labs have been connected to KernelCI. Using the "LAVA Job" button for jobs scheduled on those would then error out, the base URL being `undefined`.

Add the missing labs so we can more easily inspect LAVA logs and job definitions.


nuclearcat (approved):
LGTM

PR #548, opened 2024-09-19
fix(requirements.txt): Add correct pymongo/motor dependencies

nuclearcat:
pymongo got updated and it was not fixed version in dependencies, as result it became incompatible with motor.
Trying to fix versions to working ones.

### DIFF
--- a/docker/api/requirements.txt
+++ b/docker/api/requirements.txt
@@ -4,7 +4,8 @@ fastapi-pagination==0.9.3
 fastapi-users[beanie, oauth]==10.4.0
 fastapi-versioning==0.10.0
 MarkupSafe==2.0.1
-motor==3.3.2
+pymongo==4.9
+motor==3.6.0
 passlib==1.7.4
 pydantic==1.10.13
 pymongo-migrate==0.11.0


docker/api/requirements.txt
```diff
@@ -4,7 +4,8 @@ fastapi-pagination==0.9.3
 fastapi-users[beanie, oauth]==10.4.0
 fastapi-versioning==0.10.0
 MarkupSafe==2.0.1
-motor==3.3.2
+pymongo==4.9
```
JenySadadia, line 7:
```suggestion
pymongo==4.9.0
```

pyproject.toml
```diff
@@ -18,7 +18,8 @@ dependencies = [
   "fastapi-users[beanie, oauth] == 10.4.0",
   "fastapi-versioning == 0.10.0",
   "MarkupSafe == 2.0.1",
-  "motor == 3.3.2",
+  "pymongo == 4.9.0",
+  "motor == 3.6.0",
```
JenySadadia, line 22: Does it work if we keep packages in alphabetical order or do we need to pin `pymongo` first?
```suggestion
  "motor == 3.6.0",
  "pymongo == 4.9.0",
```

nuclearcat: Yes, seems there is no requirements, so we can reorder it alphabetically

JenySadadia (approved):

JenySadadia:
LGTM. Let's wait for staging results.

PR #552, opened 2024-10-02
feat(main.py): Non-paginated get_nodes 

nuclearcat:
This is non-paginated version of get_nodes query,
to test if it solves performance/latency issues.


nuclearcat:
Tested on staging, used in kci-dev

PR #553, opened 2024-10-10
feat(viewer.html): Add modal window to show transfer in progress

nuclearcat:
When we request large amount of data mongo/pagination might be slow, thus we need to show user we are in progress, and it is not a bug, that nothing appeared yet.


nuclearcat:
Tested on staging

PR #559, opened 2024-10-24
feat(api): Add key/value store with namespace emulation

nuclearcat:
For storing pipeline settings and variable and make them a bit more stateful we need reliable key/value store.


nuclearcat:
tested on staging

PR #567, opened 2024-11-04
feat(viewer): Highlight manually submitted entries different way

nuclearcat:
We need to spot in viewer manually submitter entries somehow.


nuclearcat:
tested on staging

PR #569, opened 2024-11-06
api: add `metrics.py`

JenySadadia:
Move `Metrics` class implementation to separate file to reduce number of lines in `api.main` module and to keep the various modules more organized.
Otherwise pylint will complain with the below error: "api/main.py: C0302: Too many lines in module (1044/1000) (too-many-lines)"

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -11,7 +11,6 @@
 import os
 import re
 from typing import List, Union, Optional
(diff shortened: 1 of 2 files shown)


JenySadadia:
Merging this to fix GH actions in other PRs.

Issue #575, opened 2024-11-26
BUG: Not able to create a super user


JenySadadia:
Observed a couple of issues while registering a super user with `is_superuser=1` flag in the request dictionary.

Request command:
```
curl -X 'POST' 'http://localhost:8001/latest/user/register' -H 'accept: application/json'   -H 'Content-Type: application/json'  -H 'Authorization: Bearer TOKEN' -d '{"username": "test_admin", "email": "test-admin@kernelci.org", "password": "admin", "is_superuser": 1, "groups": ["kernelci"]}'
```
 
API logs:
```
kernelci-api          | User 674572be9c8d350bbb789ad7 test_admin has registered.
kernelci-api          | exception: update must either be a list or an instance of dict, bson.son.SON, or any other type that inherits from collections.Mapping
kernelci-api          | INFO:     172.18.0.1:41080 - "POST /latest/user/register HTTP/1.1" 500 Internal Server Error
```

This request generates a regular user account instead of a super user.

PR #577, opened 2024-11-26
api.db: fix `Database.update` method

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/575

Database `update` method was breaking while creating a `User` model with `is_superuser=1` field value.
Fix the method by separating `Node` specific logic from other model updates based on below facts:
- `obj.update()` will trigger `BeanieBaseUser.update()` for `User` model and not `DatabaseModel.update()`
- `User` model doesn't have `parent` field


pawiecz (approved):
LGTM

PR #581, opened 2024-12-04
fix(main.py): Introduce request time limit on fast requests

nuclearcat:
If user abuse db query, it should not block HTTP server for long time.


nuclearcat:
Tested on staging

PR #585, opened 2024-12-05
fix(main.py): Improve bulk-set robustness

nuclearcat:
Do not fail if one of nodes have wrong ownership, just ignore this node.


nuclearcat:
Tested on staging

PR #588, opened 2024-12-12
fix(setup_admin_user): Script should be flexible on docker-compose

nuclearcat:
Newer version of docker-compose is called as "docker compose", so try automatically to use it, if docker-compose is missing.


nuclearcat:
Tested on staging

PR #593, opened 2025-01-16
tests/unit_tests: sync unit tests after model changes

pawiecz:
Changes (https://github.com/kernelci/kernelci-core/pull/2746 and https://github.com/kernelci/kernelci-core/pull/2927) to the model caused unit tests to get out-of-sync with its new state.

Tests adjusted in this patch should be kept in sync with [Node class from "kernelci/api/models.py" in "kernelci/kernelci-core" repo](https://github.com/kernelci/kernelci-core/blob/ee64d7e4cdaeef2176d9755c9112ff97c8ad8e3a/kernelci/api/models.py#L184)


JenySadadia (approved):
Thank you!

PR #598, opened 2025-02-24
docker-compose: remove obsolete 'version' attribute

padovan:
Fix this warning:
WARN[0000] kernelci-api/docker-compose.yaml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -3,7 +3,6 @@
 # Copyright (C) 2021 Collabora Limited
 # Author: Guillaume Tucker <guillaume.tucker@collabora.com>
 
-version: '3'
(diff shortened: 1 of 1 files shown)


nuclearcat:
Same issue as in https://github.com/kernelci/kernelci-pipeline/pull/1045#issuecomment-2679728448
Fails with error:
```
ERROR: Invalid interpolation format for "api" option in service "services": "${KERNELCI_API_IMAGE:-kernelci/staging-kernelci}:${KERNELCI_API_TAG:-api}"
```
We might need either invoke shell or make this option fixed.

PR #602, opened 2025-04-08
fix: autoremove admin user creation ephemeral container

tollsimy:
This PR prevents dangling containers (e.g. `kernelci-api-api-run-2e5cf347e346`) when setting up the admin user using `scripts/setup_admin_user` script.


nuclearcat:
Thank for spotting it! Tested on staging

PR #606, opened 2025-04-21
Add query parameter to events endpoint to filter on result

kwilczynski:
Even though the `data.result` query parameter can be added when calling the `/events` endpoint, add the `result` parameter for convenience.

This will also shorten the entire query and hide some of the internal implementation details from the end-user, as in the future, the query parameters might be sanitised such that any unknown or unsupported values could be removed.


nuclearcat (approved):
Thanks a lot Krzysztof, looks good!

PR #615, opened 2025-06-17
api.db: enable `in` and `nin` query operators

JenySadadia:
`in` operator: Enable support for query operator to return documents by matching field value from a list of multiple values.
For instance nodes with state `running` and `done` can be retrieved by the following request:
`http://API_URL/nodes?state__in=running,done`
Multiple values should be comma-separated in the
request URL.

`nin` operator: Enable support for query operator to return documents that do not match any of the field value from a list of multiple values.
For instance nodes with state other than `running` and `done` can be retrieved by the following request:
`http://API_URL/nodes?state__nin=running,done`
Multiple values should be comma-separated in the
request URL just as `in` operator.

### DIFF
--- a/api/db.py
+++ b/api/db.py
@@ -37,6 +37,8 @@ class Database:
         'gte': '$gte',
         'ne': '$ne',
         're': '$regex',
+        'in': '$in',
+        'nin': '$nin',
     }
 
     BOOL_VALUE_MAP = {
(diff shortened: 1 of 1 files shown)


pawiecz (approved):
LGTM

E501 from the failed workflow [is ignored in the Pipeline](https://github.com/kernelci/kernelci-pipeline/commit/4a261a1dcd73d13156cce8c8ddc75840706f6613) - maybe these should be aligned?

One day it would be useful to revive https://github.com/kernelci/kernelci-api/blob/main/tests/unit_tests/test_node_handler.py with query examples

JenySadadia:
> LGTM
> 
> E501 from the failed workflow [is ignored in the Pipeline](https://github.com/kernelci/kernelci-pipeline/commit/4a261a1dcd73d13156cce8c8ddc75840706f6613) - maybe these should be aligned?
> 

I don't think that the flag should be used. Too long lines aren't good for the sake of readability.

> One day it would be useful to revive https://github.com/kernelci/kernelci-api/blob/main/tests/unit_tests/test_node_handler.py with query examples

PR #618, opened 2025-07-08
api/viewer: try to view logs in modal

nuclearcat:


nuclearcat:
CSRF protection kill all the fun, need to patch storage for that, so likely postponed or will be done different way

PR #622, opened 2025-08-05
api/pubsub: Add automatic cleanup of stale subscriptions

nuclearcat:
Implement background task to periodically remove subscriptions that haven't been polled recently, preventing memory leaks from abandoned connections. Configurable intervals and retry logic implemented too.


JenySadadia (approved):
Thanks for the fix.
LGTM.

PR #623, opened 2025-08-05
template/stats: Make limit configurable

nuclearcat:
Turned out, on production we have almost 1000 builds per week, so default is not enough. Let's make few options more.


nuclearcat:
Tested on staging: https://staging.kernelci.org:9000/stats

JenySadadia (approved):
LGTM. Thanks.

PR #627, opened 2025-09-02
Fix: FastAPI /docs OpenAPI schema generation for PydanticObjectId

nuclearcat:
Add custom JSON schema generator to handle Beanie's PydanticObjectId validator which was causing schema generation to fail. The custom generator provides a string schema representation for ObjectId fields, allowing the OpenAPI documentation to be generated successfully.


JenySadadia:
Hello, out of curiosiry, I checked `/docs` endpoint on my local setup. It's working.
Also, https://staging.kernelci.org/docs is working.
Looks like the issue is with the production setup only.

PR #637, opened 2025-11-30
pubsub: Implement optional durable event delivery

nuclearcat:
Major feature: removing "fire-and-forget" event delivery mode with Redis to allow durable event delivery over MongoDB.
Unified storage: Events stored once in eventhistory (was duplicated)
Auto-migration: Detects old 24h TTL format, drops and recreates
Backwards compatible: Without subscriber_id = fire-and-forget (old behavior)
Durable mode: With subscriber_id = catch-up on missed events
Implicit ACK: Polling for next event acknowledges previous


nuclearcat:
tested on staging

